PR #8, opened 2018-06-21
Scaffold out a Resource for an API

pwittrock:


pkg/scaffold/resource/version_suitetest.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if c, err = client.New(cfg, client.Options{Scheme: scheme.Scheme}); err != nil {
+		log.Fatal(err)
+	}
+
+	code := m.Run()
+	t.Stop()
+	os.Exit(code)
+}
+`
```
droot: +1 on using vanilla `go test` tooling :)

PR #23, opened 2018-06-27
Rebase scaffolding on newest controller-runtime

pwittrock:


pkg/scaffold/controller/controller.go
```diff
@@ -158,19 +160,26 @@ func add(mrg manager.Manager, r reconcile.Reconcile) error {
...
-	client client.Client
+	client.Client
+	scheme *runtime.Scheme
 }
 
 // Reconcile reads that state of the cluster for a {{ .Resource.Kind }} object and makes changes based on the state read
 // and what is in the {{ .Resource.Kind }}.Spec
 // TODO(user): Modify this Reconcile function to implement your Controller logic.  The scaffolding writes
 // a Deployment as an example
+{{ if .Resource.CreateExampleReconcileBody -}}
```
Liujingfang1: Is this a new field? What flag should we use in `kubebuilder` to match this field?

pwittrock: This is new in controller-tools because controller-tools scaffolds out a whole example which kubebuilder does not.  Maybe keep the flags the same?

pwittrock:
Adding the labels since you said LGTM in the comment

PR #28, opened 2018-07-02
fix test for core type controller

mengqiy:


pkg/scaffold/controller/controllertest.go
```diff
@@ -113,7 +114,12 @@ func TestReconcile(t *testing.T) {
 	defer close(StartTestManager(mgr, g))
 
 	// Create the {{ .Resource.Kind }} object and expect the Reconcile{{ if .Resource.CreateExampleReconcileBody }} and Deployment to be created{{ end }}
-	g.Expect(c.Create(context.TODO(), instance)).To(gomega.Succeed())
+	err = c.Create(context.TODO(), instance)
```
droot: context.Background() ?

pkg/scaffold/controller/controllertest.go
```diff
@@ -113,7 +114,12 @@ func TestReconcile(t *testing.T) {
...
+	// If getting an Invalid error, you should
+	// 1) modify the "instance" object to make it valid e.g. adding required fields.
+	// 2) remove the following if statement.
+	if apierrors.IsInvalid(err) { return }
```
droot: Lets use `t.Logf` to log the above comments as warning messages so that user can discover that he/she needs to improve these tests.

pkg/scaffold/controller/controllertest.go
```diff
@@ -113,7 +114,12 @@ func TestReconcile(t *testing.T) {
 	defer close(StartTestManager(mgr, g))
 
 	// Create the {{ .Resource.Kind }} object and expect the Reconcile{{ if .Resource.CreateExampleReconcileBody }} and Deployment to be created{{ end }}
-	g.Expect(c.Create(context.TODO(), instance)).To(gomega.Succeed())
+	err = c.Create(context.TODO(), instance)
+	// If getting an Invalid error, you should
+	// 1) modify the "instance" object to make it valid e.g. adding required fields.
+	// 2) remove the following if statement.
```
droot: May be state the reasoning first to get user's attention:
The `instance` object may not be a valid object because it might be missing some required fields. Modify the `instance` object to add required fields and then uncomment the following statement (have the line with error being nil assertion) ?

pkg/scaffold/controller/controllertest.go
mengqiy: context.TODO() has been used multiple time in this test and many other files.
I guess it's better to deal with `context` stuff in a followup PR :)
Filed https://github.com/kubernetes-sigs/controller-tools/issues/29

pkg/scaffold/controller/controllertest.go
mengqiy: Good idea.

pkg/scaffold/controller/controllertest.go
mengqiy: Do you mean we should comment everything below?

IMO we should let it return early when encountering an Invalid error. IOW we should run everything if no Invalid error.

mengqiy: To make it clear, IMO if we get an error from `Create`, it doesn't make a lot of sense to run the remaining test.

mengqiy:
PTAL

PR #30, opened 2018-07-03
added rbac manifests generation pkg and CLI

droot:
Note: need feedback specifically about the pkg and command's location. 

Added `rbac` new commandline tool and `rbac` pkg under `pkg/generate/rbac` for generating RBAC manifests.  It scans the Go source files under given input-dir and parses the RBAC annotations and generates the RBAC manifests. The annotations could be anywhere in the Go files not necessarily tied to a Type or Function (Thats one reason `gengo` wasn't useful). Parser uses `Go` standard packages (ast, parser) instead of `gengo` because requirements were very simple. 

Realized that we will have to generate RBAC annotations while scaffolding the resources so that this tool can generate RBAC manifests.

### DIFF
--- a/cmd/rbac/main.go
+++ b/cmd/rbac/main.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/controller-tools/pkg/generate/rbac"
+)
+
+func main() {
+	o := &rbac.ManifestOptions{}
+	cmd := &cobra.Command{
+		Use:   "rbac",
+		Short: "Generates RBAC manifests",
+		Long: `Generate RBAC manifests from the RBAC annotations in Go source files.
+Usage:
+# rbac generate [--name manager] [--input-dir input_dir] [--output-dir output_dir]
+`,
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := rbac.Generate(o); err != nil {
+				log.Fatal(err)
+			}
+			fmt.Printf("RBAC manifests generated under '%s' directory\n", o.OutputDir)
+		},
+	}
+
+	registerFlags(cmd, o)
+
+	if err := cmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func registerFlags(cmd *cobra.Command, o *rbac.ManifestOptions) {
+	f := cmd.Flags()
+	f.StringVar(&o.Name, "name", "manager", "Name to be used as prefix in identifier for manifests")
+	f.StringVar(&o.InputDir, "input-dir", "./pkg", "input directory pointing to Go source files")
+	f.StringVar(&o.OutputDir, "output-dir", "./config", "output directory where generated manifests will be saved.")
+}
--- a/pkg/generate/rbac/manifests.go
+++ b/pkg/generate/rbac/manifests.go
@@ -1,0 +1,124 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


pkg/generate/rbac/manifests.go
```diff
@@ -0,0 +1,125 @@
...
+
+func getClusterRoleBindingManifest(o *ManifestOptions) ([]byte, error) {
+	rolebinding := &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "rbac.authorization.k8s.io/v1",
+			Kind:       "ClusterRoleBinding",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-rolebinding", o.Name),
+			Namespace: fmt.Sprintf("%s-system", o.Name),
```
mengqiy: IIRC namespace will be ignored.
I think you can drop this line.

pkg/generate/rbac/parser.go
```diff
@@ -0,0 +1,145 @@
...
+		value := kv[1]
+		if strings.HasPrefix(value, "\"") && strings.HasSuffix(value, "\"") {
+			value = value[1 : len(value)-1]
+		}
+		values := strings.Split(value, ";")
+		switch kv[0] {
+		case "groups":
+			normalized := []string{}
+			for _, v := range values {
+				if v == "core" {
```
mengqiy:
```
// +rbac:groups="",resources=pods,verbs=get;list;watch;delete
```
IMO using an empty string for core group should also be valid.
WDYT?

pkg/generate/rbac/parser.go
```diff
@@ -0,0 +1,145 @@
...
+	for _, commentGroup := range cmap.Comments() {
+		for _, comment := range strings.Split(commentGroup.Text(), "\n") {
+			comment := strings.TrimSpace(comment)
+			if strings.HasPrefix(comment, "+rbac") {
+				if ann := getAnnotation(comment, "rbac", ":"); ann != "" {
+					rules = append(rules, parseRBACTag(ann))
+				}
+			}
+			if strings.HasPrefix(comment, "+kubebuilder:rbac") {
+				if ann := getAnnotation(comment, "kubebuilder:rbac", ":"); ann != "" {
```
mengqiy: Are we going support 2 annotations forever?
If not, we should print warning for the one that we eventually will drop support.

pkg/generate/rbac/parser_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}{
+		{
+			content: `package foo
+	import (
+		"fmt"
+		"time"
+	)
+
+	// RBAC annotation with kubebuilder prefix
+	// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;delete
```
mengqiy: We may want to have more test cases to cover
1) core group
2) using `""`

pkg/generate/rbac/manifests.go
droot: Done.

pkg/generate/rbac/parser_test.go
droot: Added.

pkg/generate/rbac/parser.go
droot: Yes, it is. Current code is supporting both "" and "core".
Not dropping the support for "core" yet because old projects might have them, so it might help with the migration.

pkg/generate/rbac/parser.go
droot: We might be able to drop the "rbac" annotation completely for new projects. Though keeping it around will help with the migration. 
Would like to deprecate it in separate PR (along with other deprecations).

cmd/rbac/main.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func main() {
+	o := &rbac.ManifestOptions{}
+	cmd := &cobra.Command{
+		Use:   "rbac",
+		Short: "Generates RBAC manifests",
+		Long: `Generate RBAC manifests from the RBAC annotations in Go source files.
+Usage:
+# rbac generate [--name manager] [--input-dir input_dir] [--output-dir output_dir]
```
lichuqiang: I think the sub command "generate" here did not take effect. i.e. ”rbac [--name manager] [--input-dir input_dir] [--output-dir output_dir]“ can also work.

Maybe add it as a sub command explicitly

pkg/generate/rbac/manifests.go
```diff
@@ -0,0 +1,124 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ManifestOptions represent options for generating the RBAC manifests.
+type ManifestOptions struct {
+	InputDir  string
+	OutputDir string
+	Name      string
+	Labels    map[string]string
```
Liujingfang1: Where is Labels read in? I don't see a flag for it

cmd/rbac/main.go
droot: Good pt. Created a new command `controller-gen` and added `rbac` as a subcommand. So here is the plan:

we will invoke `controller-gen all` using `go generate` directive from the generated project. `all` subcommand will trigger all the generators for a project.

pkg/generate/rbac/manifests.go
droot: Didn't implement the flag part in this version :). Will address it a separate PR.

droot:
Addressed the command. Main change is created a new command controller-gen and added rbac as a subcommand. So here is the plan:

we will invoke controller-gen all using go generate directive from the generated project. all subcommand will trigger all the generators for a project.

@mengqiy PTAL.

PR #36, opened 2018-07-10
controller-gen: integrate crd generator

droot:


droot:
@mengqiy updated it, PTAL.

PR #40, opened 2018-07-12
Moved rbac/crd manifests generation to makefile

droot:
Highlight:
 - Removed rbac/crd from scaffolding workflow
 - RBAC/CRD will be generated by Makefile target 'manifests'
 - Made Makefile target 'manifests' pre-requiste for other targets

Switched manifests generation command from 'go run' to 'go build'
because of the way test directory is setup. test directory's vendor
points to parent controller-tools directory, so Makefile inside test can
not refer with relative syntax (./vendor/.../controller-tools..) because
that will introduce cycle in vendor tree.


pkg/scaffold/project/makefile.go
```diff
@@ -55,12 +55,18 @@ run: generate fmt vet
...
+# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: manifests
+	kubectl apply -f config/rbac
+	kubectl apply -f config/crds
+	kubectl apply -f config/manager
+
 # Generate manifests e.g. CRD, RBAC etc.
 manifests:
-	go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
+	go build -o /tmp/controller-gen sigs.k8s.io/controller-tools/cmd/controller-gen && /tmp/controller-gen all
```
mengqiy: Why change it to this way now?

test/config/crds/creatures_v2alpha1_kraken.yaml
```diff
@@ -1,13 +1,33 @@
...
+          type: string
+        metadata:
+          type: object
+        spec:
+          type: object
+        status:
+          type: object
+      type: object
+  version: v2alpha1
+status:
```
mengqiy: Why those new field got generated?

droot: This is because we are YAML marshaling the CRD object, the previous generation was template based.

droot:
@mengqiy PTAL

PR #41, opened 2018-07-13
kustomization.yaml and kustomize image patch scaffolding

droot:


pkg/scaffold/project/kustomize.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		c.Prefix = filepath.Base(dir)
+	}
+	c.TemplateBody = kustomizeTemplate
+	c.Input.IfExistsAction = input.Error
+	return c.Input, nil
+}
+
+var kustomizeTemplate = `# Adds namespace to all resources.
+namespace: {{.Prefix}}-system
```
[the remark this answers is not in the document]
droot: scaffolding creates resource for namespace, not in a separate file though. Its in the same file as `manager.yaml`.

pkg/scaffold/project/kustomize.go
```diff
@@ -0,0 +1,77 @@
...
+#commonLabels:
+#  someName: someValue
+
+# Each entry in this list must resolve to an existing
+# resource definition in YAML.  These are the resource
+# files that kustomize reads, modifies and emits as a
+# YAML string, with resources separated by document
+# markers ("---").
+resources:
+- ../rbac/*.yaml
```
mengqiy: Do we really need `*.yaml`? What else do we plan to put under `/rbac` and `/manager`?

test/config/default/manager_image_patch.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - image: IMAGE_URL
```
pwittrock: Add a comment telling user to replace this with their image

test/config/default/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
...
+#  someName: someValue
+
+# Each entry in this list must resolve to an existing
+# resource definition in YAML.  These are the resource
+# files that kustomize reads, modifies and emits as a
+# YAML string, with resources separated by document
+# markers ("---").
+resources:
+- ../rbac/*.yaml
+- ../manager/*.yaml
```
pwittrock: crds/*.yaml

pkg/scaffold/project/kustomize.go
```diff
@@ -0,0 +1,79 @@
...
+- ../manager/*.yaml
```
pwittrock: ../crds/*.yaml

test/config/default/kustomization.yaml
droot: Not adding CRDs because name-prefix transformer changes the CRD names.

pkg/scaffold/project/kustomize.go
droot: Same comment as above.

test/config/default/manager_image_patch.yaml
droot: Done.

pkg/scaffold/project/kustomize.go
droot: files under rbac/manager may change over time, so don't want Kustomization.yaml scaffolding to be coupled with that.

PR #44, opened 2018-07-13
scaffold crd samples

droot:


pkg/scaffold/resource/crd_sample.go
```diff
@@ -0,0 +1,66 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resource
+
+import (
+	"fmt"
+	"path/filepath"
+
```
mengqiy: nit: remove this newline.

pkg/scaffold/resource/crd_sample.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var _ input.File = &CRD{}
+
+// CRDSample scaffolds a manifest for CRD sample.
+type CRDSample struct {
+	input.Input
+
+	// Scope is Namespaced or Cluster
+	Scope string
```
mengqiy: Creating a type for Scope instead of plain string?

pkg/scaffold/resource/crd_sample.go
droot: Done.

pkg/scaffold/resource/crd_sample.go
droot: removed scope because we aren't using it in this case.

PR #45, opened 2018-07-14
fixed api-group and resource field in RBAC annotation

droot:
Found this bug during testing today where RBAC manifests for CRD types didn't include fully qualified api-group-name (with domain suffix etc.) and also discovered `resources` were not being pluralized properly.

### DIFF
--- a/pkg/scaffold/controller/controller.go
+++ b/pkg/scaffold/controller/controller.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/markbates/inflect"
 	"sigs.k8s.io/controller-tools/pkg/scaffold/input"
 	"sigs.k8s.io/controller-tools/pkg/scaffold/resource"
 )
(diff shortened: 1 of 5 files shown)


pkg/scaffold/controller/controller.go
```diff
@@ -68,6 +72,11 @@ func (a *Controller) GetInput() (input.Input, error) {
 		a.GroupDomain = a.Resource.Group + "." + a.Domain
 	}
 
+	if a.Plural == "" {
+		rs := inflect.NewDefaultRuleset()
+		a.Plural = rs.Pluralize(strings.ToLower(a.Resource.Kind))
```
mengqiy: I guess you need a `Kind` to `Resources` converter.
https://github.com/kubernetes/kubernetes/blob/a30f4591609301bd6160abeb22731f70e7dae2e0/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L123
This function has been used in several places in k/k.

droot: Thanks for the tip. There are couple of places where we mapping a kind to a resource in controller-tools. I can use the recommended way in a separate PR for all the places to keep it consistent.

PR #47, opened 2018-07-16
controller-tools path in Makefile scaffolding

droot:


cmd/controller-scaffold/cmd/project.go
```diff
@@ -158,3 +160,10 @@ func BoilerplateForFlags(f *flag.FlagSet) *project.Boilerplate {
 		"Owner to add to the copyright")
 	return b
 }
+
+// MakefileForFlags registers flags for Makefile fields and returns the Makefile
+func MakefileForFlags(f *flag.FlagSet) *project.Makefile {
+	m := &project.Makefile{Image: imgName}
+	f.StringVar(&m.ControllerToolsPath, "controller-tools-path", "", "path to controller tools repo")
```
mengqiy: Maybe call it `controller-tools-gopath` and make it hidden

droot: This will not be exposed in Kubebuilder at all, so hiding it here doesn't make sense.

PR #54, opened 2018-08-02
Fix domain resource issues in controller

fanzhangio:
bind resource for controller based on whether resource is created.
1) If resource is created in project, bind with created resource.
2) If resource not created, gkv matches core types, bind with core-types.
3) Else, using specified resourcePath by flag --resource-pkg-path (TODO)

Fix issues:
https://github.com/kubernetes-sigs/kubebuilder/issues/342
https://github.com/kubernetes-sigs/kubebuilder/issues/335
https://github.com/kubernetes-sigs/kubebuilder/issues/338

### DIFF
--- a/pkg/scaffold/controller/controller.go
+++ b/pkg/scaffold/controller/controller.go
@@ -17,6 +17,8 @@ limitations under the License.
 package controller
 
 import (
+	"fmt"
+	"os"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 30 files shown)


pkg/scaffold/controller/controller.go
```diff
@@ -87,6 +81,24 @@ func (a *Controller) GetInput() (input.Input, error) {
...
+			if domain != "" {
+				a.GroupDomain = a.Resource.Group + "." + domain
+			}
+			return
+		}
+		// TODO: need to support '--resource-pkg-path' flag for specifying resourcePath
+	}
+	a.ResourcePackage = path.Join(a.Repo, "pkg", "apis")
+	a.GroupDomain = a.Resource.Group + "." + a.Domain
+}
```
droot: we can capture this logic in a separate function so that it can be called by both Test/Controller methods ?

pkg/scaffold/controller/controllertest.go
```diff
@@ -59,17 +61,26 @@ func (a *Test) GetInput() (input.Input, error) {
...
+			return
+		}
+		// TODO: need to support '--resource-pkg-path' flag for specifying resourcePath
+	}
+	a.ResourcePackage = path.Join(a.Repo, "pkg", "apis")
+}
+
 var controllerTestTemplate = `{{ .Boilerplate }}
 
 package {{ lower .Resource.Kind }}
```
droot: We should add two tests by modifying the `generate_golden.sh` by creating controller for existing type under `pkg/apis` which matches the group with core and second one with controller-only for core resource.

pkg/scaffold/controller/controller.go
fanzhangio: Done

pkg/scaffold/controller/controllertest.go
fanzhangio: No problem. Added to test.

PR #56, opened 2018-08-09
Fix bug #351 type referenced issue in CRD generation

fanzhangio:
- remvoe core type validation in resource filter
- add resource annotation in types

Fix bug : 
https://github.com/kubernetes-sigs/kubebuilder/issues/351

### DIFF
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -48,18 +48,6 @@ func IsAPIResource(t *types.Type) bool {
 		}
 	}
 
-	typeMetaFound, objMetaFound := false, false
-	for _, m := range t.Members {
-		if m.Name == "TypeMeta" && m.Type.String() == "k8s.io/apimachinery/pkg/apis/meta/v1.TypeMeta" {
-			typeMetaFound = true
-		}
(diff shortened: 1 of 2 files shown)


pkg/internal/codegen/parse/util.go
```diff
@@ -48,18 +48,6 @@ func IsAPIResource(t *types.Type) bool {
...
-			typeMetaFound = true
-		}
-		if m.Name == "ObjectMeta" && m.Type.String() == "k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta" {
-			objMetaFound = true
-		}
-		if typeMetaFound && objMetaFound {
-			return true
-		}
-	}
 	return false
```
droot: This will break existing projects because of unavailability of `resource` annotations on the types.

fanzhangio: The annotation works for multiple api packages, and could be the best simple approach to solve problem. Yes, the existing project needs to be added by one line of `resource ` annotation. We can go two options:
1) We add a function or scripts to check `resource` annotation for existing projects (including imports), and add it if one missing. This function can be implemented in `ValidateAndInitFields()` under `controller-gen`, which can be called individually or by `Makefile`
2) Let user add this one line annotation by themselves as they needed.

PR #57, opened 2018-08-10
Prompt to install CRD manually when Kind no found

interma:
Fix: https://github.com/kubernetes-sigs/kubebuilder/issues/337

New output:
```
$ make run
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
go run ./cmd/manager/main.go
2018/08/10 14:44:52 Registering Components.
2018/08/10 14:44:52 No kind {core.hawq.org MyResource} found, please install CRD in advance.
exit status 1
make: *** [run] Error 1
```

Thanks for the review.

### DIFF
--- a/pkg/scaffold/manager/cmd.go
+++ b/pkg/scaffold/manager/cmd.go
@@ -46,6 +46,7 @@ import (
 	"log"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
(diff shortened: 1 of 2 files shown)


pkg/scaffold/manager/cmd.go
```diff
@@ -70,12 +71,17 @@ func main() {
...
-		log.Fatal(err)
+		log.Fatalf("Add apis failed, error: %#v.", err)
 	}
 
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
-		log.Fatal(err)
+		switch err.(type) {
+		case *meta.NoKindMatchError:
+			log.Fatalf("No kind %v found, please install CRD in advance.", err.(*meta.NoKindMatchError).GroupKind)
```
droot: I am wondering if we should add more context in the error in the `controller-runtime` library itself instead of putting the logic in generated code, WDYT ?

interma: My thoughts:
Generally, lib (controller-tools) emits all errors and client (kubebuilder) decides what behavior to perform. Based on this point, my PR make sense.

But considering our client code is generated, more concise the better. I think to handle this error in controller-tools is also a good idea. Do you have some ideas about which code should I modify?

Thanks!

interma: @droot just a ping, any thoughts?

droot: Sorry, didn't follow this thread.

Actually, our client code is no longer generated. Kubebuilder project uses client pkg from https://github.com/kubernetes-sigs/controller-runtime. May be, we can just wrap the error when returning from client's pkg. Can you pl. investigate.

interma: Sure, I will check later. 
I will ping you if have some problems. Thanks.

interma:
@droot I opened a new PR in controller-runtime:
https://github.com/kubernetes-sigs/controller-runtime/pull/129

Please help to review, thanks.

droot:
closing this PR for now.

PR #58, opened 2018-08-13
controller-gen: derive apispkg from project-dir

droot:
fixes https://github.com/kubernetes-sigs/kubebuilder/issues/351

This change improves determining whether a type belong to the current project or not by explicitly setting the `apisPkg` in the parser.


droot:
@Liujingfang1 @fanzhangio @mengqiy removed unused code in the second commit.

PR #63, opened 2018-08-22
cleaned up Gopkg.toml with pinning only k8s 1.10 deps

droot:
Note: since `test` project resides in the same repo, `vendor` dir at the root contains `sigs.k8s.io/controller-runtime`. I will pin `controller-runtime` to semver once `controller-runtime` tag is updated to `0.1.0`.


Gopkg.toml
```diff
@@ -5,192 +5,84 @@ required = [
...
 
-[[override]]
-name="k8s.io/kube-openapi"
-revision="f08db293d3ef80052d6513ece19792642a289fea"
+# TODO(droot): pin the following deps to a version when they are tagged.
+#[[constraint]]
+#  name = "github.com/go-openapi/spec"
+#
+#[[constraint]]
+#  name = "github.com/golang/glog"
```
mengqiy: To be consistent with `controller-runtime`, `controller-tools` should migrate to `logr` from `glog`.
cc: @DirectXMan12

Gopkg.toml
```diff
@@ -5,192 +5,84 @@ required = [
...
-name="sigs.k8s.io/testing_frameworks"
-revision="f53464b8b84b4507805a0b033a8377b225163fea"
+  name = "gopkg.in/fsnotify.v1"
+  source = "https://github.com/fsnotify/fsnotify.git"
+  version="v1.4.7"
 
-[[constraint]]
-  name = "github.com/spf13/afero"
-  version = "1.1.1"
+[prune]
```
mengqiy: Can we prune unused pkg as well?

Gopkg.toml
droot: Removed it. Using standard 'log' in this case because use-case is print something on stdout as part of CLI command.

Gopkg.toml
droot: Not doing it in this PR because we are using `vendor` directory in root for the `test` project as well `pkg/crd/generator/testData` as well. Once we decouple the deps, then it will be safe to do so.

PR #69, opened 2018-08-28
Feature: support scale sub resource by tag in controller-tools

fanzhangio:
- Add Scale Subresource tag validation for parsing CRD.
- Usage to enable scale subresource in CRD:
   1) add tag close above Resource struct, like:
```golang
      // +kubebuilder:subresource:scale:specpath=.spec.replica,statuspath=.status.replica,selectorpath=.jsonpath
```

`specpath`, `statuspath`, `selectorpath` key values, represents for JSONPath of **SpecReplicasPath**,  **StatusReplicasPath**,  **LabelSelectorPath** separately.
`specpath` and `statuspath` must be provided. `selectorpath` is optional

   2) Modify corresponding Spec and Status struct in types.
   3) Modify types_test.go files accordingly.

https://github.com/kubernetes-sigs/kubebuilder/issues/303

### DIFF
--- a/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
+++ b/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
@@ -11,6 +11,11 @@ spec:
     kind: Toy
     plural: toys
   scope: Namespaced
+  subresources:
+    scale:
+      specReplicasPath: .spec.replicas
+      statusReplicasPath: .status.replicas
+    status: {}
   validation:
     openAPIV3Schema:
       properties:
@@ -59,6 +64,9 @@ spec:
               - 3
               format: int64
               type: integer
+            replicas:
+              format: int32
+              type: integer
             template:
               type: object
             winner:
@@ -66,8 +74,15 @@ spec:
           required:
           - rank
           - template
+          - replicas
           type: object
         status:
+          properties:
+            replicas:
+              format: int32
+              type: integer
+          required:
+          - replicas
           type: object
       type: object
   version: v1alpha1
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -26,7 +26,6 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/pkg/errors"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -76,19 +75,36 @@ func (b *APIs) parseCRDs() {
 						resource.CRD.Spec.Scope = "Namespaced"
 					}
 
-					if HasCategories(resource.Type) {
+					if hasCategories(resource.Type) {
 						categoriesTag := getCategoriesTag(resource.Type)
 						categories := strings.Split(categoriesTag, ",")
 						resource.CRD.Spec.Names.Categories = categories
 						resource.Categories = categories
 					}
 
-					if HasStatusSubresource(resource.Type) {
-						subresources := &v1beta1.CustomResourceSubresources{
-							Status: &v1beta1.CustomResourceSubresourceStatus{},
+					if hasStatusSubresource(resource.Type) {
+						if resource.CRD.Spec.Subresources == nil {
+							resource.CRD.Spec.Subresources = &v1beta1.CustomResourceSubresources{}
 						}
-						resource.CRD.Spec.Subresources = subresources
-						resource.HasStatusSubresource = true
+						resource.CRD.Spec.Subresources.Status = &v1beta1.CustomResourceSubresourceStatus{}
+					}
+
+					if hasScaleSubresource(resource.Type) {
+						if resource.CRD.Spec.Subresources == nil {
+							resource.CRD.Spec.Subresources = &v1beta1.CustomResourceSubresources{}
+						}
+						jsonPath, err := parseScaleParams(resource.Type)
+						if err != nil {
+							log.Fatalf("failed in parsing CRD, error: %v", err.Error())
+						}
+						resource.CRD.Spec.Subresources.Scale = &v1beta1.CustomResourceSubresourceScale{
+							SpecReplicasPath:   jsonPath[specReplicasPath],
+							StatusReplicasPath: jsonPath[statusReplicasPath],
+						}
+						labelSelctor, ok := jsonPath[labelSelectorPath]
+						if ok && labelSelctor != "" {
+							resource.CRD.Spec.Subresources.Scale.LabelSelectorPath = &labelSelctor
+						}
 					}
 
 					if len(resource.ShortName) > 0 {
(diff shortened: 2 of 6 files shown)


pkg/internal/codegen/parse/crd.go
```diff
@@ -84,10 +84,24 @@ func (b *APIs) parseCRDs() {
...
+						}
+						jsonPath := parseJSONPath(resource.Type)
+						labelSelctor := jsonPath["selectorpath"]
+						resource.CRD.Spec.Subresources.Scale = &v1beta1.CustomResourceSubresourceScale{
+							SpecReplicasPath:   jsonPath["specpath"],
+							StatusReplicasPath: jsonPath["statuspath"],
+							LabelSelectorPath:  &labelSelctor,
 						}
-						resource.CRD.Spec.Subresources = subresources
 						resource.HasStatusSubresource = true
```
droot: s/HasStatus/HasScale

Wondering if this is being used anywhere ?

pkg/internal/codegen/parse/util.go
```diff
@@ -151,8 +151,22 @@ func HasStatusSubresource(t *types.Type) bool {
 	return false
 }
 
+// HasScaleSubresource returns true if t is an APIResource annotated with
+// +kubebuilder:subresource:scale
+func HasScaleSubresource(t *types.Type) bool {
```
droot: wondering if these needs to be *really* exported ? I suspect, no. if not, then lets un-export all of these sort of helpers.

pkg/internal/codegen/parse/util.go
```diff
@@ -326,3 +340,21 @@ func checkType(props *v1beta1.JSONSchemaProps, s string, enums *[]v1beta1.JSON)
 		*enums = append(*enums, v1beta1.JSON{Raw: []byte(`"` + s + `"`)})
 	}
 }
+
+// Scale subresource requires specpath, statuspath, selectorpath key values, represents for JSONPath of
+// SpecReplicasPath, StatusReplicasPath, LabelSelectorPath separately. e.g.
+// +kubebuilder:subresource:scale:specpath=.spec.replica,statuspath=.status.replica,selectorpath=
+func parseJSONPath(t *types.Type) map[string]string {
```
droot: This function can encounter errors during parsing, so it should return error.

pkg/internal/codegen/parse/util.go
```diff
@@ -326,3 +340,21 @@ func checkType(props *v1beta1.JSONSchemaProps, s string, enums *[]v1beta1.JSON)
...
+	jsonPath := make(map[string]string)
+	for _, c := range t.CommentLines {
+		if strings.Contains(c, "+kubebuilder:subresource:scale") {
+			paths := strings.Split(c, ":")[3:][0]
+			path := strings.Split(paths, ",")
+			for _, s := range path {
+				kv := strings.Split(s, "=")
+				jsonPath[kv[0]] = kv[1]
```
droot: Lets be conservative and do all sort of checking here. For ex.
len(kv) > 1
k has to be one of ("specpath", "statuspath", "selectorpath")
return an error with good text with recommended fix.

pkg/internal/codegen/parse/util.go
```diff
@@ -326,3 +340,21 @@ func checkType(props *v1beta1.JSONSchemaProps, s string, enums *[]v1beta1.JSON)
...
+			}
+		}
```
droot: We should probably break out of the loop on first comment with scale sub-resource ?

pkg/internal/codegen/parse/util.go
```diff
@@ -326,3 +340,21 @@ func checkType(props *v1beta1.JSONSchemaProps, s string, enums *[]v1beta1.JSON)
...
+	}
+	return jsonPath
+}
```
droot: We can unit test this func as well if you want to perform any negative testing.

pkg/internal/codegen/parse/util.go
fanzhangio: For now, the usage is within the `parse` package. It's true, not necessary to export. Will clean up all these util helpers in `parse` package.

pkg/internal/codegen/parse/crd.go
fanzhangio: Good catch! This is heritage code from old kubebuilder version, will clean it.

pkg/internal/codegen/parse/util.go
fanzhangio: Done

pkg/internal/codegen/parse/util.go
fanzhangio: Good point.

pkg/internal/codegen/parse/util.go
fanzhangio: Let me think about it.

pkg/internal/codegen/parse/util.go
```diff
@@ -326,3 +357,32 @@ func checkType(props *v1beta1.JSONSchemaProps, s string, enums *[]v1beta1.JSON)
 		*enums = append(*enums, v1beta1.JSON{Raw: []byte(`"` + s + `"`)})
 	}
 }
+
+// Scale subresource requires specpath, statuspath, selectorpath key values, represents for JSONPath of
+// SpecReplicasPath, StatusReplicasPath, LabelSelectorPath separately. e.g.
+// +kubebuilder:subresource:scale:specpath=.spec.replica,statuspath=.status.replica,selectorpath=
+func parseJSONPath(t *types.Type) (map[string]string, error) {
```
droot: We should consider renaming it something like `parseScaleParams()` ?

pkg/internal/codegen/parse/util.go
```diff
@@ -326,3 +357,32 @@ func checkType(props *v1beta1.JSONSchemaProps, s string, enums *[]v1beta1.JSON)
...
+	jsonPath := make(map[string]string)
+	for _, c := range t.CommentLines {
+		if strings.Contains(c, "+kubebuilder:subresource:scale") {
+			paths := strings.Split(c, ":")[3:][0]
+			path := strings.Split(paths, ",")
+			if len(path) != 3 {
```
droot: is selectorPaths is necessary ?

pkg/internal/codegen/parse/util.go
```diff
@@ -326,3 +357,32 @@ func checkType(props *v1beta1.JSONSchemaProps, s string, enums *[]v1beta1.JSON)
...
+				return nil, fmt.Errorf(jsonPathError)
+			}
+			for _, s := range path {
+				kv := strings.Split(s, "=")
+				if kv[0] == specReplicasPath || kv[0] == statusReplicasPath || kv[0] == labelSelectorPath {
+					jsonPath[kv[0]] = kv[1]
+				}
```
droot: else part is error. If we do that, then we don't need the second loop below. We should add unit test to verify the logic.

pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
```diff
@@ -11,6 +11,12 @@ spec:
     kind: Toy
     plural: toys
   scope: Namespaced
+  subresources:
+    scale:
+      labelSelectorPath: ""
+      specReplicasPath: .spec.replicas
+      statusReplicasPath: .status.replicas
+    status: {}
```
droot: Can we verify that YAML works against a real cluster :) ?

fanzhangio: The `labelSelectorPath ` is optional, so get it removed when value is empty string in `parsingScaleScheme()`.
Tested the CRD yaml with scale and status in minikube cluster. It is validated and installed into minikube cluster.

PR #77, opened 2018-09-06
Generate versioned crd runtime object

fanzhangio:
This PR is for https://github.com/kubernetes-sigs/kubebuilder/issues/369

Adding an effort to `getCrds()` for generating a crd runtime object in  `zz_generated_crd.go` files in the same path of types.  This generation is default behavior. The auto generated crd runtime object can be exported and referred as external crd for controller.

Currently, support validation and sub resource.

Open to discussions

### DIFF
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -98,7 +98,7 @@ Usage:
 			if err := g.Do(); err != nil {
 				log.Fatal(err)
 			}
-			fmt.Printf("CRD files generated, files can be found under path %s.\n", g.OutputDir)
+			fmt.Printf("CRD files generated, files can be found under path %s\n", g.OutputDir)
 		},
 	}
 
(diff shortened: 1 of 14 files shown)


fanzhangio:
Can we skip `gosec` linter in CI ?

mengqiy:
CI should be passing this run.

droot:
@fanzhangio I have updated the issue with few thoughts about this approach https://github.com/kubernetes-sigs/kubebuilder/issues/369

Pl. take look.

droot:
@fanzhangio l have updated the original issue with my thoughts, let me know if you still think we need this PR. 
Also, another thing to note is `pkg/scaffold` has been moved to kubebuilder repo to streamline dev, so if we need to continue working on this, we will have to issue the PR in kubebuilder rep, so I am closing this PR at this moment.

PR #78, opened 2018-09-10
uses controller-runtime logging in scaffolding code

interma:
Fix: https://github.com/kubernetes-sigs/kubebuilder/issues/380

Thanks for the review.

### DIFF
--- a/pkg/scaffold/controller/controller.go
+++ b/pkg/scaffold/controller/controller.go
@@ -104,7 +104,6 @@ package {{ lower .Resource.Kind }}
 
 import (
 {{ if .Resource.CreateExampleReconcileBody }}	"context"
-	"log"
 	"reflect"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -120,8 +119,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
+)
+
+var logf = log.Log.WithName("controller")
 {{ else }}	"context"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -133,8 +136,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
-{{ end -}}
 )
+{{ end -}}
 
 /**
 * USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
@@ -245,7 +248,7 @@ func (r *Reconcile{{ .Resource.Kind }}) Reconcile(request reconcile.Request) (re
 	found := &appsv1.Deployment{}
 	err = r.Get(context.TODO(), types.NamespacedName{Name: deploy.Name, Namespace: deploy.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
-		log.Printf("Creating Deployment %s/%s\n", deploy.Namespace, deploy.Name)
+		logf.Info("Creating Deployment %s/%s\n", deploy.Namespace, deploy.Name)
 		err = r.Create(context.TODO(), deploy)
 		if err != nil {
 			return reconcile.Result{}, err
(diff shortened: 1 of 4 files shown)


pkg/scaffold/manager/cmd.go
```diff
@@ -43,44 +43,56 @@ var cmdTemplate = `{{ .Boilerplate }}
...
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"{{ .Repo }}/pkg/apis"
 	"{{ .Repo }}/pkg/controller"
 )
 
 func main() {
+	// Set logger
+	logf.SetLogger(logf.ZapLoggerTo(ginkgo.GinkgoWriter, true))
```
interma: for reviewer:
Should use `GinkgoWriter` here? Or other loggers?

DirectXMan12: this isn't a test.  we shouldn't be using ginkgo.  Just use `ZapLogger`.

pkg/scaffold/manager/cmd.go
```diff
@@ -43,44 +43,56 @@ var cmdTemplate = `{{ .Boilerplate }}
...
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{})
 	if err != nil {
-		log.Fatal(err)
+		log.Error(err, "Create manager failed")
+		return
 	}
 
-	log.Printf("Registering Components.")
+	log.V(1).Info("Registering Components.")
```
DirectXMan12: Keep these as `V(0)`

interma:
Fixed all comment issues.

@DirectXMan12 
I checked all scaffold code in `pkg/scaffold/` directory, nothing missing.
Is it need to check other directories?
Thanks.

pkg/scaffold/manager/cmd.go
```diff
@@ -43,44 +43,56 @@ var cmdTemplate = `{{ .Boilerplate }}
 package main
 
 import (
-	"log"
+	"github.com/onsi/ginkgo"
```
mengqiy: This line needs to be removed.

mengqiy:
It seems you have not cover all occurrences. e.g. https://github.com/kubernetes-sigs/controller-tools/blob/bfd3eefb7f22ed714b2acd38730c36992ee6af70/pkg/scaffold/controller/controller.go#L107
Can you please scan one more pass?

interma:
@mengqiy Sorry for missing them, I will do a double check later, thanks.

Btw: 
I am resigning CLA via as an employee now, will ping you after all done.

pkg/scaffold/manager/cmd.go
```diff
@@ -73,7 +73,7 @@ func main() {
 		return
 	}
 
-	log.V(1).Info("Registering Components.")
+	log.V(0).Info("Registering Components.")
```
DirectXMan12: can just do `log.Info` here

interma:
@droot @DirectXMan12 PTAL again, thanks.

If there is no problem, I will rebase commits later.

interma:
Ping @DirectXMan12 
Thanks.

pkg/scaffold/controller/controller.go
```diff
@@ -120,8 +119,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
+)
+
+var logf = log.Log.WithName("controller")
```
DirectXMan12: we usually do the other way around -- `logf` is the package alias, and `log` is the name of the actually base logger.

interma: Yes, I used `log` before, but it conflicted with "the import log pkg" in test file (I assumed it should not to use the controller-runtime logger here, right?).
e.g. 
https://github.com/kubernetes-sigs/controller-tools/blob/master/test/pkg/controller/firstmate/firstmate_controller_suite_test.go#L20.

So one of them should change and I choose to change here to `logf`.
Is there a better way? Or a more appropriate name instead of `logf`? 
Thanks.

interma: @mengqiy 
Yes, I noticed it before and the test code of my pr has been generated by the script. 
If used `var log = ...`, still have this confilction, reason see my previous comments.

Thanks.

DirectXMan12: It probably shouldn't be importing the standard "log" package -- just the controller runtime log package.

interma: @DirectXMan12 Yes, agree with you.
So what's your idea about "log" name confliction? Change `logf` to a more appropriate name?
Thanks.

mengqiy:
@interma The test project under `test` is now gone, can you please rebase and retry https://github.com/kubernetes-sigs/controller-tools/pull/78#discussion_r220244965?

interma:
@mengqiy Sure, I will rebase them.
But a little confused, I found all files under `pkg/scaffold` are removed and cannot find the new replacement of 
"pkg/scaffold/controller/controller.go" and "pkg/scaffold/manager/cmd.go".

So could you give me some hints to how to deal with it?
Thanks!

mengqiy:
> But a little confused, I found all files under pkg/scaffold are removed and cannot find the new replacement of
"pkg/scaffold/controller/controller.go" and "pkg/scaffold/manager/cmd.go".

They have been moved to the `kubebuilder` repo under `pkg/scaffold`. That means you will need to move part of the change in this PR to a separate PR against kubebuilder repo.

interma:
Thanks, Mengyi.
I will open a new PR in kubebuilder later.

interma:
New PR is here:
https://github.com/kubernetes-sigs/kubebuilder/pull/463

PR #80, opened 2018-09-11
fix CRD gen to not fail CRD validation

mengqiy:


.travis.yml
```diff
@@ -15,7 +15,7 @@ go_import_path: sigs.k8s.io/controller-tools
 install:
 - go get -u github.com/golang/dep/cmd/dep
 - go get -u golang.org/x/lint/golint
-- go get -u gopkg.in/alecthomas/gometalinter.v2 && gometalinter.v2 --install
+- curl -L https://git.io/vp6lP | sh
```
Liujingfang1: This change looks wired to me.

mengqiy: I have the same feeling :/
The old approach is broken and not working.
They now recommend using [this approach](https://github.com/alecthomas/gometalinter/blob/master/README.md#binary-releases).

PR #82, opened 2018-09-20
crd: add finalizers support in CRD generation

droot:
fixes https://github.com/kubernetes-sigs/controller-runtime/issues/151

### DIFF
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -295,6 +295,12 @@ func getCategoriesTag(c *types.Type) string {
 	return resource
 }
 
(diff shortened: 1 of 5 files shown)


pkg/internal/codegen/parse/util.go
```diff
@@ -370,9 +376,6 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
 			if len(path) < 2 {
 				return nil, fmt.Errorf(jsonPathError)
 			}
-			for _, s := range path {
-				fmt.Printf("\n[debug] %s", s)
-			}
```
fanzhangio: Thank you @droot . this is going to be deleted in my separate PR.

Issue #94, opened 2018-10-12
Add support for nested output layout


ichekrygin:
Currently, CRD `generate` emit CRD files in a "flat" layout structure: `foo_v1alpha1_bar.yaml`
It would be convenient/helpful to emit CRD files in a nested layout that resembles apis types organization:
`foo/v1alpha1/bar.yaml`

Issue #81, opened 2018-09-20
controler-gen fails if "group/version/types.go" is not an immediate child of "apis"


ichekrygin:
### Overview
Current project organization expects versioned types go files to be 2 directories deep under `apis` directory. For example:
```
./pkg/apis/foo/v1alpha1/types.go
```
Adding any additional top-level grouping (but still under `apis`) - breatk `controller-gen` functionality.
For example:
```
pkg/apis
├── addtoscheme_foo_v1alpha1.go
├── apis.go
└── foo
    ├── buz
    │   ├── group.go
    │   └── v1alpha1
    │       ├── bar_types.go
    │       ├── bar_types_test.go
    │       ├── doc.go
    │       ├── register.go
    │       ├── v1alpha1_suite_test.go
    │       └── zz_generated.deepcopy.go
    └── group.go

3 directories, 10 files

make manifests
go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
panic: Cannot get version for unversioned type github.com/user/foo/pkg/apis/foo/buz/v1alpha1.Bar

goroutine 1 [running]:
github.com/user/foo/vendor/sigs.k8s.io/controller-tools/pkg/internal/codegen/parse.GetVersion(0xc427252b00, 0xc4218e0ad4, 0x3, 0xc4218e0ad4, 0xc42728ea50)
        /home/illya/go/src/github.com/user/foo/vendor/sigs.k8s.io/controller-tools/pkg/internal/codegen/parse/util.go:218 +0x112
github.com/user/foo/vendor/sigs.k8s.io/controller-tools/pkg/internal/codegen/parse.(*APIs).parseIndex(0xc4272878c0)
        /home/illya/go/src/github.com/user/foo/vendor/sigs.k8s.io/controller-tools/pkg/internal/codegen/parse/index.go:68 +0x1ec
github.com/user/foo/vendor/sigs.k8s.io/controller-tools/pkg/internal/codegen/parse.NewAPIs(0xc42727b020, 0xc4202554a0, 0xc42027b600, 0xc, 0xc42015b913, 0x1f, 0xc42015b7c0)
        /home/illya/go/src/github.com/user/foo/vendor/sigs.k8s.io/controller-tools/pkg/internal/codegen/parse/parser.go:69 +0xc7
[...]

PR #95, opened 2018-10-12
Improve CRD Group handling in codegen

ichekrygin:
### Overview
This PR contains a few improvements to CRD group handling.

#### Output Layout
Add support for nested output layout via command line `-nested` boolean flag, which allows emitting CRD files in a nested `group/version/kind.yaml` layout, vs default flat `group_version_kind.yaml`
- Tracking #94

#### Annotated Group Name
Currently, CRD group name is retrieved using `parent` directory to the versioned type, i.e. 
`./pkg/apis/foo/v1alpha1/bar.go` - results in the `groupName=foo`.

The main limitation of this approach, it is not allowing for more complicated group structure as reported in #81. For example: `./pkg/apis/foo/buz/v1alpha1/bar.go` - is not supported, and generate reports an error

To address this functionality issue, there are two changes:
- Add support for using comment type annotation for group name: `// +groupName=foo.bar`.
If the annotated group value is not provided - default to the original `parent` dir model.
- Allow nested groups under `apis` directory 

**Note***: both of these changes are additive and all original behavior is preserved and is a default functionality.

### DIFF
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -105,6 +105,7 @@ Usage:
 	f := cmd.Flags()
 	f.StringVar(&g.RootPath, "root-path", "", "working dir, must have PROJECT file under the path or parent path if domain not set")
 	f.StringVar(&g.OutputDir, "output-dir", "", "output directory, default to 'config/crds' under root path")
+	f.BoolVar(&g.NestedOutput, "nested", false, "nested output layout: group/version/kind.yaml")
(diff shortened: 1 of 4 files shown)


droot:
@ichekrygin sorry, I dropped the ball on this one. We will be adding support for multi-version + version-conversion to CRD generation soon and this will have significant overlap with that functionality for ex. multiple versions will be part of the same CRD etc, so would like to wait on this change to ensure we aren't introducing complexity for the new features.

ichekrygin:
@droot , np, thx for the update.

PR #97, opened 2018-10-22
Refactor pkg/generate

fanzhangio:
This PR is the first step of crd generation refactor focusing on moving `rbac`, `webhook` from `pkg/generate` to `pkg/` to align with `crd` in the same package level.

**Move below interface in separate PR**
Add a generic `Generator` interface in `pkg/internal/codegen` which is on top of `gengo` for all generation usages in controller-tool


mengqiy:
There seems to be a problem with travis CI...
Let's try close and reopen.

droot:
@fanzhangio I am having trouble reviewing the PR because it is very large PR. would have been nice to split it into multiple commits. Is it possible to do so ?

droot:
@fanzhangio Are you planning to split it or you want me to review as-is ?

droot:
@fanzhangio lets chat offline how do we want to move forward with this PR.

fanzhangio:
@droot Sure, I marked it with `WIP` for better discussion.

cmd/controller-gen/main.go
```diff
@@ -22,7 +22,7 @@ import (
 	"path/filepath"
 
 	"github.com/spf13/cobra"
-	crdgenerator "sigs.k8s.io/controller-tools/pkg/crd/generator"
+	"sigs.k8s.io/controller-tools/pkg/generate/crd"
```
droot: `generate` is not a good pkg name as per Go guidelines. Given that we have moved scaffolding pkgs to `kubebuilder` repo, it will be safe to assume that the pkgs that live here are related to generation. So may be move the others `rbac` and `webhook` under `pkg` ?

cmd/controller-gen/main.go
```diff
@@ -84,7 +84,7 @@ Usage:
 }
 
 func newCRDCmd() *cobra.Command {
-	g := &crdgenerator.Generator{}
+	o := &crd.Options{}
```
droot: If the pkg name becomes `crd` then `Generator` seems like more appropriate to capture generator options ?

pkg/generate/crd/crd.go
```diff
@@ -26,58 +26,50 @@ import (
...
-	crdutil "sigs.k8s.io/controller-tools/pkg/crd/util"
 	"sigs.k8s.io/controller-tools/pkg/internal/codegen"
 	"sigs.k8s.io/controller-tools/pkg/internal/codegen/parse"
-	"sigs.k8s.io/controller-tools/pkg/util"
+	util "sigs.k8s.io/controller-tools/pkg/util"
 )
 
-// Generator generates CRD manifests from API resource definitions defined in Go source files.
-type Generator struct {
+// Options receive optional arguments from command line
```
droot: Sorry, but the new comment doesn't give much useful information.

pkg/generate/crd/crd.go
```diff
@@ -26,58 +26,50 @@ import (
...
+type Options struct {
 	RootPath          string
 	OutputDir         string
 	Domain            string
 	Namespace         string
 	SkipMapValidation bool
-
+	apisPkg           string
```
droot: The previous comment was very useful for code understanding. Is there a reason we are deleting that ?

pkg/generate/crd/crd.go
```diff
@@ -86,70 +78,57 @@ func (c *Generator) ValidateAndInitFields() error {
...
 
 	return nil
 }
 
-// Do manages CRD generation.
-func (c *Generator) Do() error {
-	arguments := args.Default()
-	b, err := arguments.NewBuilder()
+// Run generates CRD and stores CRD into output dir.
+func (c *Options) Run() error {
```
droot: I found the previous abstractions easier to grok. 
Generator takes input options to generate CRD vs Options having Run method to generate the CRD. I find previous one more intuitive.

cmd/controller-gen/main.go
fanzhangio: Agree. So, I will move all other packages, like `rbac`, `webhook` into `pkg` in this PR.

cmd/controller-gen/main.go
fanzhangio: Make sense. So we can keep generator approach, but just split the generation procedure into some exposed APIs. I also wanted to have a more generic `Generator` upfront, which wraps `gengo` for all kinds of generation in this repo. This kind of `Generator` can live in `pkg/internal/codegen`

pkg/generate/crd/crd.go
fanzhangio: Make sense.

pkg/generate/crd/crd.go
fanzhangio: OK. I will keep the comments.

pkg/generate/crd/crd.go
fanzhangio: I see. I will rebase and update the PR. How about we do it by a series of changes. First, I just focus on moving packages in this PR, and then refactor the `Generator` ?

pkg/internal/codegen/generator.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// NameSystems returns the name system used by the generators in this package.
+// e.g. black-magic
+func NameSystems() namer.NameSystems {
+	return namer.NameSystems{
+		"public": namer.NewPublicNamer(1),
+		"raw":    namer.NewRawNamer("", nil),
+	}
+}
```
droot: Can we move this in a separate PR ? preferrably with use-case in existing code to demonstrate the utility ?

fanzhangio: Sure. Thanks

mengqiy:
Looks like the comments have been resolved.
/lgtm

PR #100, opened 2018-10-25
Special case 'unstructured'

Kashomon:
Add a special case for the unstructured.Unstructured type from apimachinery. This can
be arbitrary JSON/YAML, so, like ObjectMeta, set type to just 'Object'
This is needed for the ClusterBundle, which relies heavily on unstructured objects.

### DIFF
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -130,7 +130,7 @@ func (b *APIs) getTime() string {
 }`
 }
 
-func (b *APIs) getMeta() string {
+func (b *APIs) objSchema() string {
 	return `v1beta1.JSONSchemaProps{
     Type:   "object",
 }`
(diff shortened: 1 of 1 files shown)


Kashomon:
On further thought, unstructured.Unstructured is not quite what I want. What I really want is to handle `map[string]interface{}` -- aka arbitrary JSON/YAML

When I run the command I get

```
2018/10/24 20:53:45 Unknown supported Kind Interface
```

I could add 

```go
	case types.Interface:
		v, s = v1beta1.JSONSchemaProps{
			Type: "object",
		}, b.objSchema()
```

In that same file, but it's sort of a hack.

Kashomon:
I've gone ahead and added this suggestion to the PR (since that's what I need). I can remove the unstructured stuff I added initially.

pkg/internal/codegen/parse/crd.go
```diff
@@ -171,6 +176,10 @@ func (b *APIs) typeToJSONSchemaProps(t *types.Type, found sets.String, comments
 		v, s = b.typeToJSONSchemaProps(t.Elem, found, comments, false)
 	case types.Alias:
 		v, s = b.typeToJSONSchemaProps(t.Underlying, found, comments, false)
+	case types.Interface:
```
Kashomon: Can probably remove this.

Kashomon:
I've now removed the interface case.

Issue #103, opened 2018-11-16
Add support for additionalPrinterColums


pmorie:
Related: kubernetes-sigs/kubebuilder#447

It should be possible to generate `additionalPrinterColumns` for CRDs, possible example:

```go
//+kubebuilder:printcolumn:ready:type=string
//+kubebuilder:printcolumn:ready:description="Is the cluster ready?"
//+kubebuilder:printcolumn:ready:jsonpath=".status.conditions[?(@.type==\"Ready\")].status"
//+kubebuilder:printcolumn:ready:format=bla
//+kubebuilder:printcolumn:ready:priority=1
```

droot:
I would suggest using compressed form like we do for RBAC annotation. 
```Go
// +kubebuilder:printcolumn:ready=type=<type>,description=<desc>,JSONPath:<.spec.Name>?priority=<0|1>,?format=<format>
```

sohankunkerkar:
@droot you can close this issue.

PR #105, opened 2018-11-26
logic to control additional printer columns

sohankunkerkar:
#103

### DIFF
--- a/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
+++ b/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
@@ -67,8 +67,9 @@ type ToyStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
 // +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=
-// +kubebuilder:printcolumn:name="toy",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="descr1"
-// +kubebuilder:printcolumn:name="abc",type="string",JSONPath="status",description="descr2"
+// +kubebuilder:printcolumn:name="toy",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="descr1",format="date",priority=3
+// +kubebuilder:printcolumn:name="abc",type="integer",JSONPath="status",description="descr2",format="int32",priority=1
+// +kubebuilder:printcolumn:name="service",type="string",JSONPath=".status.conditions.ready",description="descr3",format="byte",priority=2
 type Toy struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
+++ b/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
@@ -67,6 +67,8 @@ type ToyStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
 // +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=
+// +kubebuilder:printcolumn:name="toy",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="descr1"
+// +kubebuilder:printcolumn:name="abc",type="string",JSONPath="status",description="descr2"
 type Toy struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -113,10 +113,10 @@ func (b *APIs) parseCRDs() {
 							resource.CRD.Spec.Subresources.Scale.LabelSelectorPath = &labelSelctor
 						}
 					}
-					if IsPrintColumn(resource.Type) {
+					if hasPrintColumn(resource.Type) {
 						result, err := parsePrintColumnParams(resource.Type)
 						if err != nil {
-							log.Fatalf("failed in parsing CRD, error: %v", err.Error())
+							log.Fatalf("failed to parse printcolumn annotations, error: %v", err.Error())
 						}
 						resource.CRD.Spec.AdditionalPrinterColumns = result
 					}
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -113,7 +113,13 @@ func (b *APIs) parseCRDs() {
 							resource.CRD.Spec.Subresources.Scale.LabelSelectorPath = &labelSelctor
 						}
 					}
-
+					if IsPrintColumn(resource.Type) {
+						result, err := parsePrintColumnParams(resource.Type)
+						if err != nil {
+							log.Fatalf("failed in parsing CRD, error: %v", err.Error())
+						}
+						resource.CRD.Spec.AdditionalPrinterColumns = result
+					}
 					if len(resource.ShortName) > 0 {
 						resource.CRD.Spec.Names.ShortNames = []string{resource.ShortName}
 					}
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -117,8 +117,8 @@ func IsRBAC(t *types.Type) bool {
 	return false
 }
 
-// IsPrintColumn returns true if t has a +printcolumn or +kubebuilder:printcolumn tag
-func IsPrintColumn(t *types.Type) bool {
+// hasPrintColumn returns true if t has a +printcolumn or +kubebuilder:printcolumn annotation.
+func hasPrintColumn(t *types.Type) bool {
 	for _, c := range t.CommentLines {
 		if strings.Contains(c, "+printcolumn") || strings.Contains(c, "+kubebuilder:printcolumn") {
 			return true
@@ -445,7 +445,7 @@ func helperPrintColumn(parts string, comment string) (v1beta1.CustomResourceColu
 				"keys [name=<name>,type=<type>,description=<descr>,format=<format>] "+
 				"Got string: [%s]", parts)
 		}
-		if key == "name" || key == "type" || key == "JSONPath" {
+		if key == printColumnName || key == printColumnType || key == printColumnPath {
 			count++
 		}
 		switch key {
@@ -455,7 +455,7 @@ func helperPrintColumn(parts string, comment string) (v1beta1.CustomResourceColu
 			if value == "integer" || value == "number" || value == "string" || value == "boolean" {
 				config.Type = value
 			} else {
-				return v1beta1.CustomResourceColumnDefinition{}, fmt.Errorf("Please enter the right value for type")
+				return v1beta1.CustomResourceColumnDefinition{}, fmt.Errorf("invalid value for %s printcolumn", printColumnType)
 			}
 		case printColumnFormat:
 			if config.Type == "integer" && (value == "int32" || value == "int64") {
@@ -465,7 +465,7 @@ func helperPrintColumn(parts string, comment string) (v1beta1.CustomResourceColu
 			} else if config.Type == "string" && (value == "byte" || value == "date" || value == "date-time" || value == "password") {
 				config.Format = value
 			} else {
-				return v1beta1.CustomResourceColumnDefinition{}, fmt.Errorf("Please enter the right value for format")
+				return v1beta1.CustomResourceColumnDefinition{}, fmt.Errorf("invalid value for %s printcolumn", printColumnFormat)
 			}
 		case printColumnPath:
 			config.JSONPath = value
@@ -473,7 +473,7 @@ func helperPrintColumn(parts string, comment string) (v1beta1.CustomResourceColu
 			i, err := strconv.Atoi(value)
 			v := int32(i)
 			if err != nil {
-				return v1beta1.CustomResourceColumnDefinition{}, err
+				return v1beta1.CustomResourceColumnDefinition{}, fmt.Errorf("invalid value for %s printcolumn", printColumnPri)
 			}
 			config.Priority = v
 		case printColumnDescr:
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -412,8 +412,8 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
 	return nil, fmt.Errorf(jsonPathError)
 }
 
-// PrintColumnKV parses key-value string formatted as "foo=bar" and returns key and value.
-func PrintColumnKV(s string) (key, value string, err error) {
+// printColumnKV parses key-value string formatted as "foo=bar" and returns key and value.
+func printColumnKV(s string) (key, value string, err error) {
 	kv := strings.SplitN(s, "=", 2)
 	if len(kv) != 2 {
 		err = fmt.Errorf("invalid key value pair")
@@ -439,7 +439,7 @@ func helperPrintColumn(parts string, comment string) (v1beta1.CustomResourceColu
 		fmt.Printf("\n[debug] %s", s)
 	}
 	for _, elem := range strings.Split(parts, ",") {
-		key, value, err := PrintColumnKV(elem)
+		key, value, err := printColumnKV(elem)
 		if err != nil {
 			return v1beta1.CustomResourceColumnDefinition{},
 				fmt.Errorf("//+kubebuilder:printcolumn: tags must be key value pairs.Expected "+
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -441,9 +441,10 @@ func helperPrintColumn(parts string, comment string) (v1beta1.CustomResourceColu
 	for _, elem := range strings.Split(parts, ",") {
 		key, value, err := PrintColumnKV(elem)
 		if err != nil {
-			log.Fatalf("//+kubebuilder:printcolumn: tags must be key value pairs.  Expected "+
-				"keys [name=<name>,type=<type>,description=<descr>,format=<format>] "+
-				"Got string: [%s]", parts)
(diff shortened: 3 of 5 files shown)


sohankunkerkar:
cc/ @droot

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
 	}
 	return nil, fmt.Errorf(jsonPathError)
 }
+
+// ready subresource requires name,type,JSONPath fields and rest of the field are optional
```
droot: We don't need name like `ready` anymore because we decided to use one-line specification for each print column. In the linked issue, `ready` was used to group specifications for a column because column specifications were expressed in multiple lines. 
So lets remove the requirement for additional name like `ready`.

So annotation should read as:
```
// +kubebuilder:printcolumn:name=<name1>,type=<type>,description=<desc>,JSONPath:<.spec.Name>,priority=<0|1>,format=<format>
// +kubebuilder:printcolumn:name=<name2>,type=<type>,description=<desc>,JSONPath:<.spec.Name>,priority=<0|1>,format=<format>
```

pkg/internal/codegen/parse/util.go
```diff
@@ -144,6 +160,20 @@ func HasSubresource(t *types.Type) bool {
...
 }
 
+// hasReadyStatusTag returns true if t is an APIResource annotated with
+// +kubebuilder:printcolumn:ready
+func hasReadyStatusTag(t *types.Type) bool {
+	if !IsPrintColumn(t) {
+		return false
+	}
+	for _, c := range t.CommentLines {
+		if strings.Contains(c, "+kubebuilder:printcolumn:ready") {
```
droot: This is no longer required. Read my comment about `ready` in the parse method.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
 	}
 	return nil, fmt.Errorf(jsonPathError)
 }
+
+// ready subresource requires name,type,JSONPath fields and rest of the field are optional
+// +kubebuilder:printcolumn:ready:name=<name>,type=<type>,description=<desc>,JSONPath:<.spec.Name>,priority=<0|1>,format=<format>
+func parsePrintColumnParams(t *types.Type) v1beta1.CustomResourceColumnDefinition {
+	config := v1beta1.CustomResourceColumnDefinition{}
+	for _, comment := range t.CommentLines {
+		if strings.Contains(comment, "+kubebuilder:printcolumn:ready") {
```
droot: `ready` is not required. here and in other places.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+			parts := strings.Replace(comment, "+kubebuilder:printcolumn:ready:", "", -1)
+			part := strings.Split(parts, ",")
+			if len(part) < 3 {
+				fmt.Println(fmt.Errorf("Expected required fields in this context: %s", comment))
+			}
```
droot: I think we have parsing key-value pairs helpers under `utils` pkg somewhere. See if we can use those ?

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+				case "printColumnPriority":
+					i, err := strconv.Atoi(arr[1])
+					v := int32(i)
+					if err != nil {
+						log.Fatalf("Could not parse int from %s", comment)
+					}
+					config.Priority = v
+				case "printColumnDescr":
+					config.Description = arr[1]
+				default:
```
droot: Typo: get rid of quotes for all the case statements.  "printX" --> printX

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+					log.Fatalf("Unsupport validation: %s", comment)
+
+				}
+
+			}
```
droot: You can can have helper fn for parsing a single column. That will make it unit-test friendly.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+					i, err := strconv.Atoi(arr[1])
+					v := int32(i)
+					if err != nil {
+						log.Fatalf("Could not parse int from %s", comment)
+					}
+					config.Priority = v
+				case "printColumnDescr":
+					config.Description = arr[1]
+				default:
+					log.Fatalf("Unsupport validation: %s", comment)
```
droot: return error for this case, let the caller issue fatal call. Look at my other comment about capturing this logic in separate fn to make it test-friendly. That will require returning error.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+			}
+			for _, s := range part {
+				arr := strings.Split(s, "=")
+				switch arr[0] {
+				case "printColumnName":
+					config.Name = arr[1]
+				case "printColumnType":
+					config.Type = arr[1]
+				case "printColumnFormat":
+					config.Format = arr[1]
```
droot: There are some valid allowed values specified in the [spec](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#additional-printer-columns). We should validate against those.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+				default:
+					log.Fatalf("Unsupport validation: %s", comment)
+
+				}
+
+			}
+		}
+	}
+	return config
+}
```
droot: We should use `testData` project to add tests. Add tests to cover basic scenario.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +425,45 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+					if err != nil {
+						log.Fatalf("Could not parse int from %s", comment)
+					}
+					config.Priority = v
+				case "printColumnDescr":
+					config.Description = arr[1]
+				default:
+					log.Fatalf("Unsupport validation: %s", comment)
+
+				}
```
droot: Please look at the [spec](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#additional-printer-columns) and ensure all the required fields for columns are specified.

pkg/internal/codegen/parse/util.go
```diff
@@ -30,10 +30,16 @@ import (
...
 
 const (
-	specReplicasPath   = "specpath"
-	statusReplicasPath = "statuspath"
-	labelSelectorPath  = "selectorpath"
-	jsonPathError      = "invalid scale path. specpath, statuspath key-value pairs are required, only selectorpath key-value is optinal. For example: // +kubebuilder:subresource:scale:specpath=.spec.replica,statuspath=.status.replica,selectorpath=.spec.Label"
+	specReplicasPath    = "specpath"
+	statusReplicasPath  = "statuspath"
+	labelSelectorPath   = "selectorpath"
+	jsonPathError       = "invalid scale path. specpath, statuspath key-value pairs are required, only selectorpath key-value is optinal. For example: // +kubebuilder:subresource:scale:specpath=.spec.replica,statuspath=.status.replica,selectorpath=.spec.Label"
```
droot: I did not see these four being used anywhere ?

pkg/internal/codegen/parse/util.go
sohankunkerkar: yeah, I've seen that function here https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/generate/internal/util.go#L50 However, I can not reuse it because the package is named `internal`. It's a golang feature to provide encapsulation of the package internals. One workaround here is to copy the same function as is. Let me know your thoughts on this information.

pkg/internal/codegen/parse/util.go
pmorie: It's gofmt :)

pkg/internal/codegen/parse/util.go
sohankunkerkar: Ack, I will make the necessary changes.

pkg/internal/codegen/parse/util.go
sohankunkerkar: yeah

pkg/internal/codegen/parse/util.go
sohankunkerkar: ack

pkg/internal/codegen/parse/util.go
sohankunkerkar: I can't reuse  `ParseKV` function listed under `pkg/generate/internal` since there are a few test cases in this feature which are not handled in the existing KV function. Ultimately, I need to create a separate function to handle this logic.

pkg/internal/codegen/parse/util.go
sohankunkerkar: Done!

pkg/internal/codegen/parse/util.go
sohankunkerkar: ack.

pkg/internal/codegen/parse/util.go
sohankunkerkar: Yeah, will create a few test cases before pushing the next code.

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -67,6 +67,8 @@ type ToyStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
 // +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=
+// +kubebuilder:printcolumn:name="toy",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="descr1"
+// +kubebuilder:printcolumn:name="abc",type="string",JSONPath="status",description="descr2"
```
droot: We should probably add few covering `priority` and other fields ?

pkg/internal/codegen/parse/crd.go
```diff
@@ -113,7 +113,13 @@ func (b *APIs) parseCRDs() {
 							resource.CRD.Spec.Subresources.Scale.LabelSelectorPath = &labelSelctor
 						}
 					}
-
+					if IsPrintColumn(resource.Type) {
```
droot: Can we rename `IsPrintColumn` to `hasPrintColumns` to keep it consistent with other annotations above ?

pkg/internal/codegen/parse/crd.go
```diff
@@ -113,7 +113,13 @@ func (b *APIs) parseCRDs() {
...
+						result, err := parsePrintColumnParams(resource.Type)
+						if err != nil {
+							log.Fatalf("failed in parsing CRD, error: %v", err.Error())
```
droot: Update the error message to say "failed to parse printcolumn annotations,....." ?

pkg/internal/codegen/parse/util.go
```diff
@@ -111,6 +117,16 @@ func IsRBAC(t *types.Type) bool {
 	return false
 }
 
+// IsPrintColumn returns true if t has a +printcolumn or +kubebuilder:printcolumn tag
+func IsPrintColumn(t *types.Type) bool {
```
droot: IsPrintColumn --> hasPrintColumn ?

pkg/internal/codegen/parse/util.go
```diff
@@ -111,6 +117,16 @@ func IsRBAC(t *types.Type) bool {
 	return false
 }
 
+// IsPrintColumn returns true if t has a +printcolumn or +kubebuilder:printcolumn tag
```
droot: s/tag/annotation

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +411,96 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
 	}
 	return nil, fmt.Errorf(jsonPathError)
 }
+
+// PrintColumnKV parses key-value string formatted as "foo=bar" and returns key and value.
+func PrintColumnKV(s string) (key, value string, err error) {
```
droot: May be rename it to `parseKV` because I am assuming this will removed once the utils pkg is available to be used here ? Also, add a TODO with comment this is duplicated and will be removed in a followup PR.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +411,96 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+	config := v1beta1.CustomResourceColumnDefinition{}
+	var count int
+	part := strings.Split(parts, ",")
+	if len(part) < 3 {
+		return v1beta1.CustomResourceColumnDefinition{}, fmt.Errorf(printColumnError)
+	}
+
+	for _, s := range part {
+		fmt.Printf("\n[debug] %s", s)
+	}
```
droot: I am assuming this will be deleted.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +411,96 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+	for _, elem := range strings.Split(parts, ",") {
+		key, value, err := PrintColumnKV(elem)
+		if err != nil {
+			log.Fatalf("//+kubebuilder:printcolumn: tags must be key value pairs.  Expected "+
```
droot: we should return err here.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +411,96 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+				"keys [name=<name>,type=<type>,description=<descr>,format=<format>] "+
+				"Got string: [%s]", parts)
+		}
+		if key == "name" || key == "type" || key == "JSONPath" {
```
droot: use the constants since they are already defined ?

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +411,96 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+				config.Format = value
+			} else if config.Type == "number" && (value == "float" || value == "double") {
+				config.Format = value
+			} else if config.Type == "string" && (value == "byte" || value == "date" || value == "date-time" || value == "password") {
+				config.Format = value
+			} else {
+				return v1beta1.CustomResourceColumnDefinition{}, fmt.Errorf("Please enter the right value for format")
+			}
+		case printColumnPath:
+			config.JSONPath = value
```
droot: Not in this PR, but it will be good to add add validation for the json-path since have the spec for the type already.

pkg/internal/codegen/parse/util.go
```diff
@@ -395,3 +411,96 @@ func parseScaleParams(t *types.Type) (map[string]string, error) {
...
+		case printColumnPri:
+			i, err := strconv.Atoi(value)
+			v := int32(i)
+			if err != nil {
+				return v1beta1.CustomResourceColumnDefinition{}, err
```
droot: we should add some context to the error in this case, something like "invalid value for priority %s for printcolumn"

pkg/internal/codegen/parse/util_test.go
```diff
@@ -123,3 +124,63 @@ func TestParseScaleParams(t *testing.T) {
 		}
 	}
 }
+func Test_parsePrintColumnParams(t *testing.T) {
```
droot: s/Test_parsePrintColumnParams/TestParsePrintColumnParams

pkg/internal/codegen/parse/util_test.go
```diff
@@ -123,3 +124,63 @@ func TestParseScaleParams(t *testing.T) {
...
+			tag:      "+kubebuilder:printcolumn:name=toy,type=string",
+			expected: []v1beta1.CustomResourceColumnDefinition{},
+			parseErr: fmt.Errorf(printColumnError),
+		},
+		{
+			name:     "one requied parameter is missing",
+			tag:      "+kubebuilder:printcolumn:name=toy,type=string,description=.status.conditions[?(@.type==\"Ready\")].status",
+			expected: []v1beta1.CustomResourceColumnDefinition{},
+			parseErr: fmt.Errorf(printColumnError),
+		},
```
droot: Thanks for adding unit-test here. We can add for invalid priority and format as well.

pkg/internal/codegen/parse/util.go
sohankunkerkar: I will include this feature in a separate PR if that's okay with you.

sohankunkerkar: >  Also, add a TODO with comment this is duplicated and will be removed in a followup PR.

This function is not duplicated, as I said before I made changes in the logic `kv := strings.SplitN(s, "=", 2)` to support test cases for `JSONPath` printcolumn which are already listed in the test file.

pkg/internal/codegen/parse/util.go
sohankunkerkar: I will include this feature in a separate PR if that's okay with you.

pkg/internal/codegen/parse/util_test.go
sohankunkerkar: Ack. Done!

pkg/internal/codegen/parse/util.go
sohankunkerkar: follow my previous comments.

pkg/internal/codegen/parse/util.go
droot: ok. In current form, this is being exported (public function). Lets rename (lowercase) it to `parsePrintColumnKV` or `printColumnKV` to make it private.

sohankunkerkar:
Done!

droot:
Awesome! Looks all good. Thank you for your contribution. This is a very useful feature.

sohankunkerkar:
Fixed https://github.com/kubernetes-sigs/kubebuilder/issues/447

PR #106, opened 2018-11-28
Logic to support resource annotation for a given API type

sohankunkerkar:
Creating this PR to support `resource` annotation.
https://github.com/kubernetes-sigs/kubebuilder/issues/443


sohankunkerkar:
cc/ @droot

pkg/internal/codegen/parse/index.go
```diff
@@ -184,28 +183,38 @@ type resourceTags struct {
 }
 
 // ParseResourceTag parses the tags in a "+resource=" comment into a resourceTags struct
-func parseResourceTag(tag string) resourceTags {
+func parseResourceTag(t *types.Type) (resourceTags, error) {
```
DirectXMan12: I'd keep this as two separate functions -- it improves the testability and makes it easier to reuse.

pkg/internal/codegen/parse/index.go
```diff
@@ -184,28 +183,38 @@ type resourceTags struct {
...
-		//	result.REST = value
-		case "path":
-			result.Resource = value
-		//case "strategy":
-		//	result.Strategy = value
-		case "shortName":
-			result.ShortName = value
+	for _, comment := range t.CommentLines {
+		if strings.Contains(comment, "+kubebuilder:resource") {
+			path := strings.Replace(comment, "+kubebuilder:resource:", "", -1)
```
DirectXMan12: we've got `internal.GetAnnotation` for this purpose.

pkg/internal/codegen/parse/index.go
```diff
@@ -184,28 +183,38 @@ type resourceTags struct {
...
+			for _, elem := range strings.Split(path, ",") {
+				kv := strings.Split(elem, "=")
```
DirectXMan12: we've got ParseKV for this

pkg/internal/codegen/parse/index.go
```diff
@@ -184,28 +183,38 @@ type resourceTags struct {
...
-		case "path":
-			result.Resource = value
-		//case "strategy":
-		//	result.Strategy = value
-		case "shortName":
-			result.ShortName = value
+	for _, comment := range t.CommentLines {
+		if strings.Contains(comment, "+kubebuilder:resource") {
+			path := strings.Replace(comment, "+kubebuilder:resource:", "", -1)
+			for _, elem := range strings.Split(path, ",") {
```
DirectXMan12: add a utility like `GetAnnotation` or `ParseKV` for this if there's not already one (`ParseAnnotationToMap`?)

pkg/internal/codegen/parse/crd.go
```diff
@@ -120,6 +120,14 @@ func (b *APIs) parseCRDs() {
 						}
 						resource.CRD.Spec.AdditionalPrinterColumns = result
 					}
+					if IsAPIResource(resource.Type) {
+						result, err := parseResourceTag(resource.Type)
+						if err != nil {
+							log.Fatalf("failed to parse resource annotations, error: %v", err.Error())
+						}
+						resource.CRD.ObjectMeta.Name = result.Resource + "." + resource.Group + "." + resource.Domain
+						resource.CRD.Spec.Names.Plural = result.Resource
```
DirectXMan12: this seems like it should be done when we parse the original resource.

sohankunkerkar: I have added it after we parse the original resource so that if it will override the existing values (if any).

pkg/internal/codegen/parse/crd.go
```diff
@@ -65,14 +65,24 @@ func (b *APIs) parseCRDs() {
 							Group:   fmt.Sprintf("%s.%s", resource.Group, resource.Domain),
 							Version: resource.Version,
 							Names: v1beta1.CustomResourceDefinitionNames{
-								Kind:   resource.Kind,
-								Plural: resource.Resource,
+								Kind:       resource.Kind,
+								Plural:     resource.Resource,
+								ShortNames: []string{resource.ShortName},
```
sohankunkerkar: To fix this issue https://github.com/kubernetes-sigs/kubebuilder/issues/404

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -70,6 +70,7 @@ type ToyStatus struct {
 // +kubebuilder:printcolumn:name="toy",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="descr1",format="date",priority=3
 // +kubebuilder:printcolumn:name="abc",type="integer",JSONPath="status",description="descr2",format="int32",priority=1
 // +kubebuilder:printcolumn:name="service",type="string",JSONPath=".status.conditions.ready",description="descr3",format="byte",priority=2
+// +kubebuilder:resource:path=services,shortName=ty
```
droot: We should add another type where `resource` annotation is not added to cover both the scenarios because existing projects do not have resource annotation and I am afraid this change will break `crd generation` for them.

pkg/internal/codegen/parse/index.go
```diff
@@ -208,6 +209,41 @@ func parseResourceTag(tag string) resourceTags {
...
+	for _, comment := range t.CommentLines {
+		anno := general.GetAnnotation(comment, "kubebuilder:resource")
+		if len(anno) == 0 {
+			continue
+		}
+		for _, elem := range strings.Split(anno, ",") {
+			key, value, err := general.ParseKV(elem)
+			if err != nil {
+				return resourceTags{}, fmt.Errorf("// +kubebuilder:resource: tags must be key value pairs.  Expected "+
+					"keys [path=<subresourcepath>] "+
```
droot: subresourcepath ?

pkg/internal/codegen/parse/index.go
```diff
@@ -208,6 +209,41 @@ func parseResourceTag(tag string) resourceTags {
...
+					"Got string: [%s]", comment)
+			}
+			for _, s := range anno {
+				fmt.Printf("\n[debug] %v", s)
+			}
```
droot: get rid of these before the final commit.

pkg/internal/codegen/parse/index.go
```diff
@@ -208,6 +209,41 @@ func parseResourceTag(tag string) resourceTags {
...
 }
 
+// parseResourceAnnotation parses the tags in a "+resource=" comment into a resourceTags struct.
+func parseResourceAnnotation(t *types.Type) (resourceTags, error) {
+	result := resourceTags{}
+	for _, comment := range t.CommentLines {
+		anno := general.GetAnnotation(comment, "kubebuilder:resource")
+		if len(anno) == 0 {
+			continue
+		}
```
droot: So if users have specified resource tag in multiple lines, we will pick the values from the last one and will not warn users about it. 

As per my understanding, resource annotation should only exists once per type, so if we encounter multiple, we should fail. Also, we can restructure this logic to extract resourceAnnotation once through a helper fn (that should return error if multiple annotations encountered) and then call the parsing only once ?

pkg/internal/codegen/parse/crd.go
```diff
@@ -65,14 +65,24 @@ func (b *APIs) parseCRDs() {
...
 							},
 							Validation: &v1beta1.CustomResourceValidation{
 								OpenAPIV3Schema: &resource.JSONSchemaProps,
 							},
 						},
 					}
+
+					result, err := parseResourceAnnotation(resource.Type)
+					if err != nil {
```
droot: this code assumes the annotation will always exists ? which isn't the case for existing projects.

DirectXMan12:
I'll try and do another pass on this shortly, but I agree with @droot's comments for the time being.

pkg/general/util.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package internal
+package general
```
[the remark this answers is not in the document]
sohankunkerkar: Yeah, I see what you're saying with you, however, there is already one `internal` folder inside the `pkg` folder
```
[skunkerk@skunkerk pkg]$ ls
crd  general  generate  internal  util
```
I know it doesn't make sense if we want to keep that package private but for the time being, I thought to rename it to something like`general`. Let me know if you have any suggestions for this.

pkg/internal/codegen/parse/index.go
sohankunkerkar: My bad. I will change that.:)

pkg/internal/codegen/parse/index.go
sohankunkerkar: Sure thing!

pkg/internal/codegen/parse/index.go
sohankunkerkar: Yeah, I will make those changes. Thanks for clarifying that information.

pkg/internal/codegen/parse/crd.go
sohankunkerkar: If you look at this linehttps://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/crd.go#L41
which concludes the following thing: https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/util.go#L59
If I'm not wrong, it assumes that every CRD will have a `resource` annotation. That's the only reason why I didn't include any condition for this code. Let me know your inputs for this.

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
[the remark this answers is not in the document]
sohankunkerkar: Do you want me to add this change after this PR gets merged? I would like to understand some more information regarding this feedback.

sohankunkerkar: > We should add another type where `resource` annotation is not added to cover both the scenarios.

Could you please explain this a bit more in detail?

pkg/general/util.go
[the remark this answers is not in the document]
DirectXMan12: you could just move the contents of `general` into the `pkg/internal` folder.

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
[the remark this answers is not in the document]
DirectXMan12: I'm not certain explicitly generating that annotation is necessary -- we want to just do the right thing in the default case, right?  It's really only needed if you want to set a short name (in which case you should be able to *just* set a short name), or if you need to override the resource path.

DirectXMan12: add a test case without the resource annotation, just to make sure we do the sane thing if the annotation isn't present.

pkg/general/util.go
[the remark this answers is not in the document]
DirectXMan12: or, move the parts that are necessary for general use.

pkg/internal/codegen/parse/crd.go
DirectXMan12: It assumes *either* you have a resource annotation *or* you have TypeMeta and ObjectMeta.  The comment on the function isn't clear -- it should probably be clarified (you can do so in a separate commit in this PR).

pkg/internal/codegen/parse/crd.go
```diff
@@ -65,14 +65,24 @@ func (b *APIs) parseCRDs() {
...
+						log.Fatalf("failed to parse resource annotations, error: %v", err.Error())
+					}
+					resource.CRD.ObjectMeta.Name = result.Resource + "." + resource.Group + "." + resource.Domain
```
DirectXMan12: it should be possible to *just* specify the shortname here and use the default for the resource name.

pkg/internal/codegen/parse/crd.go
```diff
@@ -120,9 +130,9 @@ func (b *APIs) parseCRDs() {
 						}
 						resource.CRD.Spec.AdditionalPrinterColumns = result
 					}
-					if len(resource.ShortName) > 0 {
-						resource.CRD.Spec.Names.ShortNames = []string{resource.ShortName}
-					}
+					// if len(resource.ShortName) > 0 {
```
DirectXMan12: please don't leave commented code hanging around.

pkg/internal/codegen/parse/index.go
DirectXMan12: that's not "subresourcepath", right?  It should just be "resourcepath", I think, because we're specifying resource aliases, not subresources like `Scale` (unless I'm misreading this).

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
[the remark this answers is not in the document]
droot: @sohankunkerkar No, not this PR.

@DirectXMan12 you are right. Its not necessary. Though it helps with the discoverability for this option.

pkg/general/util.go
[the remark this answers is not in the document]
droot: +1 to @DirectXMan12 's comments.

pkg/internal/codegen/parse/crd.go
droot: I agree, we should support specifying *just* the shortname here because majority of the time users would like do that. So, we should derive the name/plural/shortname only if they were specified in the annotation.

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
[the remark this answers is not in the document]
sohankunkerkar: I've updated the code to handle both the conditions.

sohankunkerkar: The changes included in the code are handling all the conditions.

pkg/internal/codegen/parse/index.go
sohankunkerkar: As per @DirectXMan12 comments, I tried to reuse the code already present in the index.go file(in commented state).I found some issues when I uncomment this (https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/index.go#L76) code where it's using following two functions to extract resource annotation. (`parseResourceTag`(https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/index.go#L187) and `getResourceTag` (https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/index.go#L236)). 

These functions are not handling the base case where the users provide two `resource` annotations. Also, the logic for the key-value pair can be imported from `pkg/internal/general`. Hence, I thought of redesigning this code a little bit (no major changes though). Also, instead of removing those two functions (`parseResourceTag` and `getResourceTag`), I commented out just in case anyone wants to reuse it for some purpose.
Let me know if your feedback

pkg/internal/codegen/parse/index.go
```diff
@@ -74,8 +74,15 @@ func (b *APIs) parseIndex() {
...
-		// r.Resource = rt.Resource
-		// r.ShortName = rt.ShortName
+		rt, err := parseResourceAnnotation(c)
+		if err != nil {
+			log.Fatalf("failed to parse resource annotations, error: %v", err.Error())
+		}
+		if rt.Resource != "" {
+			r.Resource = rt.Resource
+		}
+		r.ShortName = rt.ShortName
```
sohankunkerkar: `If` condition for this code is handled in the crd.go file (https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/crd.go#L123)

sohankunkerkar:
fix https://github.com/kubernetes-sigs/kubebuilder/issues/404

pkg/internal/codegen/parse/index.go
```diff
@@ -74,8 +74,15 @@ func (b *APIs) parseIndex() {
 			r.Resource = strings.ToLower(inflect.Pluralize(r.Kind))
 		}
 		// rt := parseResourceTag(b.getResourceTag(c))
```
DirectXMan12: can you remove this comment here?

pkg/internal/codegen/parse/index.go
```diff
@@ -74,8 +74,15 @@ func (b *APIs) parseIndex() {
...
+		if err != nil {
+			log.Fatalf("failed to parse resource annotations, error: %v", err.Error())
+		}
+		if rt.Resource != "" {
+			r.Resource = rt.Resource
+		}
+		r.ShortName = rt.ShortName
+
 		//r.REST = rt.REST
 		//r.Strategy = rt.Strategy
```
DirectXMan12: same with this commented code (please remove)

pkg/internal/codegen/parse/index.go
```diff
@@ -183,31 +190,80 @@ type resourceTags struct {
 	ShortName string
 }
 
-// ParseResourceTag parses the tags in a "+resource=" comment into a resourceTags struct
-func parseResourceTag(tag string) resourceTags {
-	result := resourceTags{}
+//helperResourceAnnotation is a helper function to extract resource annotation.
+func helperResourceAnnotation(tag string, count int) (resourceTags, int, error) {
```
DirectXMan12: this is not a good function name.  The `helper` doesn't convey anything.  The old name (`parseResourceAnnotation`) was much clearer.  Maybe `parseResourceAnnotationValue`?

pkg/internal/codegen/parse/index.go
DirectXMan12: you shouldn't leave commented code in the file.  The fact that there was commented code here before is wrong.

DirectXMan12: otherwise, the new code looks better.

pkg/internal/codegen/parse/util.go
```diff
@@ -51,7 +51,7 @@ type Options struct {
 	SkipRBACValidation bool
 }
 
-// IsAPIResource returns true if:
+// IsAPIResource returns true if either of the two conditions become true:
```
DirectXMan12: :+1:

pkg/internal/codegen/parse/index.go
sohankunkerkar: Done!

pkg/general/util.go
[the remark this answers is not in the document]
sohankunkerkar: Done!

pkg/internal/codegen/parse/index.go
sohankunkerkar: done!

pkg/internal/codegen/parse/index.go
```diff
@@ -183,61 +115,47 @@ type resourceTags struct {
 	ShortName string
 }
 
-// ParseResourceTag parses the tags in a "+resource=" comment into a resourceTags struct
-func parseResourceTag(tag string) resourceTags {
-	result := resourceTags{}
+//parseResourceAnnotationValue is a helper function to extract resource annotation.
```
droot: nit: space missing between `//` and `parse...`

pkg/internal/codegen/parse/index.go
```diff
@@ -183,61 +115,47 @@ type resourceTags struct {
...
+			res.Resource = value
 		case "shortName":
-			result.ShortName = value
+			res.ShortName = value
+		default:
+			return resourceTags{}, count, fmt.Errorf("The given input %s is invalid", value)
 		}
 	}
-	return result
+	count++
```
droot: in current form, count is just tracking if function was able to parse any resource annotation successfully or not which is also indicated by the returned error. So caller can decide this on the error basis. I don't see the need for count unless I am missing something ?

pkg/internal/codegen/parse/index.go
```diff
@@ -183,61 +115,47 @@ type resourceTags struct {
...
+			continue
+		}
+		result, temp, err := parseResourceAnnotationValue(anno, count)
+		if err != nil {
+			return resourceTags{}, err
+		}
+		res++
+		if res != temp {
+			return resourceTags{}, fmt.Errorf("resource annotation should only exists once per type")
+		}
```
droot: As stated in previous comment,  you just need to track if `resourceAnnotationFound` which can be toggled once `parseResourceAnnotationValue` returns success i.e. no-error. and if we encounter another resource annotation and we were able to parse it suceesfully and `resourceAnnotationFound` was already true, then we can throw an error that multiple resource annotations are not allowed.

pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
```diff
@@ -28,7 +28,9 @@ spec:
   group: fun.myk8s.io
   names:
     kind: Toy
-    plural: toys
+    plural: services
+    shortNames:
+    - ty
```
droot: Do we have test case where we don't have any resource annotations ?

pkg/internal/codegen/parse/index.go
sohankunkerkar: So Instead of using boolean, I was keeping the count of resource annotations and if it exceeds beyond one per API type then I was throwing an error. So logic wise this code is doing the same thing what you proposed. Having said, it would be easier to keep one boolean variable and toggle its value based on certain conditions. I will change that logic right away.

pkg/internal/codegen/parse/index.go
sohankunkerkar: Done!

pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
sohankunkerkar: Yes but I'm not sure how to include another test case(without resource annotation) in the same file.

sohankunkerkar: https://gist.github.com/sohankunkerkar/0e814af32d07da11827d39151a342429 (toy_types.go)
https://gist.github.com/sohankunkerkar/f1007b71d101a571ea7ccb356955c521 (crd)

pkg/internal/codegen/parse/index.go
sohankunkerkar: Follow my previous comments. I've included your feedback in this code.

DirectXMan12:
/lgtm

will let at @droot approve.

PR #108, opened 2018-11-30
Add support for the date type in +printcolumn

nkvoll:
While not part of the OpenAPI spec, "date" is the type used by the
defaulted field "age" to present a relative value (e.g 4h ago).

The full list of allowable types exists at https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L38


nkvoll:
I signed it

sohankunkerkar:
@nkvoll Thanks for bringing this into attention. I missed this part the code.

PR #109, opened 2018-12-02
add support for resource fields to accept name with underscore and hyphen

sohankunkerkar:
Fix for this issue listed in the kubebuilder repo: https://github.com/kubernetes-sigs/kubebuilder/issues/490
I have also added support for `-` in case someone wants to specify it in the resource fields.

### DIFF
--- a/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
+++ b/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
@@ -52,7 +52,7 @@ spec:
               - Wolf
               - Dragon
               type: string
-            bricks:
+            bricks-test:
               format: int32
               type: integer
             claim:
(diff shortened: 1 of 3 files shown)


DirectXMan12:
/hold

neither of those cases are compliant with the kubernetes API guidelines.  We should definitely strive to conform to the Kubernetes API guidelines, and should *probably* enforce it when we generate things.

droot:
> /hold
> 
> neither of those cases are compliant with the kubernetes API guidelines. We should definitely strive to conform to the Kubernetes API guidelines, and should _probably_ enforce it when we generate things.

Thanks for pointing it out @DirectXMan12 
@sohankunkerkar Can you pl. update the bug with the API guideline's link and close this PR. Thanks

sohankunkerkar:
Closing this PR as the Above two cases are not compliant with the kubernetes API guidelines.

PR #110, opened 2018-12-05
:sparkles: remove most k8s deps in Gopkg.toml and update to controller-runtime v0.1.8

mengqiy:
The PR removes most k8s deps from the `Gopkg.toml` file, include:
- k8s.io/api
- k8s.io/apiextensions-apiserver
- k8s.io/apimachinery
- k8s.io/client-go

The reason is that `controller-tools` depends on `controller-runtime` and these dependency can be inferred by `dep` tool through `controller-runtime` repo.

The only k8s dep left in `Gopkg.toml` is `k8s.io/code-generator` which is not a dep on `controller-runtime`.

In the future, we will always release `controller-runtime` and `controller-tools` synchronously.


droot:
We should add description to the PR about why we are removing k8s dependencies ?

mengqiy:
> We should add description to the PR about why we are removing k8s dependencies ?

Done

DirectXMan12:
Gopkg.toml should list whatever dependencies controller-tools actually depends on directly.

DirectXMan12:
AFAICT, it looks like these are all still direct dependencies of controller-tools.  Some of that is coming from the webhook generation code calling into CR, which should probably be migrated here instead of living in CR, but at any rate, we should keep these dependencies listed as long as they're direct dependencies, and we should drop them when they're no longer direct dependencies.

droot:
> we should keep these dependencies listed as long as they're direct dependencies, and we should drop them when they're no longer direct dependencies.

This is what we agreed on this morning. And it is fine as long as we upgrade CT/CR in sync (which we do).

mengqiy:
Closing in favor of #113

PR #111, opened 2018-12-06
Include description in the OpenAPI schema under fields property

### LINKED ISSUE #102: Could CRD generation include field descriptions?
It'd be useful if the comment above each field in the type definition was included as the `description` field for that property in the OpenAPI schema for the generated CRDs.

sohankunkerkar:
Fix #102 
List of assumptions while creating this PR.

1.  Not including `+kubebuilder` and `+optional` comments in the description.
2.  The `description` field will not be created if there are no comments above the field in the type definition.

### DIFF
--- a/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
+++ b/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
@@ -41,8 +41,14 @@ spec:
     openAPIV3Schema:
       properties:
         apiVersion:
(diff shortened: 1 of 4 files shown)


sohankunkerkar:
> The `description` is not getting parsed for those fields whose datatype is not included in this function.
> https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/crd.go#L236
> Working on it to get this resolved.

@droot I handle this condition. Ready for review.

PR #112, opened 2018-12-06
Handle "intstr.IntOrString" use case for the generated openAPI validation

sohankunkerkar:
Fix https://github.com/kubernetes-sigs/kubebuilder/issues/442

### DIFF
--- a/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
+++ b/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
@@ -18,6 +18,7 @@ package v1alpha1
 import (
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -50,7 +51,7 @@ type ToySpec struct {
 	Template v1.PodTemplateSpec       `json:"template"`
 	Claim    v1.PersistentVolumeClaim `json:"claim,omitempty"`
 
-	Replicas *int32 `json:"replicas"`
+	Replicas *intstr.IntOrString `json:"replicas"`
```
droot: The previous field-type was for testing *int32, so can we retain the previous type and introduce new field for testing `IntOrString` ?

sohankunkerkar: Done!

PR #118, opened 2018-12-15
:bug: Fix description generation to work for all types of fields

sohankunkerkar:
-  What does this do, and why do we need it?
This PR fix the description field generation for all types of fields under `Openschema API`.
For more details, please go through this issue #116

### DIFF
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -336,14 +336,16 @@ func parseByteValue(b []byte) string {
 
 // parseDescription parse comments above each field in the type definition.
 func parseDescription(res []string) string {
-	var temp, data string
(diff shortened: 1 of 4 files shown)


pkg/internal/codegen/parse/util.go
```diff
@@ -336,11 +336,13 @@ func parseByteValue(b []byte) string {
 
 // parseDescription parse comments above each field in the type definition.
 func parseDescription(res []string) string {
-	var temp, data string
+	var temp strings.Builder
+	var data string
```
droot: Consider renaming `data` to `desc` to reflect the usage ?

sohankunkerkar: Done!

PR #119, opened 2018-12-19
:sparkles: Object scaffolding logic from KubeBuilder

DirectXMan12:
This extracts KubeBuilder's object scaffolding logic into a separate
command.  Creating the basic structure for Kubernetes objects is a
non-opinionated task, so it can be reused for other purposes
(incrementally adding types to files, non-KubeBuilder scaffolding
projects, etc).

### DIFF
--- a/pkg/typescaffold/resource.go
+++ b/pkg/typescaffold/resource.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package typescaffold
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/markbates/inflect"
+)
+
+// Resource contains the information required to scaffold files for a resource.
+type Resource struct {
+	// Namespaced is true if the resource is namespaced
+	Namespaced bool
+
+	// Kind is the API Kind.
+	Kind string
+
+	// Resource is the API Resource.
+	Resource string
+}
+
+// Validate checks the Resource values to make sure they are valid.
+func (r *Resource) Validate() error {
+	if len(r.Kind) == 0 {
+		return fmt.Errorf("kind cannot be empty")
+	}
+
+	rs := inflect.NewDefaultRuleset()
+	if len(r.Resource) == 0 {
+		r.Resource = rs.Pluralize(strings.ToLower(r.Kind))
+	}
+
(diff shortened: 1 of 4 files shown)


cmd/object-scaffold/main.go
```diff
@@ -0,0 +1,72 @@
...
+func main() {
+	opts := &object.ScaffoldOptions{
+		Resource: object.Resource{
+			Namespaced: true,
+		},
+	}
+	scaffoldCmd := &cobra.Command{
+		Use:   "object-scaffold",
+		Short: "Quickly scaffold out basic bits of a Kubernetes object.",
+		Long: `Quickly scaffold out the structure of a Kubernetes root type and associated types.
```
droot: May be take out `root` out ? When I read for the first time, I didn't get it. Then I realized, we are referring to the object type. 

Thinking more... an alternative good name for the command could be `type-scaffold` ?

pkg/object/scaffold.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Scaffold prints the Kubernetes object scaffolding to the given output.
+func (o *ScaffoldOptions) Scaffold(out io.Writer) error {
+	return typesTemplate.Execute(out, o)
+}
```
droot: Initially we had scaffolding tools under controller-tools and fixing bugs required multiple steps across two repos which is:
 
 - fixing the bug in controller-tools
 - tagging the controller-tools.
 - then, updating the deps in kubebuilder

Which felt a bit slow. Since we didn't have any users for scaffold tools in controller-tools, we moved the scaffold tools to kubebuilder repo. It simplified the dev/test/release workflow. But then we were iterating faster because KB was in early days, we may not feel the same pain now because our requirements are matured now.

DirectXMan12:
Added tests.  I'm not testing the direct textual output, since that seems super-fragile.  A better test seems like it would be ensuring that the necessary types and comments were there, and that the code was parsable, but that seems rather involved.

pkg/typescaffold/scaffold_test.go
```diff
@@ -0,0 +1,104 @@
...
+			err := test.opts.Validate()
+			if err != nil {
+				t.Fatalf("unable to validate scaffold opts: %v", err)
+			}
+			var out bytes.Buffer
+			if err := test.opts.Scaffold(&out); err != nil {
+				t.Fatalf("unable to scaffold types: %v", err)
+			}
+
+			// TODO(directxman12): testing the direct output seems fragile
```
droot: We used golden files for the same case in Kubebuilder. That isn't perfect because generating golden files uses the same binary (under test) but sort-of-does-the-job.

droot:
travis isn't happy because one of the files needs `goimports` run.

DirectXMan12:
goimports run.  looks good.  @mengqiy or @droot, can you re-lgtm?

pkg/typescaffold/resource.go
```diff
@@ -0,0 +1,54 @@
...
+limitations under the License.
+*/
+
+package typescaffold
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/markbates/inflect"
```
mengqiy: It seems this is deprecated. We should use `github.com/gobuffalo/flect`.

PR #125, opened 2019-01-15
✨add shortnames annotation

magicsong:
fix https://github.com/kubernetes-sigs/kubebuilder/issues/404

current implement  only supports single shortname, however, in CRD, `shortName` is an array.  this pr create a new annotation to help generating shortnames. 

Signed-off-by: magicsong <magicsong@yunify.com>

### DIFF
--- a/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
+++ b/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
@@ -30,6 +30,7 @@ spec:
     kind: Toy
     plural: services
     shortNames:
+    - to
     - ty
   scope: Namespaced
   subresources:
--- a/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
+++ b/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
@@ -93,7 +93,7 @@ type ToyStatus struct {
 // +kubebuilder:printcolumn:name="toy",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="descr1",format="date",priority=3
 // +kubebuilder:printcolumn:name="abc",type="integer",JSONPath="status",description="descr2",format="int32",priority=1
 // +kubebuilder:printcolumn:name="service",type="string",JSONPath=".status.conditions.ready",description="descr3",format="byte",priority=2
-// +kubebuilder:resource:path=services,shortName=ty
+// +kubebuilder:resource:path=services,shortName=to;ty
 type Toy struct {
 	metav1.TypeMeta   `json:",inline"`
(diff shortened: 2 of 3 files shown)


pkg/internal/codegen/parse/util.go
```diff
@@ -40,6 +40,8 @@ const (
 	printColumnFormat  = "format"
 	printColumnPri     = "priority"
 	printColumnError   = "invalid printcolumn path. name,type, and JSONPath are required kye-value pairs and rest of the fields are optinal. For example: // +kubebuilder:printcolumn:name=abc,type=string,JSONPath=status"
+	shortnamePrefix    = "+kubebuilder:shortname:"
+	shortNameError     = "invalid shortname params. Shortname should be a no-empty string and is seperated by comma. For example: // " + shortnamePrefix + "deploy,dpl"
```
DirectXMan12: nit: `non-empty string`

DirectXMan12: also could just write `non-empty comma-separated list of names`

magicsong: @DirectXMan12 thanks for your time to review this pr!  the sentence was modified as your request

pkg/internal/codegen/parse/util.go
```diff
@@ -40,6 +40,8 @@ const (
 	printColumnFormat  = "format"
 	printColumnPri     = "priority"
 	printColumnError   = "invalid printcolumn path. name,type, and JSONPath are required kye-value pairs and rest of the fields are optinal. For example: // +kubebuilder:printcolumn:name=abc,type=string,JSONPath=status"
+	shortnamePrefix    = "+kubebuilder:shortname:"
```
droot: Instead of introducing new annotation, can we extend the existing `kubebuilder:resource` annotation to support an array ? we can take `;` delimited list. An example:

```Go
// kubebuilder:resource:path=services,shortName=ty;to;

```

sohankunkerkar: @droot yes, we can certainly extend that annotation instead of creating a new one. Let me know if you want me to file a PR for this?

sohankunkerkar: @magicsong you can probably change the code here
https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/index.go#L115
https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/index.go#L119
 to accommodate your logic.

magicsong: @sohankunkerkar i have changed the code, wdyt?

sohankunkerkar: @magicsong yeah, the changes look good to me.

DirectXMan12:
/lgtm
/approve

we need to make sure to document this somewhere.

PR #126, opened 2019-01-15
✨ Add support for the interface member

sohankunkerkar:
Fix for https://github.com/kubernetes-sigs/kubebuilder/issues/528
**Why do we need this change?**
The current code is not handling a condition for the `interface{}` member while generating a CRD for the given type.

### DIFF
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -220,6 +220,8 @@ func (b *APIs) typeToJSONSchemaProps(t *types.Type, found sets.String, comments 
 		v, s = b.typeToJSONSchemaProps(t.Elem, found, comments, false)
 	case types.Alias:
 		v, s = b.typeToJSONSchemaProps(t.Underlying, found, comments, false)
+	case types.Interface:
+		v, s = b.parseInterfaceValidation(t, found, comments)
 	default:
 		log.Fatalf("Unknown supported Kind %v\n", t.Kind)
 	}
@@ -417,6 +419,38 @@ func (b *APIs) parseArrayValidation(t *types.Type, found sets.String, comments [
 	return props, buff.String()
 }
 
+// parseInterfaceValidation returns a JSONSchemaProps object and its serialization in
+// Go that describe the validations for the given Interface.
+func (b *APIs) parseInterfaceValidation(t *types.Type, found sets.String, comments []string) (v1beta1.JSONSchemaProps, string) {
+	props := v1beta1.JSONSchemaProps{
+		AnyOf: []v1beta1.JSONSchemaProps{
+			{
+				Type: "string",
+			},
+			{
+				Type: "number",
+			},
+			{
+				Type: "object",
+			},
+			{
+				Type: "null",
+			},
+			{
+				Type: "array",
+			},
+			{
+				Type: "boolean",
+			},
+			{
+				Type: "integer",
+			},
+		},
+		Description: parseDescription(comments),
+	}
+	return props, b.objSchema()
+}
+
 type objectTemplateArgs struct {
 	v1beta1.JSONSchemaProps
 	Fields   map[string]string
@@ -577,6 +611,19 @@ func getValidation(comment string, props *v1beta1.JSONSchemaProps) {
 		}
 	case "Format":
 		props.Format = parts[1]
+	case "AnyOf[]":
+		result := []v1beta1.JSONSchemaProps{}
+		intialValues := strings.Split(parts[1], ";")
+		for _, s := range intialValues {
+			value := strings.Split(s, ",")
+			res, err := extractAnyOfParameters(value)
+			if err != nil {
+				log.Fatalf("Use proper kubebuilder AnyOf annotation %s: %v", comment, err)
+				return
+			}
+			result = append(result, res...)
+		}
+		props.AnyOf = result
 	default:
 		log.Fatalf("Unsupport validation: %s", comment)
 	}
--- a/testData/config/crds/fun_v1alpha1_toy.yaml
+++ b/testData/config/crds/fun_v1alpha1_toy.yaml
@@ -129,6 +129,11 @@ spec:
                   minLength: 1
                   type: string
               type: object
+            server:
+              anyOf:
+              - type: boolean
+              - type: integer
+              type: object
             template:
               description: This is a comment on an object field.
               type: object
@@ -148,11 +153,22 @@ spec:
           type: object
         status:
           properties:
+            inter:
+              anyOf:
+              - type: string
+              - type: number
+              - type: object
+              - type: "null"
+              - type: array
+              - type: boolean
+              - type: integer
+              description: Adding this field for testing interface{} validation
             replicas:
               description: It tracks the number of replicas.
               format: int32
               type: integer
           required:
+          - inter
           - replicas
           type: object
   version: v1alpha1
--- a/testData/pkg/apis/fun/v1alpha1/toy_types.go
+++ b/testData/pkg/apis/fun/v1alpha1/toy_types.go
@@ -16,7 +16,10 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"k8s.io/api/core/v1"
+	"encoding/json"
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
@@ -79,6 +82,9 @@ type ToySpec struct {
 	// Using this for testing purpose.
 	Rook *intstr.IntOrString `json:"rook"`
 
+	// +kubebuilder:validation:AnyOf[]=type:boolean;type:integer
+	Server *BoolOrInt `json:"server,omitempty"`
+
 	// Used this for testing fieldNames with number.
 	S3Log ToyS3LogConfig `json:"s3Log"`
 
@@ -98,6 +104,9 @@ type ToySpec struct {
 type ToyStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
+
+	//Adding this field for testing interface{} validation
+	Inter interface{} `json:"inter"`
 
 	// It tracks the number of replicas.
 	Replicas int32 `json:"replicas"`
@@ -134,4 +143,33 @@ type ToyList struct {
 
 func init() {
 	SchemeBuilder.Register(&Toy{}, &ToyList{})
+}
+
+// BoolOrInt is defined for testing kubebuilder anyof annotation
+type BoolOrInt struct {
+	Type    string
+	BoolVal bool
+	IntVal  string
+}
+
+// UnmarshalJSON implements the json.Unmarshaller interface.
+func (b *BoolOrInt) UnmarshalJSON(value []byte) error {
+	if value[0] == 't' || value[0] == 'f' {
+		b.Type = "bool"
+		return json.Unmarshal(value, &b.BoolVal)
+	}
+	b.Type = "int"
+	return json.Unmarshal(value, &b.IntVal)
+}
+
+// MarshalJSON implements the json.Marshaller interface.
+func (b *BoolOrInt) MarshalJSON() ([]byte, error) {
+	switch b.Type {
+	case "int":
+		return json.Marshal(b.IntVal)
+	case "bool":
+		return json.Marshal(b.BoolVal)
+	default:
+		return []byte{}, fmt.Errorf("impossible BoolOrInt.Type")
+	}
 }
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -29,17 +29,22 @@ import (
 )
 
 const (
-	specReplicasPath   = "specpath"
-	statusReplicasPath = "statuspath"
-	labelSelectorPath  = "selectorpath"
-	jsonPathError      = "invalid scale path. specpath, statuspath key-value pairs are required, only selectorpath key-value is optinal. For example: // +kubebuilder:subresource:scale:specpath=.spec.replica,statuspath=.status.replica,selectorpath=.spec.Label"
-	printColumnName    = "name"
-	printColumnType    = "type"
-	printColumnDescr   = "description"
-	printColumnPath    = "JSONPath"
-	printColumnFormat  = "format"
-	printColumnPri     = "priority"
-	printColumnError   = "invalid printcolumn path. name,type, and JSONPath are required kye-value pairs and rest of the fields are optinal. For example: // +kubebuilder:printcolumn:name=abc,type=string,JSONPath=status"
+	specReplicasPath            = "specpath"
+	statusReplicasPath          = "statuspath"
+	labelSelectorPath           = "selectorpath"
+	jsonPathError               = "invalid scale path. specpath, statuspath key-value pairs are required, only selectorpath key-value is optinal. For example: // +kubebuilder:subresource:scale:specpath=.spec.replica,statuspath=.status.replica,selectorpath=.spec.Label"
+	printColumnName             = "name"
+	printColumnType             = "type"
+	printColumnDescr            = "description"
+	printColumnPath             = "JSONPath"
+	printColumnFormat           = "format"
+	printColumnPri              = "priority"
+	printColumnError            = "invalid printcolumn path. name,type, and JSONPath are required kye-value pairs and rest of the fields are optinal. For example: // +kubebuilder:printcolumn:name=abc,type=string,JSONPath=status"
+	printAnyOfOtherError        = "AnyOf annotation for type other than string should be enter in the following format--> type:<value>"
+	printAnyOfStringError       = "AnyOf annotation for type=string should be enter in the following format--> type:<value>,format:<value>"
+	printPrimitiveError         = "The primitive type or format value is not appropriate in this case"
+	printPrimitiveTypeOnlyError = "The primitive type value is not appropriate in this case"
+	printGeneralAnyOfError      = "Ensure you entered the right kubebuilder AnyOf annotation"
 )
 
 // Options contains the parser options
@@ -223,8 +228,8 @@ func hasSingular(t *types.Type) bool {
 	if !IsAPIResource(t) {
 		return false
 	}
-	for _, c := range t.CommentLines{
-		if strings.Contains(c, "+kubebuilder:singular"){
+	for _, c := range t.CommentLines {
+		if strings.Contains(c, "+kubebuilder:singular") {
 			return true
 		}
 	}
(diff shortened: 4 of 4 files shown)


pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -74,11 +74,20 @@ type ToySpec struct {
 	Location map[string]string `json:"location"`
 }
 
+// Inter interface.
+type Inter interface{}
```
vincepri: Why not `Interface` as name?

DirectXMan12:
I think instead of saying "all interfaces get zero validation", we come up with a way to annotate the basic validation properties of things like IntOrString.  That way, you could choose to annotate your interface with `// +kubebuilder:validation:any` if you wanted to, or you could reproduce something like `IntOrString` via `// +kubebuilder:validation:one-of="integer,string"` (that syntax is just a strawman)

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
sohankunkerkar: Yeah, even if I change the name to `Interface`, the o/p remains same.

sohankunkerkar:
@DirectXMan12 do you have any suggestions for the problem you listed above?

DirectXMan12:
I think the annotation forms I suggested above are decent, unless anyone has any objection we could move forward with those.

sohankunkerkar:
> I think the annotation forms I suggested above are decent, unless anyone has any objection we could move forward with those.
@DirectXMan12  I think I haven't seen anyone putting any objections for current suggestions so I will go ahead with your approach and make changes in the exiting PR itself.

alegacy:
>.... `// +kubebuilder:validation:any` if you wanted to, or you could reproduce something like `IntOrString` via `// +kubebuilder:validation:one-of="integer,string"` (that syntax is just a strawman).
> 
> 
Can you provide a complete example of how this would work (both Go types and CRD YAML)?  If the JSON is rendered as an object how is it going to be able to validate as an int or string?    I am following this thread with interest related to my google group posting (https://groups.google.com/forum/#!topic/kubebuilder/oE8n3hMRlEg) and trying to figure out how that would apply to those 2 scenarios cleanly and with validation.

DirectXMan12:
let's take the case of the `IntOrString` type from [k8s.io/apimachinery/pkg/util/intstr](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/intstr/intstr.go).

Code-wise, it's used like this:

```go
type ServicePort struct {
    // can be a numeric port or a named port reference from a pod
    ContainerPort intstr.IntOrString
    Port int
}
```

Which looks like

```json
{"containerPort": 8443}
```

or

```json
{"containerPort": "https"}
```

It's implemented roughly like (taking some liberties for brevity):

```go
type IntOrString struct {
    Type string
    IntVal int
    StringVal string
}

func (is *IntOrString) UnmarshalJSON(value []byte) error {
    if value[0] == '"' {
        is.Type = "string"
        return json.Unmarshal(value, &is.StringVal)
    }
    is.Type = "int"
    return json.Unmarshal(value, &is.IntVal)
}
```

Validation-wise, we currently special-case it in our generation code.  It ends up looking like:

```yaml
containerPort:
  oneOf: [{type: "string"}, {type: "int"}]
```

Under the desired system, you'd instead write:

```go
type ServicePort struct {
    // kubebuilder:validation:one-of=int;string
    // can be a numeric port or a named port reference from a pod
    ContainerPort intstr.IntOrString
    Port int
}
```

which would yield the same thing.  Does that make sense?

alegacy:
@DirectXMan12   Thanks that clarifies things for me.    Your current approach allows for varying the actual "type" for an attribute.   What I would like to see added is the ability to also vary the "format".  My immediate need is for all format variations to share the same type (string) but I can see based on your example how it may be necessary to support more permutations - based on type alone, format alone, or combinations of type and format.     

My immediate need, at least for the first use case in my mailing list post, is to vary the format for a string so that I can take advantage of the syntax validation at the API server.   So, an example would be:

```go
type MyStructSpec {
  ...
  // This is a list of server IPv4 or IPv6 addresses, or FQDN hostnames
  // +kubebuilder:validation:one-of=string/ipv4;string/ipv6;string/hostname
  Servers []string  `json:"servers"`
  ...
}
```

would generate:

```yaml
servers:
  type: array
  items:
    oneOf:
      - type: string
         format: ipv4
      - type: string
         format: ipv6
      - type: string
         format: hostname
```

As for the second use case in my mailing list post I don't see a good way to accomplish that without a lot of complexity.  I think it may be better to do that type of multiple field validation as a webhook or some other way. 

I apologize if I have taken this farther away from your initial Interface use case than you intended to go.  I can open a separate issue if you prefer to separate these use cases.

pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
```diff
@@ -133,11 +152,18 @@ spec:
           type: object
         status:
           properties:
+            inter:
+              description: Adding this field for testing interface{} validation
+              oneOf:
+              - type: string
+              - type: integer
+              - type: object
```
sohankunkerkar: @DirectXMan12  Do you think `boolean` and `array` type would be valid in this case?

DirectXMan12:
> I think it may be better to do that type of multiple field validation as a webhook or some other way.

Nah, if there's support in the underlying CRD object, we should have a way to accomplish it.  We just need to fiddle with the syntax a bit.

Initial options off the top of my head (I'd lean towards the middle one, but WDYT):

#### Totally not just compact YAML

 `// kubebuilder:validation:one-of={type: string, format: ipv6}, {type: string, format: ipv4}`

Cons:
- not particularly pretty
- requires YAML parsing
- restricted to a single line, so it could get *really* long

#### Single line, URL query string style

```
// kubebuilder:validation:one-of[]=Type=string;Format:ipv6
// kubebuilder:validation:one-of[]=Type=string;Format:ipv4
```

Pros:
- each segment looks like what we have now
- vaguely similar to other kubebuilder multi-value annotations
- easy to split across multiple lines
Cons:
- not particular pretty

#### Lots of comment lines

```
// kubebuilder:validation:one-of[0]:Type=string
// kubebuilder:validation:one-of[0]:Format=ipv4
// kubebuilder:validation:one-of[1]:Type=string
// kubebuilder:validation:one-of[1]:Format=ipv6
```

Pros:
- looks closest to what we do now
- Easy to split across multiple lines
Cons:
- requires lots of lines
- not immediately visually obvious what's associated with what

alegacy:
@DirectXMan12 I agree, the second approach seems like the best option.   This would allow us to specify other validation properties on each variation (i.e., Pattern, MaxLength, Minimum, Maximum, etc...).  I would only swap out the "one-of" for "OneOf" to match the capitalization pattern of other ones like "MaxLength".

pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
j-vizcaino: According to http://json-schema.org/latest/json-schema-core.html#rfc.section.4.2.1 and if we want this to be exhaustive and tolerated any JSON type, this should include `boolean`, `array`, `null` and `number` as well DYT?

j-vizcaino:
Late to the party but I'm also super interested in seeing this handled.

My use-case is a bit simpler than what's proposed I guess, since my object has a `map[string]interface{}` that is basically free-form (ie. no validation on the interface values)

As posted [here](https://github.com/kubernetes-sigs/kubebuilder/issues/528#issuecomment-466449483), it seems there are two issues related to supporting interfaces:

* declaring a `map[string]interface{}` as a struct member
  * `deepcopy-gen` is having trouble creating the functions
  * CRD generation identifies the type has being a `Map` but cannot handle the value type.
* declaring a `type Foo map[string]interface{}` and creating the DeepCopy manually
  * solves the `deepcopy-gen` issue
  * CRD generation still fails, to what I suspect is the inability to identify `Foo` as being a map alias? (looks like the code has support for this but the error is still `Unknown supported Kind Interface`)

A very simple solution for my use-case would be to have the ability, using comments, to tell the generator that the member should be treated as an `object`.

DirectXMan12:
> I would only swap out the "one-of" for "OneOf" to match the capitalization pattern of other ones like "MaxLength"

:+1:

> `deepcopy-gen` is having trouble creating the functions

can you file another bug about that?

j-vizcaino:
@DirectXMan12 The `deepcopy-gen` issue has already been reported in the past: https://github.com/kubernetes/gengo/issues/128
While the guidance provided here helps, it still requires the user to figure out what's wrong. It would be nice to either have deepcopy-gen error out immediately (that is not generate code that will not compile)
Do you still want me to create an issue for this?

DirectXMan12:
Yeah, deepcopy-gen should probably error out.  That should probably be a separate issue, but mention the previous issue.

sttts:
> declaring a `type Foo map[string]interface{}` and creating the DeepCopy functions manually
> * solves the `deepcopy-gen` issue

This is the right approach. If you have polymorphic interfaces, i.e. many different interface implementation, take a look at `// +k8s:deepcopy-gen:interfaces=....`.

DirectXMan12:
@sttts helpfully pointed out that we probably want `anyOf` support here instead of `oneOf`(elsewhere in CT), so we should probably update this PR with that (although I still think oneOf can be valuable too, but that's another matter).

It also needs to be updated with the newer syntax.

pkg/internal/codegen/parse/crd.go
```diff
@@ -402,6 +404,26 @@ func (b *APIs) parseArrayValidation(t *types.Type, found sets.String, comments [
...
+	props := v1beta1.JSONSchemaProps{
+		OneOf: []v1beta1.JSONSchemaProps{
+			{
+				Type: "string",
+			},
+			{
+				Type: "integer",
+			},
+			{
+				Type: "object",
```
sttts: why should an interface be no slice? no number? Feels wrong. We shouldn't invent semantics which is not generic.

sohankunkerkar: @sttts yeah, my bad. I will rectify this error.

testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -28,6 +28,10 @@ import (
 type ToyS3LogConfig struct {
 	// +kubebuilder:validation:MinLength=1
 	Bucket string `json:"bucket,omitempty"`
+	// Must be an IPv4 or IPv6 address
+	Address string `json:"address"`
+	// Must be a value in range that agrees with type of "Address" (i.e., 1-32 for IPv4, 1-128 for IPv6).
+	Prefix int `json:"prefix"`
```
alegacy: I don't quite understand why ```address``` and ```prefix``` are part of this struct.   Line 89 requires that entries in this array be of string or integer.  (e.g., ```kubebuilder:validation:AnyOf=string,interface```)

testData/config/crds/fun_v1alpha1_toy.yaml
```diff
@@ -125,13 +125,50 @@ spec:
...
+                    type: string
+                  prefix:
+                    description: Must be a value in range that agrees with type of
+                      "Address" (i.e., 1-32 for IPv4, 1-128 for IPv6).
+                    format: int64
+                    type: integer
+                required:
+                - address
+                - prefix
+                type: object
```
alegacy: The ```inter``` field is defined as an Interface in the types.go file.  Why is this object included here with ```address```, ```prefix```, and ```bucket``` fields?

testData/pkg/apis/fun/v1alpha1/toy_types.go
DirectXMan12: not quite following this either -- I'd imagine `string,integer` would be for something like port.  I could see `AnyOf` for types that have custom deserialization and look like this (e.g. `AnyOf[]=type:string;format=ipv4` and `AnyOf[]=type:string;format:ipv6`), but this example isn't clear.

sohankunkerkar: @alegacy I was having some different interpretation about this annotation, however, @DirectXMan12 cleared my confusion. I will update the code shortly.

testData/config/crds/fun_v1alpha1_toy.yaml
sohankunkerkar: @alegacy The `type: object` is getting included here because of this code https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/crd.go#L436

alegacy: Yes, I see where```type:object``` is coming from, but I was questioning where the ```properties``` subsection is coming from.

alegacy: Ok, ignore this whole comment... for some reason I thought that ```S3Log``` and ```TestConfig" where of different types.

sohankunkerkar:
PTAL @droot @DirectXMan12

pkg/internal/codegen/parse/crd.go
```diff
@@ -577,6 +608,31 @@ func getValidation(comment string, props *v1beta1.JSONSchemaProps) {
...
+	case "AnyOf[]":
+		anyof := []v1beta1.JSONSchemaProps{}
+		intialValues := strings.Split(parts[1], ";")
+		for _, s := range intialValues {
+			value := strings.Split(s, ",")
+
+			if len(value) < 2 {
+				log.Fatalf("please enter the annotation in type:<value>,format:<value> %s:", comment)
+			}
+			typeValue := value[0]
```
DirectXMan12: shouldn't we actually be parsing the sub-value out of here instead of assuming that type is first, etc?

testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -79,6 +82,9 @@ type ToySpec struct {
 	// Using this for testing purpose.
 	Rook *intstr.IntOrString `json:"rook"`
 
+	// +kubebuilder:validation:AnyOf[]=type:boolean,format:ipv6;type:number,format:ipv4
+	Server *BoolOrInt `json:"server,omitempty"`
```
DirectXMan12: not sure this format makes sense here -- the format directive only applies for strings.

pkg/internal/codegen/parse/crd.go
sohankunkerkar: Ack. I've handled this logic in the code

testData/pkg/apis/fun/v1alpha1/toy_types.go
sohankunkerkar: My bad. I changed the code to make the format value more relevant.

testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -79,6 +82,9 @@ type ToySpec struct {
 	// Using this for testing purpose.
 	Rook *intstr.IntOrString `json:"rook"`
 
+	// +kubebuilder:validation:AnyOf[]=type:boolean,format:bool;type:number,format:int32
```
DirectXMan12: no, I mean having format with anything other that `type:string` is invalid openapi/jsonschema, IIRC.  we need to support it being there, but also not being there.

sohankunkerkar: @DirectXMan12  Ack. I've included your feedback in this PR. Kindly review it.

alegacy:
@sohankunkerkar has this change been abandoned?

sohankunkerkar:
> @sohankunkerkar has this change been abandoned?

@alegacy It appears to me that this change has been included in the latest release.
https://github.com/kubernetes-sigs/kubebuilder/issues/528#issuecomment-497159362

szediktam:
It is not fixed in master. Error code:
```go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd paths=./apis/... output:crd:dir=/tmp/crds
XXXXXXXX/types.go:55:30: unsupported AST kind *ast.InterfaceType
run `controller-gen crd paths=./pkg/apis/aiops/... output:crd:dir=/tmp/crds -w` to see all available markers, or `controller-gen crd paths=./pkg/apis/aiops/... output:crd:dir=/tmp/crds -h` for usage
exit status 1```

**Comment by DirectXMan12** [comment]
We've currently explicitly shied away from supporting `interface{}` in favor of `json.RawMessage`, `runtime.RawExtension`, etc.  That being said, I'm willing to reconsider if there are a number of users requesting the feature with usecases not supported by the above.  Please discuss on #294

**Comment by NicolasT** [comment]
While https://github.com/kubernetes-sigs/controller-tools/issues/294 is about `object` (the JSON type) values embedded in a CRD, I have a (related but similar) need for (I guess) `interface{}` embedding in a CRD struct.

End goal: allow the user to provide *jsonpatch* patches, which would be applied against a JSON config structure generated by the controller, based on info from the CR before the final config is presented to the managed application.

Every patch itself is a JSON object, so I could make `configPatches` a `[]json.RawMessage` or `[]unstructured.Unstructured`. However, for starters, using `json.RawMessage` is not really working because this renders the OpenAPI spec as `type: string` and `format: byte`, so input JSON object would
1. need to be serialized into a string, then embedded in the CR YAML :confused: 
2. need to be base64-encoded in the CR YAML :confused: :confused:

Using `[]unstructure.Unstructured` would work, but at the same time I'd loose *all* type information at the OpenAPI (doc) and validation level: any generic JSON `object` would be allowed, whilst a *jsonpatch* is not just any type of object.

What I'd like to achieve instead is to define a `type ... struct` which represents a mostly-valid *jsonpatch*, i.e., include `Op string`, `Path string` and `From string` fields, and a `Value` field. However, here I'm stuck. The *value* of the `value` field in a *jsonpatch* `object` can be *any valid JSON value*, not just `object`. So, using `unstructured.Unstructured` here would not work, since it requires an `object` value. `json.RawMessage` could, but then I'm back to the `string` serialization, base64 encoding,...

Basically, I'd need a type for the `Value` member of my `JSONPatch` struct that results in a Go value of any (valid JSON) type, so I guess either `interface{}` or a discriminated union of all JSON types, and an OpenAPI field declaration without type definition (such that it can be anything), or (again) list all possible JSON types (and use `oneOf`).

Looking at the 'well known types' in https://godoc.org/github.com/kubernetes-sigs/controller-tools/pkg/crd#pkg-variables didn't show anything useful for my scenario...

**Comment by NicolasT** [comment]
Ugh, now to see that `k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.JSON` may suit my needs...

**Comment by NicolasT** [comment]
For future reference, a code snippet displaying how to include "any JSON value" in a CRD definition (using `kubebuilder`) can be found at https://kubernetes.slack.com/archives/CAR30FCJZ/p1602515995453500?thread_ts=1602513628.452900&cid=CAR30FCJZ. Case at hand is defining a valid JSONPatch structure.

**Comment by chenmeng1996** [comment]
> For future reference, a code snippet displaying how to include "any JSON value" in a CRD definition (using `kubebuilder`) can be found at https://kubernetes.slack.com/archives/CAR30FCJZ/p1602515995453500?thread_ts=1602513628.452900&cid=CAR30FCJZ. Case at hand is defining a valid JSONPatch structure.

I don't have access to Slack, can you show it here please? Thank you.

**Comment by blinkeye** [comment]
> > For future reference, a code snippet displaying how to include "any JSON value" in a CRD definition (using `kubebuilder`) can be found at https://kubernetes.slack.com/archives/CAR30FCJZ/p1602515995453500?thread_ts=1602513628.452900&cid=CAR30FCJZ. Case at hand is defining a valid JSONPatch structure.
> 
> I don't have access to Slack, can you show it here please? Thank you.

The workaround is to use `apiextensionsv1.JSON` instead of `interface{}`, example:
```go
import (
	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
)

type GranularitySpec struct {
	QueryGranularity *apiextensionsv1.JSON `json:"queryGranularity,omitempty"`
}
```
I additionally needed to:
```bash
go get k8s.io/apiextensions-apiserver@v0.23.0
```

PR #128, opened 2019-01-24
:running: upgrade k8s deps to 1.13

droot:
:running: upgrade k8s deps to 1.13


mengqiy:
The code LGTM.
It need a rebase when the v0.1.10 controller-runtime release comes out.

PR #130, opened 2019-01-30
✨ Add support for `nested` output flag and functionality

ichekrygin:
Add functionality to support nested output format for generated CRD files.

`go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all`
result:
```
config/crds/
└── ships_v1beta1_sloop.yaml
```

go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all --nested`
result:
```
config/crds/
└── ships
    └── v1beta1
        └── sloop.yaml
```

Tracking: #94

### DIFF
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -107,6 +107,7 @@ Usage:
 	f := cmd.Flags()
 	f.StringVar(&g.RootPath, "root-path", "", "working dir, must have PROJECT file under the path or parent path if domain not set")
 	f.StringVar(&g.OutputDir, "output-dir", "", "output directory, default to 'config/crds' under root path")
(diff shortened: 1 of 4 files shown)


DirectXMan12:
code seems fine to me, but I otherwise don't really have opinions on the concept -- I'm not entirely certain as to why it's necessary, but I also can't convince myself to dis-approve of it.  What do other people think?

ichekrygin:
@DirectXMan12 , thank you for the review, and, yes, I agree, for the simple project the benefits of the `nested` output is not apparent, and perhaps even not needed. On the flip side, for more complicated [projects](https://github.com/crossplaneio/crossplane/tree/master/cluster/charts/crossplane/crds), with multiple types, and multiple group nesting (related to #81, #131) the nested output improves overall organization.

DirectXMan12:
this would need to be updated to the 0.2.0 logic if you're still interested.

PR #131, opened 2019-01-30
✨ Add support for nested groups.

ichekrygin:
Add support for project with nested groups:
```
pkg/apis
├── addtoscheme_foo_v1alpha1.go
├── apis.go
└── foo
    ├── buz
    │   ├── group.go
    │   └── v1alpha1
    │       ├── bar_types.go
    │       ├── bar_types_test.go
    │       ├── doc.go
    │       ├── register.go
    │       ├── v1alpha1_suite_test.go
    │       └── zz_generated.deepcopy.go
    └── group.go
```

Tracking: #81

### DIFF
--- a/pkg/crd/generator/generator.go
+++ b/pkg/crd/generator/generator.go
@@ -144,8 +144,20 @@ func (c *Generator) writeCRDs(crds map[string][]byte) error {
 	return nil
 }
 
+func getGroupElements(group string) []string {
+	ss := strings.Split(group, ".")
+	if len(ss) > 1 {
+		// reverse slice
+		for i := len(ss)/2 - 1; i >= 0; i-- {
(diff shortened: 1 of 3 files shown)


droot:
@ichekrygin is nesting groups a common pattern ?  

/cc @DirectXMan12

DirectXMan12:
I think we might generally want to revisit how we do discovery of locations in controller-tools -- we're starting to see more patterns.  I think we might just want to switch over to something like `go:generate`-style annotations on the files instead, and just discovering them whereever they are.

ichekrygin:
@droot I am not sure if netsting group is a pattern, but I can imagine for the projects that contain multiple resources it could be useful to break them into "more" granular groups. For example: https://github.com/crossplaneio/crossplane/tree/master/pkg/apis

PR #136, opened 2019-02-01
:sparkles: Move webhook generator

mengqiy:
This PR
- add webhook subcommand as part of `controller-gen all`
- move the generator code from controller-runtime to controller-tools and make some necessary changes.
- move some file to make CT not depend on CR from `dep` perspective.

Updating the annotation will be done in a separate PR, since it requires more than just updating the annotation. 
Annotation format is at https://github.com/jetstack/cert-manager/blob/ccd2dd853321c3ac568736b0d66ab9d3ad494d7b/pkg/controller/cainjector/controller.go#L40


mengqiy:
PTAL

pkg/webhook/generator.go
```diff
@@ -0,0 +1,334 @@
...
+	if len(mutatingWebhooks) > 0 {
+		return &admissionregistration.MutatingWebhookConfiguration{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: metav1.GroupVersion{Group: admissionregistration.GroupName, Version: "v1beta1"}.String(),
+				Kind:       "MutatingWebhookConfiguration",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: o.mutatingWebhookConfigName,
+				Annotations: map[string]string{
+					"admissionwebhook.alpha.kubebuilder.io/ca-secret-name": "webhook-cert",
```
droot: Two comments:
 - We should add a comment explaining (or linking to an issue/doc) why this annotation so readers understand it.
 - annotation name "alpha" needs to come before admissionwebhooks ? `alpha.admissionwebhooks.kubebuilder.io/ca-secret-name`
 - "webhook-cert" is hardcoded ?

pkg/webhook/generator_test.go
```diff
@@ -0,0 +1,157 @@
...
+          protocol: TCP
+        volumeMounts:
+        - mountPath: /tmp/test-cert
+          name: cert
+          readOnly: true
+      volumes:
+      - name: cert
+        secret:
+          defaultMode: 420
+          secretName: webhook-secret
```
droot: Two concerns:
 - Looks like input for CA injections (secret-name, service name etc.) is split across the objects. It might make the cert controller implementation difficult. I am wondering if we can combine all these and apply it only on field. Annotation value could be then in JSON ? 
 - webhook-secret is hardcoded. We should think of a way to override it (by default ?)

pkg/webhook/writer.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Validate validates the input options.
+func (o *WriterOptions) Validate() error {
+	if _, err := o.inFs.Stat(o.InputDir); err != nil {
+		return fmt.Errorf("invalid input directory '%s' %v", o.InputDir, err)
+	}
+	return nil
+}
+
+func (o *WriterOptions) writeObjectsToDisk(objects ...runtime.Object) error {
```
droot: This can probably go in a util pkg. 

I wonder if we should have "k8s/ioutil" pkg which implements k8s object reader/writer :)

pkg/webhook/generator_test.go
mengqiy: > Looks like input for CA injections (secret-name, service name etc.) is split across the objects. It might make the cert controller implementation difficult.

For generating cert, everything it needs is in the service object.
For syncing CA, everything it needs is in the WebhookConfiguration object.
It should not make the controller hard to implement. Instead, it should be quite easy to implement it :)

> webhook-secret is hardcoded.

This one is not.

pkg/webhook/generator.go
mengqiy: I now uses `alpha.admissionwebhook.kubebuilder.io/ca-secret-name`.

> "webhook-cert" is hardcoded ?

fixed.

pkg/webhook/admission.go
```diff
@@ -0,0 +1,104 @@
...
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// admissionWebhook contains bits needed for generating a admissionWebhook Configuration
+type admissionWebhook struct {
+	// name is the name of the webhook
+	name string
+	// t is the webhook type, i.e. mutating, validating
+	t webhookType
```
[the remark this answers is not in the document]
mengqiy: Not sure if `typ` is better than using `t` or not.
It's a partial word which reads strange.

Why you think `typ` is better?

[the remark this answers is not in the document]
droot: Oh.. just wanted it be more descriptive from readability perspective. Suggested `typ` because `type` is probably reserved word.

pkg/webhook/generator_test.go
droot: Got it. You are right, those are decoupled annotations, so it should be simpler :)

pkg/webhook/generator_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/spf13/afero"
+)
+
+var expected = map[string]string{
+	"config/webhook/webhookmanifests.yaml": `apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  annotations:
+    alpha.admissionwebhook.kubebuilder.io/ca-secret-name: test-system/webhook-secret
```
droot: I am wondering how will these value in these annotations will play with kustomization ?

mengqiy:
> how well it will work with kustomize ?

Kustomize by default doesn't know how to update the secret name in the annotation.
@droot we had a brief discussion last week. The outcome is that we don't use kustomize to manage the serving cert secret. We hand off this work to cert-manager. If the secret is not ready, let the pod wait for the secret.

mengqiy:
Cherry-picker @DirectXMan12's commit which cleanup the vendor dir.
Resolved the conflicts caused by #142.

droot:
>The outcome is that we don't use kustomize to manage the serving cert secret. We hand off this work to cert-manager. If the secret is not ready, let the pod wait for the secret. 

Sounds good.

mengqiy:
@droot @DirectXMan12 We need to decide what format the annotations should look like.

I'm currently using 
[alpha.admissionwebhook.cert-manager.io: "true"](https://github.com/kubernetes-sigs/controller-tools/pull/136/files#diff-a234f042a1a9e047854cf40c268e80c4R31)
and
[alpha.service.cert-manager.io/serving-cert-secret-name: some-secret-name](https://github.com/kubernetes-sigs/controller-tools/pull/136/files#diff-a234f042a1a9e047854cf40c268e80c4R90)
in my code, since I don't remember what Solly were using in the PoC controller code.

Besides the annotation format, other code is ready for another round of review.

mengqiy:
PTAL
No big changes.
Rebased to address merge conflicts.

mengqiy:
Rebased.
Commit `Clean up vendor, remove dep on CR` cause merge conflicts, due to moving some packages around.

Issue #123, opened 2019-01-10
Format validation key specified for string typed field is being ignored


sslavic:
Following http://kubebuilder.netlify.com/beyond_basics/generating_crd.html#validation I tried to specify Format validation key for a string typed field e.g.
```
	// +kubebuilder:validation:Format=hostname
	Hostname string `json:"hostname"`
```
Generated CRD validation openAPIV3Schema included the property but format was ignored:
```
          properties:
            hostname:
              type: string
```
I expected 
```
          properties:
            hostname:
              type: string
              format: hostname
```

Bug seems to be in https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/crd.go#L279
where additionally `f` should be set to `props.Format` if one is set (i.e. non-empty string).

Affected version: "v0.1.8"

alegacy:
I would like to take a stab at fixing this issue if no one has already started.

PR #138, opened 2019-02-06
🐛 Fix OpenAPI type for metav1.Duration fields

munnerz:
Previously when generating a validation schema for metav1.Duration fields, the field would be marked as type `object` instead of `string` as set here: https://github.com/kubernetes/apimachinery/blob/4b4ea28f279089e0cfafcc2d47479a5bd1a0964f/pkg/apis/meta/v1/duration.go#L56

Similar to how controller-tools currently handles metav1.Time, I've added an exception for metav1.Duration so this will be correctly typed 😄

### DIFF
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -136,6 +136,12 @@ func (b *APIs) getTime() string {
 }`
 }
 
+func (b *APIs) getDuration() string {
+	return `v1beta1.JSONSchemaProps{
+    Type:   "string",
+}`
+}
+
 func (b *APIs) objSchema() string {
 	return `v1beta1.JSONSchemaProps{
     Type:   "object",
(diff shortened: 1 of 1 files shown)


munnerz:
Based on https://book.kubebuilder.io/beyond_basics/generating_crd.html#validation and #123, it looks like I *can* override Format. I don't think it's currently possible to override Type however?

munnerz:
You can see an example of the result of the change here: https://github.com/jetstack/cert-manager/pull/1322/commits/ce3d565d69383877f2d7375ae8740e0a4034ccc1

droot:
@munnerz I think supporting types like `metav1.Time` and `metav1.Duration` out-of-the-box is useful. Thanks for the PR :)

> This patch does work for now, but if I were to define any of my own types which have their own `OpenAPISchemaType` and `OpenAPISchemaFormat` methods, I'd be stuck.

I agree. So introducing `// +kubebuilder:validation:type` annotation and fixing issue #123 can help. 

@fanzhangio Would you be interested in adding support for `// +kubebuilder:validation:type` and fixing #123 ?

droot:
This is related https://github.com/kubernetes-sigs/kubebuilder/issues/301

PR #142, opened 2019-02-11
✨ Add support for CRD singular name

damemi:
This adds a `+kubebuilder:singular=foo` tag to specify singular names for CRDs if desired

### DIFF
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -217,6 +217,20 @@ func HasDocAnnotation(t *types.Type) bool {
 	return false
 }
 
+// hasSingular returns true if t is an APIResource annotated with
+// +kubebuilder:singular
+func hasSingular(t *types.Type) bool {
+	if !IsAPIResource(t) {
+		return false
+	}
+	for _, c := range t.CommentLines{
+		if strings.Contains(c, "+kubebuilder:singular"){
+			return true
+		}
+	}
+	return false
+}
+
 // IsUnversioned returns true if t is in given group, and not in versioned path.
 func IsUnversioned(t *types.Type, group string) bool {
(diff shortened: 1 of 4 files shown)


damemi:
On an unrelated note, the PR submission says to see VERSIONING.md for which icon to add, but I couldn't find any versioning file in this repo. So I have gone with sparkles based on looking at a few other PRs, if this is incorrect I can update.

DirectXMan12:
whoops.  [VERSIONING.md is here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/VERSIONING.md) (in controller-runtime).

pkg/internal/codegen/parse/util.go
```diff
@@ -311,6 +325,16 @@ func getCategoriesTag(c *types.Type) string {
 	return resource
 }
 
+// getSingularName returns the value of the +kubebuilder:singular tag
+func getSingularName(c *types.Type) string {
+	comments := Comments(c.CommentLines)
+	singular := comments.getTag("kubebuilder:singular", "=")
+	if len(singular) == 0 {
+		panic(errors.Errorf("Must specify +kubebuilder:singular comment for type %v", c.Name))
```
DirectXMan12: this error doesn't make any sense with how singular is used.  maybe reword to say "must specify a value to use with singular" or something?

sohankunkerkar: @damemi  could you please add one test case here for this change? You can add your test case here https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go#L97
and run this file to see if you're getting the proper o/p.
https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/generator/generator_test.go

damemi:
@DirectXMan12 @sohankunkerkar thanks, I've updated the error message and added a test case for this. Could either of you please take another look?

PR #143, opened 2019-02-13
:bug: Removes extraneous "format" key from array objects

### LINKED ISSUE #141: Should array schema include "format" key?
Does it make sense to include the "format" key within a schema when the type is "array"?   For example, I created an array of strings and wanted those strings to be formatted as IPv4 addresses.  After fixing #123  to ensure that "format" keys are included for strings I found that the "format" key was included in both the array definition and the "items" definitions.   I cannot think of a reason why the outer format would be required.

With this in my types.go struct:

```
	// +kubebuilder:validation:Format=ipv4
	// list of IP addresses
	Addresses []string `json:"addresses,omitempty"`
```

I get this in my CRD yaml (notice the "format: ipv4" at both the array and items levels).  I am proposing to remove the one at the array level unless there is a reason for this that I have not yet uncovered:

```
        spec:
          properties:
            addresses:
              description: list of IP addresses
              format: ipv4
              items:
                format: ipv4
                type: string
              type: array
```

alegacy:
This commit removes the "format" key from array objects since any format annotations on arrays are assumed to be applicable to the array items rather than to the array itself.  Byte arrays are handled as a special case since they are represented as byte strings in the CRD YAML rather than actual arrays.

Closes #141

I wasn't sure how long it would take to get PR #140 merged so I went ahead and created this PR on top of the fix for PR #140 since there was some overlap in the test data.   When PR #140 merges I will gladly rebase this PR to exclude the other commit.

### DIFF
--- a/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
+++ b/pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
@@ -79,6 +79,10 @@ type ToySpec struct {
 	// This is an IPv4 address.
 	// +kubebuilder:validation:Format=ipv4
 	Address string `json:"address"`
+
+	// This is a list of IPv4 addresses.
+	// +kubebuilder:validation:Format=ipv4
+	Addresses []string `json:"addresses"`
 }
 
 // ToyStatus defines the observed state of Toy
--- a/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
+++ b/pkg/crd/generator/testData/config/crds/fun_v1alpha1_toy.yaml
@@ -60,6 +60,12 @@ spec:
               description: This is an IPv4 address.
               format: ipv4
               type: string
+            addresses:
+              description: This is a list of IPv4 addresses.
+              items:
+                format: ipv4
+                type: string
+              type: array
             alias:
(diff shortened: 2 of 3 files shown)


alegacy:
pinging @DirectXMan12 , @droot and @pwittrock for review.

pkg/crd/generator/testData/pkg/apis/fun/v1alpha1/toy_types.go
```diff
@@ -76,6 +76,13 @@ type ToySpec struct {
...
 	// +kubebuilder:validation:Format=ipv4
 	Address string `json:"address"`
+
+	// This is a list of IPv4 addresses.
+	// +kubebuilder:validation:Format=ipv4
+	Addresses []string `json:"addresses"`
+
+	// This is a byte array.
+	Bytes []byte `json:"bytes"`
 }
```
alegacy: I just realized that "Comment" on line 83 is also a []byte array... i will remove this one.

DirectXMan12:
i've approved the other PR, so we'll wait till that merges and then we'll get this one merged.  Ping me if I seem to forget ;-)

alegacy:
@DirectXMan12  thanks!   I have refreshed this PR to rebase it to master now that the other PR has merged.

PR #144, opened 2019-02-13
✨Add flags to customize API path and contexts

damemi:
We have some use cases where API types are defined in a different package or location than the default `pkg/apis` currently hardcoded. For example, we have vendored types in an API library that correspond to CRDs in multiple different repos; this would allow us to establish a workflow such as 

1. Update API types in library
2. Bump that library in specific CRD repos
3. Generate new CRDs in those repos (wherein those types exist under, eg, `vendor/org/library/some-api/v1/`)

By specifying a known directory and context, we can still use this tool with only minor changes (the optional flags added in this PR) and this gives us more ability to customize it to our use cases.

I'm aware that we could accomplish something similar with `--output-dir`, by going in the opposite direction and running this tool from our API library while specifying the output as those specific repos. But I think controlling the CRD updates from each repo itself couples the CRD update with the corresponding, necessary dependency bump well.

### DIFF
--- a/cmd/crd/cmd/generate.go
+++ b/cmd/crd/cmd/generate.go
@@ -66,5 +66,7 @@ func GeneratorForFlags(f *flag.FlagSet) *crdgenerator.Generator {
 	// TODO: Do we need this? Is there a possibility that a crd is namespace scoped?
 	f.StringVar(&g.Namespace, "namespace", "", "CRD namespace, treat it as root scoped if not set")
 	f.BoolVar(&g.SkipMapValidation, "skip-map-validation", true, "if set to true, skip generating validation schema for map type in CRD.")
+	f.StringVar(&g.APIsPath, "apis-path", "pkg/apis", "the path to search for apis relative to the current directory")
+	f.StringVar(&g.APIsPkg, "apis-pkg", "", "package to consider as the project api package")
 	return g
 }
--- a/pkg/crd/generator/generator.go
+++ b/pkg/crd/generator/generator.go
@@ -42,6 +42,10 @@ type Generator struct {
 	Namespace         string
 	SkipMapValidation bool
 
+	// APIsPath and APIsPkg allow customized generation for Go types existing under directories other than pkg/apis
+	APIsPath string
+	APIsPkg  string
+
 	// OutFs is filesystem to be used for writing out the result
 	OutFs afero.Fs
 
@@ -80,13 +84,7 @@ func (c *Generator) ValidateAndInitFields() error {
 		c.Domain = crdutil.GetDomainFromProject(c.RootPath)
 	}
 
-	// Validate apis directory exists under working path
-	apisPath := path.Join(c.RootPath, "pkg/apis")
-	if _, err := os.Stat(apisPath); err != nil {
-		return fmt.Errorf("error validating apis path %s: %v", apisPath, err)
-	}
-
-	c.apisPkg, err = crdutil.DirToGoPkg(apisPath)
+	err = c.setAPIsPkg()
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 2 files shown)


damemi:
If I'm way out of left field with this, I am totally open to course corrections. I'm still figuring this tool out totally

cmd/crd/cmd/generate.go
```diff
@@ -66,5 +66,8 @@ func GeneratorForFlags(f *flag.FlagSet) *crdgenerator.Generator {
 	// TODO: Do we need this? Is there a possibility that a crd is namespace scoped?
 	f.StringVar(&g.Namespace, "namespace", "", "CRD namespace, treat it as root scoped if not set")
 	f.BoolVar(&g.SkipMapValidation, "skip-map-validation", true, "if set to true, skip generating validation schema for map type in CRD.")
+	f.StringVar(&g.ApisPath, "apis-path", "pkg/apis", "the path to search for apis")
```
DirectXMan12: nit: I generally prefer `API` instead of `Api` here.

cmd/crd/cmd/generate.go
```diff
@@ -66,5 +66,8 @@ func GeneratorForFlags(f *flag.FlagSet) *crdgenerator.Generator {
...
+	f.StringVar(&g.ContextDir, "context-dir", "./pkg/apis", "context directory for api paths")
+	f.StringVar(&g.ApisPkg, "apis-pkg", "", "package to consider as the project api package")
```
DirectXMan12: it's unclear to me why these need to be three different options.  Presumably, `apis-path` and `context-dir` can be a single option, and `apis-pkg` can be automatically inferred from `apis-path`, unless I'm missing something.

cmd/crd/cmd/generate.go
droot: here and in method names (`s/setApisPkg/setAPIsPkg`) as well (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

cmd/crd/cmd/generate.go
droot: I think we can combine the `apis-path` and `context-dir`. As per current code, I see `apis-pkg` is optional (inferred by default). Being able to override can be useful in some cases.

cmd/crd/cmd/generate.go
damemi: done

cmd/crd/cmd/generate.go
damemi: @droot exactly, I agree that those two can/should be combined (done)

damemi:
@droot @DirectXMan12 I've updated this PR, could either of you please take another look at this?

damemi:
bump, any more feedback on this?

pkg/crd/generator/generator.go
```diff
@@ -112,7 +109,7 @@ func (c *Generator) Do() error {
 		return fmt.Errorf("failed switching working dir: %v", err)
 	}
 
-	if err := b.AddDirRecursive("./pkg/apis"); err != nil {
+	if err := b.AddDirRecursive(fmt.Sprintf("./%s", c.APIsPath)); err != nil {
```
droot: If this is always relative, then we should update the help text for `api-path` commandline param to reflect that. 

I wish `filepath.Join(".", c.APIsPath)` worked for this case, but looks like doesn't.

nit: probably `"./" + c.APIsPath` reads better in this case.

damemi: @droot updated

pkg/crd/generator/generator.go
```diff
@@ -112,7 +109,7 @@ func (c *Generator) Do() error {
 		return fmt.Errorf("failed switching working dir: %v", err)
 	}
 
-	if err := b.AddDirRecursive("./pkg/apis"); err != nil {
+	if err := b.AddDirRecursive("./%s" + c.APIsPath); err != nil {
```
DirectXMan12: `"path/filepath".Join` is probably safer here.

damemi: that drops the leading `./` :(

droot:
@damemi I think this introduced regression https://github.com/kubernetes-sigs/controller-tools/issues/148 Can you pl. provide a fix ? (master branch is unusable because of that)

PR #147, opened 2019-02-20
Configuration options

### LINKED ISSUE #145: [RBAC] Allow to configure output file names
Would allow to execute multiple rbac generatios in the same project without having to rename the files after each operation in case all need to be ketp.

Would enable what was proposed in https://github.com/kubernetes-sigs/kubebuilder/issues/601#issuecomment-465510993.

/kind feature
/priority important-longterm
/sig api-machinery

### LINKED ISSUE #146: [RBAC] Allow to configure ServiceAccount
Would enable what was proposed in kubernetes-sigs/kubebuilder#601.

/kind feature
/priority important-longterm
/sig api-machinery

Adirio:
**What type of PR is this?**
/kind feature
/sig api-machinery
/priority important-longterm

**What this PR does / why we need it**:
Introduces some configuration options for RBAC generation that allows to bind cluster roles to custom service accounts instead of `"default"` in `"system"` and also allows to specify the name of the output files.

**Which issue(s) this PR fixes**:
Fixes #145 
Fixes #146

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
The default output files have been changed from "rbac_role.yaml" and "rback_role_binding.yaml" to "manager_role.yaml" and "manager_role_binding.yaml".
```

### DIFF
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -78,6 +78,7 @@ Usage:
 	f := cmd.Flags()
 	f.StringVar(&o.Name, "name", o.Name, "name to be used as prefix in identifier for manifests")
 	f.StringVar(&o.ServiceAccount, "service-account", o.ServiceAccount, "service account to bind the role to")
+	f.StringVar(&o.Namespace, "service-account-namespace", o.Namespace, "namespace of the service account to bind the role to")
 	f.StringVar(&o.InputDir, "input-dir", o.InputDir, "input directory pointing to Go source files")
 	f.StringVar(&o.OutputDir, "output-dir", o.OutputDir, "output directory where generated manifests will be saved")
 	f.StringVar(&o.RoleFile, "role-file", o.RoleFile, "output file for the role manifest")
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -77,6 +77,7 @@ Usage:
 
 	f := cmd.Flags()
 	f.StringVar(&o.Name, "name", o.Name, "name to be used as prefix in identifier for manifests")
+	f.StringVar(&o.ServiceAccount, "service-account", o.ServiceAccount, "service account to bind the role to")
 	f.StringVar(&o.InputDir, "input-dir", o.InputDir, "input directory pointing to Go source files")
 	f.StringVar(&o.OutputDir, "output-dir", o.OutputDir, "output directory where generated manifests will be saved")
 	f.StringVar(&o.RoleFile, "role-file", o.RoleFile, "output file for the role manifest")
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -76,9 +76,11 @@ Usage:
 	}
 
 	f := cmd.Flags()
-	f.StringVar(&o.Name, "name", o.Name, "Name to be used as prefix in identifier for manifests")
+	f.StringVar(&o.Name, "name", o.Name, "name to be used as prefix in identifier for manifests")
 	f.StringVar(&o.InputDir, "input-dir", o.InputDir, "input directory pointing to Go source files")
-	f.StringVar(&o.OutputDir, "output-dir", o.OutputDir, "output directory where generated manifests will be saved.")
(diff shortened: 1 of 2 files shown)


Adirio:
@droot I moved a `TODO` comment with your signature, can you check if it still applies?

pkg/rbac/manifests.go
```diff
@@ -30,33 +30,52 @@ import (
...
+	Labels         map[string]string
 }
 
 // SetDefaults sets up the default options for RBAC Manifest generator.
 func (o *ManifestOptions) SetDefaults() {
 	o.Name = "manager"
 	o.InputDir = filepath.Join(".", "pkg")
 	o.OutputDir = filepath.Join(".", "config", "rbac")
+	o.ServiceAccount = "default"
+	// TODO(droot): define Namespace as a const and share it with scaffold pkg.
```
DirectXMan12: this shouldn't apply any more, you're fine to remove it, I believe.  The scaffold package is in KubeBuilder "proper"

cmd/controller-gen/main.go
```diff
@@ -76,9 +76,13 @@ Usage:
 	}
 
 	f := cmd.Flags()
-	f.StringVar(&o.Name, "name", o.Name, "Name to be used as prefix in identifier for manifests")
+	f.StringVar(&o.Name, "name", o.Name, "name to be used as prefix in identifier for manifests")
+	f.StringVar(&o.ServiceAccount, "service-account", o.ServiceAccount, "service account to bind the role to")
+	f.StringVar(&o.Namespace, "service-account-namespace", o.Namespace, "namespace of the service account to bind the role to")
 	f.StringVar(&o.InputDir, "input-dir", o.InputDir, "input directory pointing to Go source files")
-	f.StringVar(&o.OutputDir, "output-dir", o.OutputDir, "output directory where generated manifests will be saved.")
+	f.StringVar(&o.OutputDir, "output-dir", o.OutputDir, "output directory where generated manifests will be saved")
```
DirectXMan12: we should probably lift this up to the root `controller-gen` command.

Adirio: Shouldn't that be another PR?

DirectXMan12: ah, yeah, whoops, you didn't add that, just changes the punctuation.  We can do that in a separate PR.

DirectXMan12:
@mengqiy any comments?

pkg/rbac/manifests.go
```diff
@@ -30,33 +30,51 @@ import (
...
+	Namespace      string
+	Labels         map[string]string
 }
 
 // SetDefaults sets up the default options for RBAC Manifest generator.
 func (o *ManifestOptions) SetDefaults() {
 	o.Name = "manager"
 	o.InputDir = filepath.Join(".", "pkg")
 	o.OutputDir = filepath.Join(".", "config", "rbac")
+	o.ServiceAccount = "default"
```
mengqiy: This may be problematic, since it will override it even I have specified something.
We should default only when it's empty.
I think we should check it for all these 5 fields.

Adirio: As of now, it is being used directly on an empty object so there is no issue: [link](https://github.com/kubernetes-sigs/controller-tools/blob/master/cmd/controller-gen/main.go#L60-L61).

What you suggest is more like a `FillDefaults` than a `SetDefaults`. I agree with you that it would be a better approach as in a future it may be used in a non-empty object. However I think that changing that behaviour is not responsability of this PR. I will open a new Issue to track that change.

cmd/controller-gen/main.go
Adirio: Can you open an issue to track this up-lift? I think you can justify better the reasons behind the suggested change.

pkg/rbac/manifests.go
mengqiy: Fair enough.
Thanks for opening the issue.

mengqiy: Hmm... In our current code, `SetDefaults` happens before the fields are set by the flags.
IIRC in k/k we usually set fields using flags first then do the defaulting.

PR #157, opened 2019-03-05
WIP: crd-gen: don't apply array+map comments to elements

sttts:
Before this PR tags of map and array instances were pulled into the element type generation. I.e. a minimum value above `[]int` will apply the minimum to the int, not the array. 

This looks good at first, but conflicts with tags which should be applied to the array, not the int. E.g. 1.14 will get `nullable` support (https://github.com/kubernetes/kubernetes/pull/74804). Nullable can be applied to the array and the element type int. Both makes lots of sense. We can mix tags for both use-cases.

Maybe we need something like special items tags?

### DIFF
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -328,7 +328,7 @@ var mapTemplate = template.Must(template.New("map-template").Parse(
 // parseMapValidation returns a JSONSchemaProps object and its serialization in
 // Go that describe the validations for the given map type.
 func (b *APIs) parseMapValidation(t *types.Type, found sets.String, comments []string) (v1beta1.JSONSchemaProps, string) {
-	additionalProps, result := b.typeToJSONSchemaProps(t.Elem, found, comments, false)
+	additionalProps, result := b.typeToJSONSchemaProps(t.Elem, found, nil, false)
 	additionalProps.Description = ""
 	props := v1beta1.JSONSchemaProps{
 		Type:        "object",
@@ -377,7 +377,7 @@ type arrayTemplateArgs struct {
 // parseArrayValidation returns a JSONSchemaProps object and its serialization in
 // Go that describe the validations for the given array type.
 func (b *APIs) parseArrayValidation(t *types.Type, found sets.String, comments []string) (v1beta1.JSONSchemaProps, string) {
-	items, result := b.typeToJSONSchemaProps(t.Elem, found, comments, false)
+	items, result := b.typeToJSONSchemaProps(t.Elem, found, nil, false)
 	items.Description = ""
 	props := v1beta1.JSONSchemaProps{
 		Type:        "array",


sttts:
@DirectXMan12 @droot this is concerning, i.e. not easy to fix.

sttts:
`// +k8s:kubebuilder:validation:items:minimum=42` maybe?

DirectXMan12:
Both make sense.  I'm tempted to say applying to the items will be more common, so making the special tag be the whole-list one is better from a "let's not break people as much" perspective.  From the "API purity" perspective, the per-items one makes sense, but I'm leaning towards not breaking users.

sttts:
We have that for maps and slices. Anything else?

It also depends on the tag at hand. `// nullable` I would expect to apply to the slice/map, not the values. Minimum, pattern etc. might be more sensible for the values.

Can we divide the tags into value and object tags?

droot:
> Can we divide the tags into value and object tags ?

tempting, but will there be a case where tag can be applied to both ? 

I like the idea of introducing special tag for `whole-list` that @DirectXMan12 suggested, though don't have good suggestions for it.

sttts:
The more I think about this, I think it is a bad idea to apply the standard field tags to map/slice values at all. Whatever we come up with, it's getting confusing and more complicated than it has to be. Consistency leads to simplicity.

Have we thought about dropping the `k8s:kubebuilder` prefix from the tags? This would give us the chance to support the old semantics without breaking users and in parallel introducing a clean variant.

Also there was discussion about switch the default of optional-vs-required. Such a switch would allow that as well.

DirectXMan12:
> Have we thought about dropping the `k8s:kubebuilder` prefix?

Having *a* prefix makes markers (I'm trying to get these to be called markers, since `tags` and `annotations` are super-overloaded :-P) a lot easier to match on.

I think the reason that we apply to list items is that most commonly, that was the operation that users wanted.  One other option here (for some future change) would be to allow/encourage associating data formats with *types*, such that a given type specified its OpenAPI format (using a marker).  Then, you obvious semantics on what's attached to what, and avoid the "I've updated the type but not all the uses" issues -- 

```go
type Other struct {
    // k8s:kubebuilder:nullable=true
    Addresses []IPAddress
}

// k8s:kubebuilder:format=ipv4
// k8s:kubebuilder:type=string
type IPAddress string
```

makes it a bit more obvious that the list is nullable, and the items have IPv4 format.  We'd still need per-item options (potentially with a new marker prefix), but those'd probably be used less often than just type-aliasing.

DirectXMan12:
We still need to decide how to deal with whole-list in a backwards-compatible way, but as for the "what's the obvious thing to do here", having this pattern would go a long way (IMO) towards avoiding the foot-guns of "no, I meant to say list items were IPv4 addresses, not the whole list".

sttts:
I like that IPAddress example much more. I totally understand that we have to stay compatible or change semantics in a non-breaking way, but I strongly think that apply to elements by default is the wrong direction and will cause more confusion in the long run.

How about `+k8s:openapi:...`? With `sigs.k8s.io/controller-tools` being kubebuilder independent, such a change would be a logical thing, wouldn't it?

DirectXMan12:
`+k8s:openapi` makes some amount of sense, and

> I like that IPAddress example much more. I totally understand that we have to stay compatible or change semantics in a non-breaking way, but I strongly think that apply to elements by default is the wrong direction and will cause more confusion in the long run.

Ack, I don't necessarily disagree.

> How about `+k8s:openapi:...`? With sigs.k8s.io/controller-tools being kubebuilder independent, such a change would be a logical thing, wouldn't it?

Yeah, something like that wouldn't be bad.  Technically, as per our versioning guidelines, we're allowed to break compatibility (with the appropriate version bump), so the other option is to break compatibility and specifically detect this case so that user will see it.  We can "typecheck" openapi specifications since most of these options are only valid on the individual item types.

DirectXMan12:
we don't do this in v0.2.0 -- closing

PR #158, opened 2019-03-05
crd-gen: fix intOrString type

sttts:
OneOf is a dangerous junctor (like not) because it is not monotonic. It leads to strange corner cases like `42` being a float and an integer. OneOf with `number` and `integer` will not validate `42`.

In general, rule of thumb is to avoid `OneOf` where possible.

Due to the bad properties of OneOf we cannot publish OneOf types via the swagger spec, i.e. we have to filter out those types in CRD openapi publishing (alpha in 1.14, https://github.com/kubernetes/kubernetes/pull/71192).

Luckily, `AnyOf` does the same for this purpose here and is the way to go for the upper reasons.

### DIFF
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -186,7 +186,7 @@ func (b *APIs) typeToJSONSchemaProps(t *types.Type, found sets.String, comments 
 		}, b.objSchema()
 	case intOrString:
 		return v1beta1.JSONSchemaProps{
-			OneOf: []v1beta1.JSONSchemaProps{
+			AnyOf: []v1beta1.JSONSchemaProps{
 				{
 					Type: "string",
 				},
--- a/testData/config/crds/fun_v1alpha1_toy.yaml
+++ b/testData/config/crds/fun_v1alpha1_toy.yaml
@@ -118,10 +118,10 @@ spec:
               format: int32
               type: integer
             rook:
-              description: This is a newly added field. Using this for testing purpose.
-              oneOf:
+              anyOf:
               - type: string
               - type: integer
+              description: This is a newly added field. Using this for testing purpose.
             s3Log:
               description: Used this for testing fieldNames with number.
               properties:


DirectXMan12:
makes sense, and that's the intended behavior anyway.

/lgtm
/approve

DirectXMan12:
@sttts why do we filter out "oneof" fields in published OpenAPI specs?  When validating with objects, and the fieldsets are disjoint, it shouldn't be problematic, AFAICT.  It's only problematic for non-disjoint objects.

sttts:
@DirectXMan12 how do you know that the two sets are disjoint? For the handful of types we can hardcode that information. But OneOf applies to any set of schemas, of any complexity. Try to find out whether two regex patterns are disjoint....

Our main motivation for dropping OneOf is though that we want to drop other constructs inside of the sub-schemas (for similar reason or just because we don't want to leak information or don't support certain properties in kubectl). But during that process we may only make the published spec weaker. As soon as we are under a OneOf in the recursive algorithm, dropping fields does not make the whole spec necessarily weaker, it can get stronger by that. It's just hard to reason about anything as soon as you allow non-monotonic junctors when recursively diving into sub-schemas.

DirectXMan12:
@sttts how do you validate patterns like kubernetes-style enums, then?  I'd be tempted to describe that as (quoted directly from https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#discriminatorObject)

```go
  oneOf:
  - $ref: '#/components/schemas/Cat'
  - $ref: '#/components/schemas/Dog'
  - $ref: '#/components/schemas/Lizard'
  discriminator:
    propertyName: pet_type
```

(or without discriminator for old-school enums).  AnyOf won't work there.

DirectXMan12:
(new-style Kubernetes enums could be any-of + discriminator, I believe)

sttts:
Kube style enums could look like this:

```
        spec:
          properties: {"shared": {"type": "boolean"}}
          anyOf:
          - properties: {"a":{},"shared":{}}
            additionalProperties: false
            required: ["a"]
          - properties: {"b":{},"shared":{}}
            additionalProperties: false
            required: ["b"]
          - properties: {"c":{},"shared":{}}
            additionalProperties: false
            required: ["c"]
```

We might want to add some special tags to make this easier:

```
type Spec struct {
  Shared bool
  // +kube-enum
  A *T
  // +kube-enum
  B *T
  // +kube-enum
  C *T
}
```

sttts:
Similar constructs are possible with real string discriminators.

DirectXMan12:
ah, the `additionalProperities` prevents it from matching `{"a": "foo", "b":"bar", "shared": true}`?  Otherwise (without `additionalProperties`), that seems to match fine.

And yeah, for newer-style kube enums (e.g. HPA v2), the string discriminator should match up with the `discriminator` option and solve the issue.

PR #159, opened 2019-03-05
crd-gen: apply tags to special types, kube types and maps as well

sttts:
Special types and upstream kube types should get tag support as well. In 1.14 we will get `nullable` support (https://github.com/kubernetes/kubernetes/pull/74804) and that should be applied to those types as well.


droot:
@stts change looks good, pl. rebase and resolve the conflict.

sttts:
@droot done. ptal.

PR #161, opened 2019-03-05
:warning: Change to use the default rule set for Pluralize.

kaitoy:
This change improves pluralisation on resource names in CRDs (e.g. indexes -> indices).

[Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) that depends on this project uses the default rule set for controller code generation, and so it can generate CRDs and controllers inconsistently. 
The main motivation of this fix is to address the issue.

### DIFF
--- a/pkg/internal/codegen/parse/index.go
+++ b/pkg/internal/codegen/parse/index.go
@@ -71,7 +71,8 @@ func (b *APIs) parseIndex() {
 
 		// TODO: revisit the part...
 		if r.Resource == "" {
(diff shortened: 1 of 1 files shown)


mengqiy:
/lgtm
Each time run `controller-gen` in controller-tools will overwrite the existing CRD. IOW it may change the resources name. I'm not sure if there are users depending on the old naming behavior.
So this change may end up be a breaking change. I will change the title to reflect that.

PR #164, opened 2019-03-11
✨ update the kubernetes dependencies to the latest patch release of 1.13 -  1.13.4

tariq1890:
✨

It is in the best interests of any k8s API consumer that we always vendor in the latest patch release of a minor version.

As controller-tools acts as a source of k8s-dependencies for cluster-API and its providers, best practices mandate that we update the versions at the root so that they percolate down to the Cluster API providers. If this change is not accepted, we would be forced to use override deps.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -181,7 +181,7 @@
   version = "v2.2.2"
 
 [[projects]]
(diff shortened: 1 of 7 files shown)


tariq1890:
@caesarxuchao @pwittrock Requesting a review.

DirectXMan12:
xref kubernetes-sigs/controller-runtime#349

Issue #148, opened 2019-02-26
Failure: controller-gen/main.go crd processes all directories (including vendor)


ichekrygin:
It appears to be a regression as a result of #144 where `apis-path` and `apis-pkg` were added only in `cmd/crd/cmd/generate.go`, whereas it should be added in both `cmd/crd/cmd/generate.go` and `cmd/controller-gen/main.go`.

This results in failure when running via `controller-gen/main.go` since it attempts to process all directories under root path (including vendor) and ultimately fails on:
`panic: Cannot get version for unversioned type sigs.k8s.io/controller-tools/pkg/internal/codegen/parse.IsAPIResource`

mengqiy:
The offending PR was reverted in #169

PR #169, opened 2019-03-13
Revert "✨Add flags to customize API path and contexts"

droot:
Reverts kubernetes-sigs/controller-tools#144

Since this introduced regression https://github.com/kubernetes-sigs/controller-tools/issues/148


damemi:
@droot working on a fix to that regression

PR #170, opened 2019-03-14
🐛 Fix regression #148 from pr #144

### LINKED ISSUE #148: Failure: controller-gen/main.go crd processes all directories (including vendor)

damemi:
Fixes #148 

Adds the apis-path and apis-pkg flags (and their defaults) to `cmd/controller-gen/main.go` which in a couple local test runs fixed the problem described in #148. Would also like someone else to verify if possible

### DIFF
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -114,6 +114,8 @@ Usage:
 	f.StringVar(&g.Domain, "domain", "", "domain of the resources, will try to fetch it from PROJECT file if not specified")
 	f.StringVar(&g.Namespace, "namespace", "", "CRD namespace, treat it as cluster scoped if not set")
 	f.BoolVar(&g.SkipMapValidation, "skip-map-validation", true, "if set to true, skip generating OpenAPI validation schema for map type in CRD.")
+	f.StringVar(&g.APIsPath, "apis-path", "pkg/apis", "the path to search for apis relative to the current directory")
+	f.StringVar(&g.APIsPkg, "apis-pkg", "", "package to consider as the project api package")
 
 	return cmd
 }


damemi:
cc @droot @ichekrygin

Hopefully won't need https://github.com/kubernetes-sigs/controller-tools/pull/169 with this

cmd/controller-gen/main.go
```diff
@@ -114,6 +114,8 @@ Usage:
 	f.StringVar(&g.Domain, "domain", "", "domain of the resources, will try to fetch it from PROJECT file if not specified")
 	f.StringVar(&g.Namespace, "namespace", "", "CRD namespace, treat it as cluster scoped if not set")
 	f.BoolVar(&g.SkipMapValidation, "skip-map-validation", true, "if set to true, skip generating OpenAPI validation schema for map type in CRD.")
+	f.StringVar(&g.APIsPath, "apis-path", "pkg/apis", "the path to search for apis relative to the current directory")
+	f.StringVar(&g.APIsPkg, "apis-pkg", "", "package to consider as the project api package")
```
droot: crdgenerator.Generator is being used in another command as well `newAllSubCmd()`, so this needs to be done there as well. And `all` command is what gets invoked in kubebuilder project, so that is the most important. 

Thinking more about it, I think the root cause is we aren't setting defaults in crdgenerator.Generator. I am inclined towards reverting the original PR and fixing the root cause with more testing. WDYT ?

DirectXMan12:
@damemi can you make the title a bit more descriptive so that it's easier to look at in the git log history (e.g. `reintroduce api path flags with fix for directory traversal regression`)

damemi:
@DirectXMan12 this actually doesn't fix the problem it turns out, was talking to @droot about it and like he said, we're going to work on addressing the real root cause of this

/hold

damemi:
don't want this anymore, moved to https://github.com/kubernetes-sigs/controller-tools/pull/172

PR #172, opened 2019-03-18
Add APIsPath and APIsPkg crd generator variables w/ defaults

damemi:
Re-adds the changes that were reverted in https://github.com/kubernetes-sigs/controller-tools/pull/169, with the key change that `APIsPath` has a default of `pkg/apis` validated in the `"crdgenerator.Generator".ValidateAndInitFields()` method which should be called by all commands using that type


damemi:
@droot let me know if I missed anything, but it seems like just adding the default check in `setAPIsPkg` is what we needed here. Also we don't want the default to be the full `path.Join(c.RootPath, c.APIsPath)` because `c.RootPath` defaults to the full pwd, which breaks the call to `AddDirRecursive` because that can't import an absolute path. So it needs to have the leading `./`

droot:
> @droot let me know if I missed anything, but it seems like just adding the default check in `setAPIsPkg` is what we needed here. Also we don't want the default to be the full `path.Join(c.RootPath, c.APIsPath)` because `c.RootPath` defaults to the full pwd, which breaks the call to `AddDirRecursive` because that can't import an absolute path. So it needs to have the leading `./`

Agreed. Thanks for fixing it.

Issue #156, opened 2019-03-04
Merge existing and newly generated CRD OpenAPIv3 validation blocks


estroz:
Whenever I modify a type, or add/remove/update `+kubebuilder:validation:` directives from api's in `pkg/apis/<group>/<version>/*.go`, a CRD is generated with a validation block containing these changes. However if I add some validation field in a CRD that exists on-disk for an existing api, that field should be preserved in certain situations. Example:

On-disk CRD:
```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
...
spec:
  validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          type: string
...
        spec:
          properties:
            size:
              format: int32
              type: integer
            test:
              type: string
          required:
          - size
          type: object
...
```

Then write a value to some field manually, ex. `spec.validation.openAPIV3Shema.properties.spec.properties.size.title = "foo bar"`.

Desired CRD after generation:
```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
...
spec:
  validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          type: string
...
        spec:
          properties:
            size:
              format: int32
              type: integer
              title: foo bar # << added manually
            test:
              type: string
          required:
          - size
          type: object
...
```

As far as I can tell, the CRD generator will overwrite the `title` field, even though there is no `+kubebuilder:validation:Title` annotation according to [the kubebuilder book](https://book.kubebuilder.io/beyond_basics/generating_crd.html). Is it possible to support persisting fields added manually, i.e. merge old and new CRD's, if:
* There is no `+kubebuilder:validation:` directive for that field
* The CRD generator did not generate a value for that field

hasbro17:
/cc @mengqiy @DirectXMan12 @droot 
Any idea if it's possible to preserve manually added validation fields?

mengqiy:
It currently always rewrites the whole CRD.
It's hard to distinguish if a field is added by user manually or should be deleted. e.g. if there was `+kubebuilder:validation:Maximum=100` and later it is removed. The generator should be know this field should be removed and not treat it as user added.

IMO having `+kubebuilder:validation:title` should be better.

droot:
I agree with @mengqiy's reasoning here.  Adding `+kubebuilder:validation:title` in the `types.go` seems to be the right approach.

DirectXMan12:
[...]

PR #182, opened 2019-04-10
:sparkle: pkg/internal/codegen/parse: "Title" CRD validation annotation

estroz:
This PR implements the [`SchemaProps.Title`](https://godoc.org/github.com/go-openapi/spec#SchemaProps) field as an annotation: `+kubebuilder:validation:Title="some title"`, where the quoted string following `=` is stripped of surrounding quotes and interpreted literally as a field's title. This validation is not inherited by child props.

Reference: https://github.com/kubernetes-sigs/controller-tools/issues/156#issuecomment-479632914

/cc @mengqiy @DirectXMan12 @droot

### DIFF
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -267,6 +267,9 @@ var primitiveTemplate = template.Must(template.New("map-template").Parse(
     {{ if .MinLength -}}
     MinLength: getInt({{ .MinLength }}),
     {{ end -}}
+    {{ if .Title -}}
+    Title: "{{ .Title }}",
+    {{ end -}}
 }`))
 
 // parsePrimitiveValidation returns a JSONSchemaProps object and its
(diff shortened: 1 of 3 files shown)


pkg/internal/codegen/parse/crd.go
```diff
@@ -577,6 +581,20 @@ func getValidation(comment string, props *v1beta1.JSONSchemaProps) {
 		}
 	case "Format":
 		props.Format = parts[1]
+	case "Title":
+		title := strings.TrimSpace(strings.Join(parts[1:], " "))
+		if len(title) > 2 {
+			// Remove beginning and ending quote from non-empty strings.
+			if title[0] == '"' {
```
DirectXMan12: this accepts `Title=foo"`, which doesn't seem to be valid.  Also, `Title=f` should probably be valid.  I think we may need to hold on here a bit and define more format parsing semantics for our markers (I've got some stuff in the works that I'll post soon).

estroz: We could simply require string fields be wrapped in `""` and anything inside will be interpreted literally.

DirectXMan12: I've got parsing logic written.  We should be consistent across our different annotations.

estroz: Has it been merged?

DirectXMan12: not yet, it's part of a broader overhaul being tracked on the crdgenerator feature branch

DirectXMan12:
this is integrated into v0.2.0, so it should be all set now.

PR #184, opened 2019-04-11
Remove dependency on PROJECT file for CRD generator

damemi:
https://github.com/kubernetes-sigs/controller-tools/commit/626472cffd8eac5000961685a6c15b5942f6d36f created a dependency on a PROJECT file, this removes that dependency for the CRD generator specifically if these values are manually set

### DIFF
--- a/pkg/crd/generator/generator.go
+++ b/pkg/crd/generator/generator.go
@@ -71,12 +71,16 @@ func (c *Generator) ValidateAndInitFields() error {
 		}
 	}
 
-	// Validate PROJECT file
-	if !crdutil.PathHasProjectFile(c.RootPath) {
-		return fmt.Errorf("PROJECT file missing in dir %s", c.RootPath)
+	// Validate PROJECT file if Domain or Repo are not set manually
+	if len(c.Domain) == 0 || len(c.Repo) == 0 {
+		if !crdutil.PathHasProjectFile(c.RootPath) {
+			return fmt.Errorf("PROJECT file missing in dir %s", c.RootPath)
+		}
 	}
 
-	c.Repo = crdutil.GetRepoFromProject(c.RootPath)
+	if len(c.Repo) == 0 {
+		c.Repo = crdutil.GetRepoFromProject(c.RootPath)
+	}
 
 	// If Domain is not explicitly specified,
 	// try to search for PROJECT file as a basis.


sttts:
Compare the dirty workaround code above https://github.com/kubernetes-sigs/controller-tools/pull/183/commits/4b1e915c8bf2011b79529556ac3227dbebcdda4b#diff-d77ea52d9745ec923cd65d3e9ab0ac35R96, creating a PROJECT file.

In other words, the `crdgenerator.Generator` should be enough to run the generator, without parsing PROJECT. Such parsing can be done in crd-gen's plumbing code.

pkg/crd/generator/generator.go
```diff
@@ -71,12 +71,16 @@ func (c *Generator) ValidateAndInitFields() error {
 		}
 	}
 
-	// Validate PROJECT file
-	if !crdutil.PathHasProjectFile(c.RootPath) {
-		return fmt.Errorf("PROJECT file missing in dir %s", c.RootPath)
+	// Validate PROJECT file if Domain or Repo are not set manually
+	if len(c.Domain) == 0 || len(c.Repo) == 0 {
```
ingvagabund: `crd` sub-command does not allow `Repo` field to be set so this condition always falls to requiring `PROJECT` file to exist:

```
$ bin/controller-gen crd --help
Generate CRD manifests from the Type definitions in Go source files.
Usage:
# controller-gen crd [--domain k8s.io] [--root-path input_dir] [--output-dir output_dir]

Usage:
  controller-gen crd [flags]

Flags:
      --apis-path string      the path to search for apis relative to the current directory (default "pkg/apis")
      --apis-pkg string       the absolute Go pkg name for current project's api pkg.
      --domain string         domain of the resources, will try to fetch it from PROJECT file if not specified
  -h, --help                  help for crd
      --namespace string      CRD namespace, treat it as cluster scoped if not set
      --output-dir string     output directory, default to 'config/crds' under root path
      --root-path string      working dir, must have PROJECT file under the path or parent path if domain not set
      --skip-map-validation   if set to true, skip generating OpenAPI validation schema for map type in CRD. (default true)
```

Is that an intent? Or, rather should it be `if len(c.Domain) == 0 && len(c.Repo) == 0 {` instead?

Issue #137, opened 2019-02-01
Respect `+optional` comment tag during CRD generation


twiest:
According to https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required there are valid use cases for using the `+optional` comment tag without using `omitempty`.

Also note that according to the api-conventions.md doc linked above, `omitempty`'s behavior in Kubernetes "may change in [the] future".

Therefore when generating CRDs, fields should be considered optional if they have either the `+optional` comment tag or `omitempty`.

Currently only `omitempty` is being respected.

A Prototype for this feature can be seen here:
https://github.com/kubernetes-sigs/controller-tools/compare/master...twiest:add_optional_comment_tag_support?expand=1

sohankunkerkar:
@twiest Thanks for bringing this point to our attention. Your changes look good to me! Could you please file a PR for this and add a corresponding test case?

twiest:
@sohankunkerkar ok, sounds good. I'll look into adding that soon.

deads2k:
> According to https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required there are valid use cases for using the `+optional` comment tag without using `omitempty`.

Yeah, there definitely are.  Because CRD validation is tied up with the spec generation, the same forces that drove protobuf to give warnings about required fields apply.  A required field means that old clients may not be able to create new content because they don't know about the new fields being added.  It almost seems like the tag is backwards and we want to give explicit opt-in consideration for required since that represents potentially breaking behavior.

@liggitt I suspect you've given similar advice.

twiest:
@deads2k I created a PR based on your advice.

Since this is changing behavior there are some concerns. Would you mind weighing in on the [PR](https://github.com/kubernetes-sigs/controller-tools/pull/185) I created please?

alexkreidler:
As of now, [v0.2.0-beta.3](https://github.com/kubernetes-sigs/controller-tools/tree/v0.2.0-beta.3) includes PR #211, which adds a `// +kubebuilder:validation:[Required/Optional]` at both the package and field level, which although won't respect the `+optional` tag, should provide a good alternative for most users.

I hope the main `kubebuilder` integrates this ASAP so that we can manage required and optional fields better.

damemi:
[...]

PR #185, opened 2019-04-11
🐛 Add +required tag and make members optional by default.

twiest:
- [x] Change the default behavior for members to be optional.
- [x] Allow members to be set with a +required comment flag.
- [x] Filter out +require comment flags from CRD description field.
- [x] Change testData to ensure +required is working as expected.

This addresses issue #137.

### DIFF
--- a/pkg/internal/codegen/parse/crd.go
+++ b/pkg/internal/codegen/parse/crd.go
@@ -628,7 +628,7 @@ func (b *APIs) getMembers(t *types.Type, found sets.String) (map[string]v1beta1.
 			m, r := b.typeToJSONSchemaProps(member.Type, found, member.CommentLines, false)
 			members[name] = m
 			result[name] = r
-			if !strings.HasSuffix(strat, "omitempty") {
+			if hasRequired(member) {
 				required = append(required, name)
 			}
 		}
--- a/pkg/internal/codegen/parse/util.go
+++ b/pkg/internal/codegen/parse/util.go
@@ -223,8 +223,22 @@ func hasSingular(t *types.Type) bool {
 	if !IsAPIResource(t) {
 		return false
 	}
-	for _, c := range t.CommentLines{
-		if strings.Contains(c, "+kubebuilder:singular"){
+	for _, c := range t.CommentLines {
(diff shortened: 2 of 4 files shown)


damemi:
@sttts I think this would bring our fork in line with this repo

damemi:
bump, any thoughts on this? I'm in favor of this behavior but it's basically the opposite of how crd generation works now

pkg/internal/codegen/parse/crd.go
```diff
@@ -628,7 +628,7 @@ func (b *APIs) getMembers(t *types.Type, found sets.String) (map[string]v1beta1.
 			m, r := b.typeToJSONSchemaProps(member.Type, found, member.CommentLines, false)
 			members[name] = m
 			result[name] = r
-			if !strings.HasSuffix(strat, "omitempty") {
+			if hasRequired(member) {
```
droot: Changing the default behavior in this case will break all the existing projects which is unacceptable.

sttts:
We could make this package specific: look for doc.go upwards from the package of the current type until `// +optional=true` is found. Then `// +required` switch a field to being required.

twiest:
@droot I'm looking at rebasing this PR for master, but I have a couple of questions first.

1) If `+required` is too much of a behavior change, would a PR that simply respected `+optional` as well as `omitempty` as discussed in issue #137 be acceptable?

2) In `master`, I don't see any unit tests for CRD generation. Am I missing them or do they not exist yet?

sttts:
@twiest please sync with @damemi. We plan to implement a variant where you can switch the default of `+required/+optional` by putting the tag into 
  * doc.go
  * possibly the file (does that work in gengo?)
  * the type
  * the field

The default might stay as it is. But you can put `+optional` on the package and then `+required` is allowed to opt-in.

damemi:
Also, according to @DirectXMan12 version 0.2 is a breaking release anyway, so if we *do* decide to switch the behavior this would be an appropriate time to do it

twiest:
@sttts @damemi sounds great, thanks!

DirectXMan12:
closing as per #211

DirectXMan12:
thanks for your efforts though :-)

PR #189, opened 2019-04-13
add a bit more tests for CRD

mengqiy:


pkg/crd/v2/multiversion_test.go
```diff
@@ -0,0 +1,165 @@
...
+	// +kubebuilder:validation:Minimum=1
+
+	// Replicas is a number
+	Replicas int32 ` + "`" + `json:"replicas"` + "`" + `
+}
+`),
+			},
+			expectedCrdSpecs: map[schema.GroupKind][]byte{
+				schema.GroupKind{Group: "foo.bar.com", Kind: "Toy"}: []byte(`group: foo.bar.com
+names:
```
redborian: Just a thought, if we need to write more tests, we can think about having these in their own separate yaml files.

mengqiy: Generally, the best practice is that "unit tests should not access filesystem".
Echo some benefits mentioned in https://github.com/spf13/afero#using-afero-for-testing
```
Much faster than performing I/O operations on disk
Avoid security issues and permissions
Far more control. 'rm -rf /' with confidence
Test setup is far more easier to do
No test cleanup needed
```
We can revisit this if it grow to too large.

PR #194, opened 2019-04-22
:bug: Fix import path for references to testData/

rramkumar1:
It appears that the testData/ directory was moved to its current location in a different PR but the previous  path was not modified in a few files. This causes `go get sigs.k8s.io/controller-tools` to fail.

With this fix, the mentioned command will work. I noticed that even with these fixes, `make all` fails but I think that is out of scope for this PR.

### DIFF
--- a/testData/Dockerfile
+++ b/testData/Dockerfile
@@ -2,16 +2,16 @@
 FROM golang:1.10.3 as builder
 
 # Copy in the go src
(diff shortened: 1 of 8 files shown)


DirectXMan12:
/lgtm
/approve

I noticed this the other day too, and was meaning to fix it.  Thanks!

Issue #167, opened 2019-03-12
Allow generation of validation for types from k8s.io/api


DirectXMan12:
We currently explicitly skip types from `k8s.io/api`, for no apparent reason.  We should allow generating validation for such types, even if it's only structural (optional vs required, basic type information)

/kind feature
/priority important-soon

sttts:
/cc @liggitt 

This would force us into at least `IntOrString`.

liggitt:
it also will cause problems with the recursively-defined managedFields subtree

sttts:
@liggitt can you elaborate? Would we need a "stop generation here" tag?

In general, I would very much like to see a KEP about a unified OpenAPI tag language we can use in k8s.io/api types as well, and with that make embedded much easier. This is something we might want to target in the 1.16 time frame.

liggitt:
> @liggitt can you elaborate?

without `$ref` support, recursively defined structures can't render a complete schema for use inside a CRD

sttts:
> without `$ref` support, recursively defined structures can't render a complete schema for use inside a CRD

It's about schema generation here. Why shouldn't crd-gen be able to parse k8s.io/api types. It will flatten the output of course, without `$ref`.

liggitt:
How do you flatten a schema for a recursively defined type without use of $ref?

sttts:
@liggitt you drop it. But now I see what you mean. But we should find a way to cut that off from the schema, and allow 99% of the rest.

`+no-schema`

DirectXMan12:
For the moment, you can take a look at the behavior in the new version of controller-tools -- it explicitly overrides a few types, like `IntOrString`, since they don't have markers (tags) on them already.

DirectXMan12:
closing, fixed by the overrides section in v0.2.0-xyz

PR #190, opened 2019-04-16
:warning: Marker Parsing and Structure

DirectXMan12:
This re-implements our parsing and generation logic in a more structured form.
I've tried to write up extended guides in the godoc for each new package, so check that
out until I put together a cohesive guide.

TODO:

- [x] a couple of documentation spots
- [ ] design doc overview
- [ ] better tests
- [ ] iron out last few backwards-incompatibilities
- [x] support type-level schema modifications

PR #195, opened 2019-04-23
controller-gen crd: generate MetaData validation

ingvagabund:
Without validation of metadata field on the level of CRD definition,
cluster can accept a manifest that has invalid metadata field.
Object with invalid metadata field makes informer error with:

`sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1beta1.MachineSet: v1beta1.MachineSetList.Items: []v1beta1.MachineSet: v1beta1.MachineSet.Spec: v1beta1.MachineSetSpec.Template: v1beta1.MachineTemplateSpec.Spec: v1beta1.MachineSpec.ObjectMeta: v1.ObjectMeta.Labels: ReadMapCB: expect { or n, but found ", error found in #10 byte of ...|"labels":"ssd:\"true|..., bigger context ...|":"badmachineset"}},"spec":{"metadata":{"labels":"ssd:\"true\""},"providerSpec":{"value":{"ami":{"id|...`

Thus, making a controller inoperable until all invalid objects are deleted.

Partially resolving https://github.com/kubernetes-sigs/controller-tools/issues/167


ingvagabund:
After applying the patch my CRDs got extended with (shortened version):
```
diff --git a/install/0000_30_machine-api-operator_03_machineset.crd.yaml b/install/0000_30_machine-api-operator_03_machineset.crd.yaml
index c35546d..c15d69c 100644
--- a/install/0000_30_machine-api-operator_03_machineset.crd.yaml
+++ b/install/0000_30_machine-api-operator_03_machineset.crd.yaml
@@ -52,6 +31,149 @@ spec:
             submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
           type: string
         metadata:
+          properties:
+            annotations:
+              additionalProperties:
+                type: string
+              description: 'Annotations is an unstructured key value map stored with
+                a resource that may be set by external tools to store and retrieve
+                arbitrary metadata. They are not queryable and should be preserved
+                when modifying objects. More info: http://kubernetes.io/docs/user-guide/annotations'
+              type: object
+            clusterName:
+              description: The name of the cluster which the object belongs to. This
+                is used to distinguish resources with same name and namespace in different
+                clusters. This field is not set anywhere right now and apiserver is
+                going to ignore it if set in create or update request.
+              type: string
+            creationTimestamp:
+              description: 'CreationTimestamp is a timestamp representing the server
+                time when this object was created. It is not guaranteed to be set
+                in happens-before order across separate operations. Clients may not
+                set this value. It is represented in RFC3339 form and is in UTC.  Populated
+                by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata'
+              format: date-time
+              type: string
+            deletionGracePeriodSeconds:
+              description: Number of seconds allowed for this object to gracefully
+                terminate before it will be removed from the system. Only set when
+                deletionTimestamp is also set. May only be shortened. Read-only.
+              format: int64
+              type: integer
+            deletionTimestamp:
+              description: 'DeletionTimestamp is RFC 3339 date and time at which this
+                resource will be deleted. This field is set by the server when a graceful
+                deletion is requested by the user, and is not directly settable by
+                a client. The resource is expected to be deleted (no longer visible
+                from resource lists, and not reachable by name) after the time in
+                this field, once the finalizers list is empty. As long as the finalizers
+                list contains items, deletion is blocked. Once the deletionTimestamp
+                is set, this value may not be unset or be set further into the future,
+                although it may be shortened or the resource may be deleted prior
+                to this time. For example, a user may request that a pod is deleted
+                in 30 seconds. The Kubelet will react by sending a graceful termination
+                signal to the containers in the pod. After that 30 seconds, the Kubelet
+                will send a hard termination signal (SIGKILL) to the container and
+                after cleanup, remove the pod from the API. In the presence of network
+                partitions, this object may still exist after this timestamp, until
+                an administrator or automated process can determine the resource is
+                fully terminated. If not set, graceful deletion of the object has
+                not been requested.  Populated by the system when a graceful deletion
+                is requested. Read-only. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata'
+              format: date-time
+              type: string
+            finalizers:
+              description: Must be empty before the object is deleted from the registry.
+                Each entry is an identifier for the responsible component that will
+                remove the entry from the list. If the deletionTimestamp of the object
+                is non-nil, entries in this list can only be removed. +patchStrategy=merge
+              items:
+                type: string
+              type: array
+            generateName:
+              description: 'GenerateName is an optional prefix, used by the server,
+                to generate a unique name ONLY IF the Name field has not been provided.
+                If this field is used, the name returned to the client will be different
+                than the name passed. This value will also be combined with a unique
+                suffix. The provided value has the same validation rules as the Name
+                field, and may be truncated by the length of the suffix required to
+                make the value unique on the server.  If this field is specified and
+                the generated name exists, the server will NOT return a 409 - instead,
+                it will either return 201 Created or 500 with Reason ServerTimeout
+                indicating a unique name could not be found in the time allotted,
+                and the client should retry (optionally after the time indicated in
+                the Retry-After header).  Applied only if Name is not specified. More
+                info: https://git.k8s.io/community/contributors/devel/api-conventions.md#idempotency'
+              type: string
+            generation:
+              description: A sequence number representing a specific generation of
+                the desired state. Populated by the system. Read-only.
+              format: int64
+              type: integer
+            initializers:
+              description: An initializer is a controller which enforces some system
+                invariant at object creation time. This field is a list of initializers
+                that have not yet acted on this object. If nil or empty, this object
+                has been completely initialized. Otherwise, the object is considered
+                uninitialized and is hidden (in list/watch and get calls) from clients
+                that haven't explicitly asked to observe uninitialized objects.  When
+                an object is created, the system will populate this list with the
+                current set of initializers. Only privileged users may set or modify
+                this list. Once it is empty, it may not be modified further by any
+                user.
+              type: object
+            labels:
+              additionalProperties:
+                type: string
+              description: 'Map of string keys and values that can be used to organize
+                and categorize (scope and select) objects. May match selectors of
+                replication controllers and services. More info: http://kubernetes.io/docs/user-guide/labels'
+              type: object
+            name:
+              description: 'Name must be unique within a namespace. Is required when
+                creating resources, although some resources may allow a client to
+                request the generation of an appropriate name automatically. Name
+                is primarily intended for creation idempotence and configuration definition.
+                Cannot be updated. More info: http://kubernetes.io/docs/user-guide/identifiers#names'
+              type: string
+            namespace:
+              description: 'Namespace defines the space within each name must be unique.
+                An empty namespace is equivalent to the "default" namespace, but "default"
+                is the canonical representation. Not all objects are required to be
+                scoped to a namespace - the value of this field for those objects
+                will be empty.  Must be a DNS_LABEL. Cannot be updated. More info:
+                http://kubernetes.io/docs/user-guide/namespaces'
+              type: string
+            ownerReferences:
+              description: List of objects depended by this object. If ALL objects
+                in the list have been deleted, this object will be garbage collected.
+                If this object is managed by a controller, then an entry in this list
+                will point to this controller, with the controller field set to true.
+                There cannot be more than one managing controller. +patchMergeKey=uid
+                +patchStrategy=merge
+              items:
+                type: object
+              type: array
+            resourceVersion:
+              description: 'An opaque value that represents the internal version of
+                this object that can be used by clients to determine when objects
+                have changed. May be used for optimistic concurrency, change detection,
+                and the watch operation on a resource or set of resources. Clients
+                must treat these values as opaque and passed unmodified back to the
+                server. They may only be valid for a particular resource or set of
+                resources.  Populated by the system. Read-only. Value must be treated
+                as opaque by clients and . More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#concurrency-control-and-consistency'
+              type: string
+            selfLink:
+              description: SelfLink is a URL representing this object. Populated by
+                the system. Read-only.
+              type: string
+            uid:
+              description: 'UID is the unique in time and space value for this object.
+                It is typically generated by the server on successful creation of
+                a resource and is not allowed to change on PUT operations.  Populated
+                by the system. Read-only. More info: http://kubernetes.io/docs/user-guide/identifiers#uids'
+              type: string
           type: object
         spec:
           properties:
```

pkg/internal/codegen/parse/crd.go
```diff
@@ -470,6 +473,30 @@ func (b *APIs) parseObjectValidation(t *types.Type, found sets.String, comments
...
+// parseNativeObjectValidation returns a JSONSchemaProps object and its serialization in
+// Go that describe the validations for the given native object type.
+func (b *APIs) parseNativeObjectValidation(t *types.Type, found sets.String, comments []string, isRoot bool) (v1beta1.JSONSchemaProps, string) {
+	buff := &bytes.Buffer{}
+	props := v1beta1.JSONSchemaProps{
+		Type:        "object",
+		Description: parseDescription(comments),
+	}
+
+	for _, l := range comments {
```
frobware: How about `c` for comment?

pkg/internal/codegen/parse/crd.go
```diff
@@ -470,6 +473,30 @@ func (b *APIs) parseObjectValidation(t *types.Type, found sets.String, comments
...
+		getValidation(l, &props)
+	}
+
+	m, result, required := b.getMembers(t, found)
+	props.Properties = m
+	props.Required = required
+
+	if err := objectTemplate.Execute(buff, objectTemplateArgs{props, result, required, isRoot}); err != nil {
+		log.Fatalf("%v", err)
```
frobware: Why doesn't the function return `err` and let the caller decide what the policy should be?

Though I later noticed that `typeToJSONSchemaProps()` doesn't return an error. hmm.

pkg/internal/codegen/parse/crd.go
```diff
@@ -188,9 +188,12 @@ func (b *APIs) typeToJSONSchemaProps(t *types.Type, found sets.String, comments
 	case quantity:
 		specialTypeProps.Type = "string"
 		return specialTypeProps, b.getQuantity()
-	case meta, unstructured, rawExtension:
+	case unstructured, rawExtension:
 		specialTypeProps.Type = "object"
 		return specialTypeProps, b.objSchema()
+	case meta:
+		// Validating ObjectMeta as part of https://github.com/kubernetes-sigs/controller-tools/issues/167
+		return b.parseNativeObjectValidation(t, found, comments, isRoot)
```
spangenberg: If you pass `specialTypeProps` here, could you remove the the re-execution of `parseDescription` and `getValidation` in `parseNativeObjectValidation` (line 480-487)?

pkg/internal/codegen/parse/crd.go
ingvagabund: The new method is just copy-paste of `parseObjectValidation` without the check for `strings.HasPrefix(t.Name.String(), "k8s.io/api")`.

sttts: Don't like this special casing of ObjectMeta. Our mechanisms should work in the general case.

sttts:
IMO we should start allowing `$ref`erences to the official ObjectMeta definition from the apiserver.

/cc @liggitt

ingvagabund:
/hold

getting `validation failure list:\nspec.metadata.creationTimestamp in body must be of type string: \"null\""` in my environment

ingvagabund:
So, CRD's creationTimestamp field has this schema:
```
creationTimestamp:
              description: 'CreationTimestamp is a timestamp representing the server
                time when this object was created. It is not guaranteed to be set
                in happens-before order across separate operations. Clients may not
                set this value. It is represented in RFC3339 form and is in UTC.  Populated
                by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata'
              format: date-time
              type: string
```

notice `type: string`. Though, given `CreationTimestamp` field in the go code is not set to anything, it defaults to null. Thus, even `Time` k8s data type needs to be generated properly.

sttts:
@ingvagabund possibly nullable will help (though I am not 100% sure right now it applies to strings).

ingvagabund:
Using
```
            creationTimestamp:
              description: 'CreationTimestamp is a timestamp representing the server
                time when this object was created. It is not guaranteed to be set
                in happens-before order across separate operations. Clients may not
                set this value. It is represented in RFC3339 form and is in UTC.  Populated
                by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata'
              anyOf:
              - type: string
                format: date-time
              - enum:
                - null
```
might actually work.

sttts:
>  anyOf:
>               - type: string
>                 format: date-time
>               - enum:
>                 - "null"

I doubt that works. "null" is a string.

But you can try to just drop the whole type specification for that field. `nullable` would be the right choice with 1.14 (and in OpenShift with our controller-tools backport).

ingvagabund:
> I doubt that works. "null" is a string.

Based on `Time.MarshalJSON` nil object is marshaled into "null" string:

```
// MarshalJSON implements the json.Marshaler interface.
func (t Time) MarshalJSON() ([]byte, error) {
	if t.IsZero() {
		// Encode unset/nil objects as JSON's "null".
		return []byte("null"), nil
	}
	buf := make([]byte, 0, len(time.RFC3339)+2)
	buf = append(buf, '"')
	// time cannot contain non escapable JSON characters
	buf = t.UTC().AppendFormat(buf, time.RFC3339)
	buf = append(buf, '"')
	return buf, nil
}
```

liggitt:
do we want CRD creators specifying a full metadata schema that duplicates (and potentially conflicts with) the metadata schema applied in-server?

> Without validation of metadata field on the level of CRD definition, cluster can accept a manifest that has invalid metadata field.

Can you give more details about this? That should have been resolved in https://github.com/kubernetes/kubernetes/pull/64267 in 1.11

ingvagabund:
> Can you give more details about this?

Machineset CRD [1] allows to create a machineset object with the following `metadata` field:
```
    metadata:
      labels:
        machine.openshift.io/cluster-api-machineset: badmachineset
    spec:
      metadata:
        labels: ssd:"true"
```

[1] https://github.com/kubernetes-sigs/cluster-api/blob/master/config/crds/cluster_v1alpha1_machineset.yaml#L68-L70

liggitt:
ah, nested spec.metadata which is expected to follow the same validation, not the top-level metadata (which the server validates/enforces)

liggitt:
this hasn't updated to 1.14-levels yet. when it does, what will the recursively-defined `metadata.managedFields` field produce?

liggitt:
> what will the recursively-defined `metadata.managedFields` field produce?

seems like without $ref support, that field is inexpressible

/cc @apelisse

apelisse:
Yeah, in that case specifically, I don't think it makes much sense to validate beside "should not bet set at all". No one should have `managedFields` outside top-level metadata I think.

DirectXMan12:
/hold

this functionality is already part of the next-gen controller-tools WIP PR (#190), as a consequence of recursive ref traversal.  Can you move the discussion on `metadata.managedFields` to an issue?

sttts:
Related https://github.com/kubernetes/kubernetes/pull/77653.

/close

PR #196, opened 2019-04-26
🏃 Replace github.com/ghodss/yaml with sigs.k8s.io/yaml

tamalsaha:


pkg/crd/generator/generator.go
```diff
@@ -32,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-tools/pkg/internal/codegen"
 	"sigs.k8s.io/controller-tools/pkg/internal/codegen/parse"
 	"sigs.k8s.io/controller-tools/pkg/util"
+	"sigs.k8s.io/yaml"
```
DirectXMan12: this block is generally intended to be for other packages within this module, so `"sigs.k8s.io/yaml"` still belongs in the second block

pkg/internal/codegen/parse/parser.go
```diff
@@ -26,7 +26,7 @@ import (
 
 	"github.com/pkg/errors"
 	rbacv1 "k8s.io/api/rbac/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
DirectXMan12: this is a weird goimports bug -- make sure you're up-to-date

tamalsaha:
@DirectXMan12 , I have addressed your comments. PTAL.

PR #198, opened 2019-04-30
✨ Add plural name tag

damemi:
Similar to https://github.com/kubernetes-sigs/controller-tools/pull/142, this adds support for a `+kubebuilder:plural=foos` tag for CRD generation


sohankunkerkar:
@damemi you can refer this code where this logic has been handled.
https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/internal/codegen/parse/index.go#L130
and it is used [here](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go#L67)

PR #199, opened 2019-05-04
:warning: New Generation Framework

DirectXMan12:
This introduces a new marker (tag) parsing and generation framework that should help easy feature addition and longterm development, as well as improve the UX of using controller-tools with different project layouts, including the new kubebuilder scaffolding v2 layout in development.

The framework itself is broken up across several packages, each of which has an overview godoc page on how it works and the logic behind it.

The core framework is composed of:

- `pkg/loader`: file loading and low-level type traversal
- `pkg/markers`: marker parsing and high-level type traversal
- `pkg/genall`: managing and running generators from shared parsing state

Additionally, the following generators exist, all of which can be called directly as a "library"

- `pkg/crd`: CRD and OpenAPI schema generation (lower-level bits are standalone, and can be used directly)
- `pkg/deepcopy`: CRD-specific replacement for deepcopy-gen
- `pkg/webhook`: WebhookConfig generation
- `pkg/rbac`: ClusterRole generation

All generators are tied together in the new version of the `controller-gen` command.  The `controller-gen` command takes generator and output arguments using the same syntax as the markers, without the leading plus.  Run with `--help` for help and examples.  You can also use the `-w` flag to check which markers a given generator can consume (for example, `controller-gen crd -w` will show all the validation and versioning markers).


DirectXMan12:
Additional thanks to @redborian and @mengqiy for a bunch of the initial CRD generation revamp work that preceded the new framework.

DirectXMan12:
closing in favor of #200

PR #203, opened 2019-05-07
:bug: Fix incompatible CRDs

DirectXMan12:
This introduces two fixes for generated CRDs:

- populate the status slices so that they don't fail client-side validation
- allow producing a "trivially versioned" CRD that's compatible with Kubernetes 1.11+ instead of 1.13+
- make the `printcolumn` fields optional when appropriate

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"go/types"
 
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	crdmarkers "sigs.k8s.io/controller-tools/pkg/crd/markers"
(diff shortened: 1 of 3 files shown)


pkg/crd/gen.go
```diff
@@ -70,6 +79,32 @@ func (Generator) Generate(ctx *genall.GenerationContext) error {
 	return nil
 }
 
+// toTrivialVersions stips out all schemata except for the storage schema,
```
mengqiy: > stips

typo?

pkg/crd/spec.go
```diff
@@ -137,5 +137,11 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind) {
 		packages[0].AddError(fmt.Errorf("CRD for %s has no storage version", groupKind))
 	}
 
+	// NB(directxman12): CRD's status doesn't have omitempty markers, which means things
+	// get serialized as null, which causes the validator to freak out.  Manually set
+	// these to empty till we get a better solution.
+	crd.Status.Conditions = []apiext.CustomResourceDefinitionCondition{}
```
sttts: An upstream PR would be helpful.

PR #186, opened 2019-04-11
Bump k8s libraries to 1.14.1

damemi:
This would allow us to bring in new features such as `+nullable` (see https://github.com/kubernetes-sigs/controller-tools/pull/180)

PR #204, opened 2019-05-08
:warning: Bump k8s libraries to 1.14.1

sttts:
https://github.com/kubernetes-sigs/controller-tools/pull/186 with another dep run.


sttts:
@DirectXMan12 @droot @pwittrock  any chance to merge this?

DirectXMan12:
We don't have our e2e tests that run in CT specifically back in shape yet.  Have you tested this with a KB sample project yet?  Flattening shouldn't break on the recursive structure in metadata, but I'd rather have a test in place.

damemi:
I tested this against the [kubebuilder project-v2](https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v2) test project using `./controller-gen paths=./api/...` and didn't seem to have any problems, and I got the same set of changes as running from controller-tools master (a couple generated CRDs and an updated zz_deepcopy.go).

The only difference in the generated output was that the fields from master
```
status:
  conditions: []
  storedVersions: []
```
got changed to `null` in this branch:
```
status:
  conditions: null
  storedVersions: null
```

DirectXMan12:
/lgtm
/approve

added special handling for managedFields to avoid issues with recursion (doesn't show up till the corresponding project is updated to 1.14, but wanted to head that off)

PR #205, opened 2019-05-09
🏃 task: enable running tests in prow

alexeldeib:
Modifies bash test scripts to install missing dependencies and allow for running inside prow. I've validated these locally using Phaino to run the Prow jobs.


alexeldeib:
This one isn't really necessary since we're building a different base image. Doesn't really hurt to help users install dependencies, though...

If you find it useful feel free to merge, else happy to close.

DirectXMan12:
eh, I'm a little nervous about curl pipe bash

alexeldeib:
@DirectXMan12 whoops, thought I updated this in both places. Fixed.

alexeldeib:
Is controller-tools fully on go modules? I was wondering if I can simply `GO111MODULE=on go get golangci@v1.16.0` or if I should do the `go get -d`, `git checkout` dance, but it seems like some parts of the script already are dependent on modules?

alexeldeib:
Nevermind, resolved.

test.sh
```diff
@@ -97,12 +93,41 @@ function setup_envs {
...
+
+function fetch_go_tools {
+  header_text "Checking for golangci-lint"
+  if ! is_installed golangci-lint; then
+    header_text "Installing golangci-lint"
+    GO111MODULE=off go get -d github.com/golangci/golangci-lint/cmd/golangci-lint
+    cd $(go env GOPATH)/src/github.com/golangci/golangci-lint
+    LATEST=v1.15.0
+    git checkout $LATEST
+    go install -ldflags="-X main.version=$LATEST" ./cmd/golangci-lint
```
DirectXMan12: can't we just do `GO111MODULE=on go get github.com/golangci/golangci-ling/cmd/golangci-ling@v1.15.0` here/

alexeldeib: I think when I opened these PRs was before going all-in in modules so I wanted it to work for people not using them. I’ll change this. 

I’ve been a bit slow because of ongoing business travel, but I’ll also update the dockerfile for the test image to go in GCR.

alexeldeib: Err, oops, this is exactly what I said above >.< but yeah, as long as you’re fine with everything depending on modules there’s no problem.

alexeldeib: Flipped it back to use modules.

test.sh
```diff
@@ -97,12 +93,36 @@ function setup_envs {
...
+  return 1
+}
+
+function fetch_go_tools {
+  header_text "Checking for golangci-lint"
+  if ! is_installed golangci-lint; then
+    GO111MODULE=on go get -d github.com/golangci/golangci-lint/cmd/golangci-lint@v1.15.0 
+  fi
+
+  DEP_LATEST=$(git describe --abbrev=0 --tags)
```
DirectXMan12: ??

alexeldeib: bleh, I swear I can write a simple bash script...This is supposed to be for installing the latest tagged version of dep.

test.sh
```diff
@@ -97,12 +93,39 @@ function setup_envs {
...
+  if ! is_installed golangci-lint; then
+    GO111MODULE=on go get -d github.com/golangci/golangci-lint/cmd/golangci-lint@v1.15.0 
+  fi
+
+
+  header_text "Checking for dep"
+  if ! is_installed dep; then
+    header_text "Installing dep"
+    GO111MODULE=off go get -d -u github.com/golang/dep
+    cd $(go env GOPATH)/src/github.com/golang/dep
```
DirectXMan12: I don't think we should pull latest here -- I think we should just download the a specific dep version as per `GOOS` and `GOARCH` -- it makes things more reproducible.

alexeldeib:
dropped the ball here and someone else is taking this up, closing.

PR #208, opened 2019-05-14
:sparkles: write multiple webhooks in one file and add defaulting

mengqiy:
This PR does 2 things:
- support writing multiple webhooks in one file
- add defaulting to admission webhook configurations
- work around the "required" caBundle field

### DIFF
--- a/pkg/genall/genall.go
+++ b/pkg/genall/genall.go
@@ -83,27 +83,29 @@ type GenerationContext struct {
 	InputRule
 }
 
-// WriteYAML writes the given object out, serialized as YAML, using the
+// WriteYAML writes the given objects out, serialized as YAML, using the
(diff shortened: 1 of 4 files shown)


pkg/genall/genall.go
```diff
@@ -83,27 +83,29 @@ type GenerationContext struct {
...
-		return err
-	}
-	if n < len(yamlContent) {
-		return io.ErrShortWrite
+	for _, obj := range objs {
+		yamlContent, err := yaml.Marshal(obj)
+		if err != nil {
+			return err
+		}
+		n, err := out.Write(append([]byte("---\n"), yamlContent...))
```
DirectXMan12: I think we may need an extra `\n---\n` to be safe here

PR #209, opened 2019-05-14
:bug: Fix panic on missing import in ref checker

DirectXMan12:
The ref checker would panic on missing imports before.  Now, we just add
an error and continue instead.


DirectXMan12:
@droot @mengqiy

PR #210, opened 2019-05-15
✨ controller-tools-maintainers is an empty alias

spiffxp:
I'm doing some org/repo-wide OWNERS parsing, an empty alias is not valid


spiffxp:
ref: encountered while doing analysis for https://github.com/kubernetes/community/issues/1673

PR #211, opened 2019-05-15
Add ability to set default field requirement status

damemi:
This introduces a `// +kubebuilder:validation:[Required/Optional]` marker which can be applied to both packages and fields. Placing it at the package level determines the default requirement for fields in that package (ie, a package marked with `// +kubebuilder:validation:Optional` will assume no fields are required unless specifically marked with `// +kubebuilder:validation:Required`).

If no package-level tag is set, this defaults to the current behavior of assuming all fields not tagged with `omitempty`, `inline`, or (now) `// +kubebuilder:validation:Optional` are required, so this change doesn't break current use cases.

### DIFF
--- a/pkg/crd/parser.go
+++ b/pkg/crd/parser.go
@@ -151,7 +151,15 @@ func (p *Parser) NeedSchemaFor(typ TypeIdent) {
 	p.Schemata[typ] = apiext.JSONSchemaProps{}
 
 	schemaCtx := newSchemaContext(typ.Package, p)
-	schema := infoToSchema(schemaCtx.ForInfo(info))
+	ctxForInfo := schemaCtx.ForInfo(info)
+
+	pkgMarkers, err := markers.PackageMarkers(p.Collector, typ.Package)
+	if err != nil {
+		typ.Package.AddError(err)
+	}
+	ctxForInfo.PackageMarkers = pkgMarkers
+
+	schema := infoToSchema(ctxForInfo)
 
 	p.Schemata[typ] = *schema
 
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -66,6 +66,7 @@ type schemaContext struct {
 	info *markers.TypeInfo
 
 	schemaRequester schemaRequester
+	PackageMarkers  markers.MarkerValues
 }
 
 // newSchemaContext constructs a new schemaContext for the given package and schema requester.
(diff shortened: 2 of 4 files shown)


damemi:
/hold

might change this to cover both cases and let people set their default or keep the current default

damemi:
I've updated this with the ability to set a package-level `// +defaultOptional=true` tag which enables the `// +required` tag, otherwise using the original default behavior as described above

sttts:
Why another tag? Aren‘t +required/optional enough?

damemi:
@sttts you mean just re-use required/optional at the package level? Yeah, I guess we could do it that way too. However `+optional` doesn't exist right now, so we'd at least have to add that new marker. Come to think of it, the implementation of that would actually add 4 new markers though, (required/optional for package and required/optional for validation).

sttts:
more markers is fine (in implementation). But for the users, it looks like one, right? just different levels.

damemi:
@sttts Updated this to have a `// +kubebuilder:validation:[Required/Optional]` marker at both package and field level. Note that I don't think this will carry through to sub-packages. Also if neither is set, we default to the current setting of everything being required by default (which makes the package-level `required` tag unnecessary, but it could be good to keep for clarity)

damemi:
/hold cancel
@DirectXMan12 any time to take another look at this? Tested with a couple projects and it works for me

pkg/crd/schema.go
```diff
@@ -342,8 +342,26 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *v1beta1.JSO
...
+		defaultMode := "required"
+		if ctx.info.PackageMarkers.Get("kubebuilder:validation:Optional") != nil {
+			defaultMode = "optional"
+		}
+
+		switch defaultMode {
+		// if this package isn't set to optional default...
+		case "required":
+			// ...everything that's not inline, omitempty, or explicitly optional is required
+			if !inline && !omitEmpty && field.Markers.Get("kubebuilder:validation:Optional") == nil {
```
DirectXMan12: how does `// +optional` factor in here? we should respect that too

pkg/crd/parser.go
```diff
@@ -121,6 +121,7 @@ func (p *Parser) indexTypes(pkg *loader.Package) {
 			Name:    info.Name,
 		}
 
+		info.PackageMarkers = pkgMarkers
```
DirectXMan12: I'd just put this in the schema context instead.

damemi:
@DirectXMan12 updated with your feedback, can you please tag?

DirectXMan12:
force-pushed an update to register the `+optional` marker, and to make the `Required` and `Optional` markers only function on fields and packages, not on types (which doesn't really make sense).

Will merge once tests pass.

pkg/crd/schema.go
```diff
@@ -342,8 +343,27 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *v1beta1.JSO
...
+		defaultMode := "required"
+		if ctx.PackageMarkers.Get("kubebuilder:validation:Optional") != nil {
+			defaultMode = "optional"
+		}
+
+		switch defaultMode {
+		// if this package isn't set to optional default...
+		case "required":
+			// ...everything that's not inline, omitempty, or explicitly optional is required
+			// TODO(damemi): Actually add support for mythical `+optional` marker
```
damemi: we can remove this comment then

damemi:
Removed the TODO for `+optional`, tests running again...

PR #215, opened 2019-05-23
:sparkles: CRD storage version marker

droot:
This change adds support for storage version marker to indicate which version type to mark for storage.

Note: I would like to double check the name for the marker, so keeping it "do-not-merge".

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -31,6 +31,7 @@ var CRDMarkers = []*markers.Definition{
 	markers.Must(markers.MakeDefinition("kubebuilder:subresource:scale", markers.DescribesType, SubresourceScale{})),
 	markers.Must(markers.MakeDefinition("kubebuilder:printcolumn", markers.DescribesType, PrintColumn{})),
 	markers.Must(markers.MakeDefinition("kubebuilder:resource", markers.DescribesType, Resource{})),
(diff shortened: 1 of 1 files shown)


pkg/crd/markers/crd.go
```diff
@@ -31,6 +31,7 @@ var CRDMarkers = []*markers.Definition{
 	markers.Must(markers.MakeDefinition("kubebuilder:subresource:scale", markers.DescribesType, SubresourceScale{})),
 	markers.Must(markers.MakeDefinition("kubebuilder:printcolumn", markers.DescribesType, PrintColumn{})),
 	markers.Must(markers.MakeDefinition("kubebuilder:resource", markers.DescribesType, Resource{})),
+	markers.Must(markers.MakeDefinition("kubebuilder:conversion:storage", markers.DescribesType, StorageVersion{})),
```
mengqiy: Is `kubebuilder:crd:storage` better?

mengqiy: Users need to specify which version is storage version, even when they don't use conversion.

droot: agreed.

droot:
Updated the PR and using `+kubebuilder:storageversion` as the marker.

droot:
@mengqiy @DirectXMan12 Can you PTAL ? Thanks

Issue #214, opened 2019-05-22
Setting scope: Namespaced on generated CRD


kragniz:
Currently, generated CRD resources without any special annotations on resources have `spec.scope: ""` set. When applied to a cluster, this defaults to `spec.scope: Namespaced`. I would expect this to be the default output from controller-gen.

The command I'm running is:

`controller-gen crd:trivialVersions=true paths=./pkg/apis/... output:crd:dir="${output}"`

ironcladlou:
See https://github.com/kubernetes-sigs/controller-tools/pull/220

PR #220, opened 2019-06-04
✨ crd: add a resource scope marker

ironcladlou:
Add a crd resource scope marker to allow generation of cluster scoped resources.
This change is backwards compatible (absence of the scope marker results in an
empty scope).

This also allows for explicit scoping of namespaced resources should the user so
desire (fixes https://github.com/kubernetes-sigs/controller-tools/issues/214).

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -36,7 +36,6 @@ var CRDMarkers = []*markers.Definition{
 
 // TODO: categories and singular used to be annotations types
 // TODO: doc
-// TODO: nonNamespaced
 
 func init() {
 	AllDefinitions = append(AllDefinitions, CRDMarkers...)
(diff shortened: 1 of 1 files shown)


ironcladlou:
Whether this fixes https://github.com/kubernetes-sigs/controller-tools/issues/214 may be debatable. If https://github.com/kubernetes-sigs/controller-tools/issues/214 wants to change the default output to an explicit `Namespaced` value, this PR will not be a fix. I deliberately chose not to change the default output for this patch.

ironcladlou:
@kragniz 
>You've added the controller-gen binary here, this should be removed.

Oops — fixed, thanks!

>I still think namespaced/nonNamespaced should be made explicit in the CRD resource, but maybe something to follow up on

Can you elaborate more on what you're proposing?

ironcladlou:
@DirectXMan12 who needs to review this?

kragniz:
@ironcladlou I mean `scope: Namespaced` should be present in the generated CRD schema

ironcladlou:
@kragniz @DirectXMan12 
>I mean scope: Namespaced should be present in the generated CRD schema

Like this? d8f1681f91394b5017e850e9b3736a749afd9b66

PR #222, opened 2019-06-06
🐛 Generate deepcopy for all struct members

kragniz:
Previously, only the first member of a struct was copied with pass-by-reference.


DirectXMan12:
good catch

PR #224, opened 2019-06-13
:sparkles: whitelisting the namespace where the webhook server runs

mengqiy:
- whitelist the namespace the webhook server runs
- generate an empty file even when no rbac rules
  - Genearting an empty file is not harmful, and it is useful for using kustomize. When kustomize expects a file, an empty file can zmake it happy.

### DIFF
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -81,20 +81,21 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		}
 	}
 
-	if len(rules) == 0 {
-		return nil
+	var objs []interface{}
+	if len(rules) != 0 {
+		objs = append(objs, &rbacv1.ClusterRole{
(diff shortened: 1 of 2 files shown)


DirectXMan12:
This whitelists *all* controller-manager namespaces, no?  That's probably fine (it's why adding labels to a namespace can be considered a privileged operation), just worth noting.

mengqiy:
> This whitelists all controller-manager namespaces, no?

Yes, since they by default have the same labels.

mengqiy:
Skipping the namespaces with `control-plane: controller-manager` will skip not only the current controller-manager but also all of other controller-managers.
This issue should be not handled at controller-tools layer. It should be handled at kubebuilder layer. Ideally through a kustomize patch.
But due to the limitation of how patch works in k8s API and kustomize, we may need to document how to skip a namespace.

PR #226, opened 2019-06-17
:bug: Smarter allOf flattening/validation preservation

### LINKED ISSUE #217: flattening wipes out field validation on type aliases
Flattening causes type alias information to override field information, which currently wipes any field specific validation.  We'll want to fix this, but we'll need to somehow merge the two sets of validation.

In case of disjoint sets of validation types this is easy.  In case of overlap (e.g. two pattern fields, two enums, etc), it's a bit more unclear what to do.  The "most correct" would probably be to use `allOf`, but that's non-structural.

@sttts

DirectXMan12:
This improves our allOf flattening so that it progressively works up the
chain, pushing allOf down until it's no longer needed, or it's at the
most specific point.  Along the way, if two allOfs would be identical,
they'll be merged.

We then make use of this by converting our ref merger to just stick
both halves in an allOf, and then let the allOf flattener deal with it.

The previous behavior of flattening simple allOfs (like the ones to
embedded references) just "fall out" of the new logic.

In simple (non-overlapping) cases, we still end up with the
simplest-looking validation, and in complex cases, we're able to
preserve "structurality" (see the structure schema) KEP by dealing with
those fields explicitly when we need to.  However, just letting the
flattener do its thing often solves the issue.

Fixes #217


DirectXMan12:
We *really* need tests for this.  Might try to get to them tomorrow, but wanted to get this up for review in the mean time.

@sttts can you sanity check the behavior here?

pkg/crd/flatten.go
```diff
@@ -26,47 +27,176 @@ import (
...
+// pkg/loader.Package that we need to record errors in places were it might not
+// make sense to have a loader.Package
+type ErrorRecorder interface {
+	// AddError records that the given error occurred.
+	// See the documentation on loader.Package.AddError for more information.
+	AddError(error)
+}
+
+// isOrNil checks if val is nil if val is of a nillable type, otherwise,
+// it comparse val to valuInt (which should probably be the zero value).
```
sttts: nit: valInt

pkg/crd/flatten.go
```diff
@@ -26,47 +27,176 @@ import (
...
+func isOrNil(val reflect.Value, valInt interface{}, zeroInt interface{}) bool {
```
sttts: why the Int postfix?

sttts: A better name would be isZeroOrNil. But also that is actually wrong: we never compare with zero in case of nillable types.

pkg/crd/flatten.go
```diff
@@ -26,47 +27,176 @@ import (
...
+		return val.IsNil()
+	default:
+		return valInt == zeroInt
+	}
+}
+
 // flattenAllOfInto copies properties from src to dst, then copies the properties
 // of each item in src's allOf to dst's properties as well.
-func flattenAllOfInto(dst *apiext.JSONSchemaProps, src apiext.JSONSchemaProps) {
+func flattenAllOfInto(dst *apiext.JSONSchemaProps, src apiext.JSONSchemaProps, errRec ErrorRecorder) {
```
sttts: why not return an aggregated error? No need for a complex ErrorRecorder construct.

pkg/crd/flatten.go
sttts: why not make a `isZero` func? nil is the zero value of nillable types.

pkg/crd/flatten.go
```diff
@@ -26,47 +27,176 @@ import (
...
+		if isOrNil(srcField, srcInt, zeroInt) {
+			// nothing to copy from src, continue
+			continue
+		}
+
+		dstField := dstVal.Field(i)
+		dstInt := dstField.Interface()
+		if isOrNil(dstField, dstInt, zeroInt) {
+			// dst is empty, continue
+			dstField.Set(srcField)
```
sttts: this does not work for nullable.

```
type: array
allOf:
- type: array
   nullable: true
```
is not equivalent to
```
type: array
nullable: true
```
because `null` is valid for the latter, but not for the former.

pkg/crd/flatten.go
```diff
@@ -26,47 +27,176 @@ import (
...
+				errRec.AddError(fmt.Errorf("conflicting types in allOf branches in schema: %s vs %s", dstInt, srcInt))
+			}
+			// keep the destination value, for now
+		// TODO(directxman12): Default -- use field?
+		// TODO(directxman12):
+		// - Dependencies: if field x is present, then either schema validates or all props are present
+		// - AdditionalItems: like AdditionalProperties
+		// - Definitions: common named validation sets that can be references (merge, bail if duplicate)
+		case "AdditionalProperties":
+			// TODO(directxman12): handle the bool too
```
sttts: we don't support `false` if I remember right. `additionalSchema: true` is a no-op logically, but has influence on pruning.

pkg/crd/flatten.go
```diff
@@ -26,47 +27,176 @@ import (
...
+			dstRemVal.Field(i).Set(dstField)
+			// ...and clear the original
+			dstField.Set(zeroVal)
+		}
 	}
-	for _, propName := range src.Required {
-		dst.Required = append(dst.Required, propName)
+
+	if hoisted {
+		dst.AllOf = append(dst.AllOf, dstRemainder, srcRemainder)
```
sttts: this will potentially not return a structural schema. Types, nullable, etc., i.e. everything from https://github.com/kubernetes/kubernetes/blob/9ecc57ef880c475c052e4ab472307202499cf8d9/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/structural.go#L47 has to be pushed into the root. Those cannot be under `allOf`.

pkg/crd/flatten.go
sttts: this needs exhaustive tests.

pkg/crd/flatten.go
DirectXMan12: there's an issue with comparibility -- I original was just comparing against the zero type, but encountered some issues with non-comparibility -- reflect is a bit weird in this regard, although it's possible I made a mistake.

> why the Int postfix

`Int` is the `interface{}` form, `val` is the `reflect.Value` form.

pkg/crd/flatten.go
DirectXMan12: aggregated errors then need to be extracted later, and this is the pattern used almost literally everywhere else in controller-tools, since it allows us to attach better error data in many cases.

pkg/crd/flatten.go
DirectXMan12: that's what I thought, but wanted to check with you first.  good to confirm.

pkg/crd/flatten.go
DirectXMan12: right -- that's what the above does -- we only `hoist` for properties that aren't listed above, so `type`, `nullable`, etc should be handled there.

pkg/crd/flatten.go
DirectXMan12: yes, agreed :-)

pkg/crd/flatten.go
DirectXMan12: we can probably get away with just re-calling `Interface` here, though -- it'll be a bit of a performance hit, but it's prob not too big a deal.  Can profile after writing tests.

pkg/crd/flatten.go
sttts: I think with structural schemas we are saved here: nullable is not allowed below allOf.

pkg/crd/flatten.go
sttts: Towards structural schema, take a look at https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/validation.go#L63. You might want to call that somewhere, or even use the Structural type directly.

DirectXMan12:
Added some unit tests around allOf flattening, and and overall integration test of the whole package (give or take) using the cronjob tutorial files.

pkg/crd/flatten.go
DirectXMan12: huh?  How does that work?

damemi:
@DirectXMan12 any updates on this? I tested and this fixes the problem I mentioned

DirectXMan12:
hmm... that's a bug

DirectXMan12:
/hold

want a couple more tests

DirectXMan12:
fixed
/unhold

should have more detailed unit tests of the non-allof flattening, but will need to re-jigger things a bit for that, potentially.

DirectXMan12:
In the mean time, the integration test should be sufficient

DirectXMan12:
better unit tests for flattening pushed anyway.  This brings us to pretty decent test coverage overall.

pkg/crd/flatten.go
DirectXMan12: let's file a follow-up issue for that.

PR #228, opened 2019-06-19
🐛Add FieldOnlyMarkers to AllDefinitions and update Nullable

damemi:
Currently the `+optional` and `+nullable` markers don't work. I've fixed `optional` by making sure the `FieldOnlyMarkers` slice that was added in https://github.com/kubernetes-sigs/controller-tools/pull/211 actually gets added to `AllDefinitions`. In the current code, `+nullable` would actually only be recognized as `+kubebuilder:validation:Nullable=true`, so I'm updating this to be more consistent with what's expected for that marker.

However, I still can't get nullable to be generated in the output CRD, so I would appreciate some input on this

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -58,7 +58,6 @@ var ValidationMarkers = mustMakeAllWithPrefix("kubebuilder:validation", markers.
 	Enum(nil),
 	Format(""),
 	Type(""),
-	Nullable(false),
 )
 
 // FieldOnlyMarkers list field-specific validation markers (i.e. those markers that don't make
@@ -67,6 +66,7 @@ var FieldOnlyMarkers = []*markers.Definition{
 	markers.Must(markers.MakeDefinition("kubebuilder:validation:Required", markers.DescribesField, struct{}{})),
 	markers.Must(markers.MakeDefinition("kubebuilder:validation:Optional", markers.DescribesField, struct{}{})),
 	markers.Must(markers.MakeDefinition("optional", markers.DescribesField, struct{}{})),
+	markers.Must(markers.MakeDefinition("nullable", markers.DescribesField, Nullable{})),
 }
 
 func init() {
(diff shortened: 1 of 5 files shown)


damemi:
Want to note that in debugging this, by inspecting the schema props around here https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/schema.go#L378 I do see `nullable: true` being set on fields that have it, but still not showing up in the final generated output. Also if I manually set `propSchema.Nullable = true` around that area, I *do* see nullable in the generated output. But not if I put it in a conditional (ie, checking for `field.Markers.Get("nullable")`, which returns true and sets the schema, but not in the final output)

damemi:
Update: the nullable problem I saw is fixed by https://github.com/kubernetes-sigs/controller-tools/pull/226

pkg/crd/markers/validation.go
```diff
@@ -222,6 +224,6 @@ func (m Type) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
 func (m Type) ApplyFirst() {}
 
 func (m Nullable) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
-	schema.Nullable = bool(m)
+	schema.Nullable = true
```
sttts: why true? Can we write `nullable: false`?

damemi: I think currently `// +nullable` is only used to set `nullable: true`, and as we've seen a lack of this field already implies that it cannot be nullable. Also, if we do want to allow `// +nullable=false` then all other existing usages of the nullable tag would have to be updated to `// +nullable=true`, or they would raise a missing parameter error when being parsed.

sttts: ok, fine with me.

DirectXMan12: you could technically make this a bool pointer if you wanted that

DirectXMan12:
Adding a test case to the integration test data (recently merged, in `pkg/crd/testdata`).  Otherwise lgtm.

DirectXMan12:
added test case, instructions for updating the integration test.

/lgtm
/approve

PR #233, opened 2019-06-26
✨ Add namespace selector for webhooks

kragniz:
This allows users to add a namespace selector to webhooks, with the following syntax:

    // +kubebuilder:webhook:selectorLabelName=webhooks.example.com,selectorLabelValue=enabled

I'm not too convinced of this way of specifying label name and label value, open to other suggestions.

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -50,6 +50,9 @@ type Config struct {
 
 	Name string
 	Path string
+
+	SelectorLabelName  *string
+	SelectorLabelValue *string
 }
 
 // verbToAPIVariant converts a marker's verb to the proper value for the API.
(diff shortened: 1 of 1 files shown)


DirectXMan12:
I'm not convinced that this encourages good behavior -- in general, I think we should strive to generate configuration that's inherently tied to the controller -- for instance, the RBAC rules are inherently tied to the controller, since they need to match with the actual requests made.  The particular namespace for the webhook isn't generally inherently tied to the webhook implementation.

WDYT?

kragniz:
Yeah, this was the first thing I reached for since I was using markers for other things. I now think it's better to add this kind of configuration in another pass over the generated files with a different tool.

prometherion:
I can just give a brief overview of why a Namespace selector for webhook could work, giving a brief example of the open-source project I'm developing.

We have to implement clastix/capsule#1 that requires a validating webhook on Pod creation: tl;dr; checking if the Pod's container images registry is allowed for the Tenant requesting it.

These pods are going to run on specific namespaces, `capsule.clastix.io/tenant` labeled.

In addition to this, we have to use the marker `failurePolicy=fail` since we don't want potentially to let run untrusted images.

But what happens if, for any reason, the controller manager Pod fails and needs to be rescheduled? That would trigger a cascading failure since the controller pod couldn't be scheduled due to the failing webhook it has to handle.

Obviously, the solution is using the [namespaceSelector](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector) in order to filter down these requests and avoid this egg/chicken problem.

Having that marker would be easier for us, avoiding to put customization through `kustomize` during manifest generation and apply.

Issue #213, opened 2019-05-19
pkg/crd: arrays are not supported as map values


abursavich:
```
$ /Users/abursavich/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./api/..." output:crd:artifacts:config=config/crd/bases
/Users/abursavich/go/src/redacted/redacted_types.go:82:24: map values must be a named type, not *ast.ArrayType
```

https://github.com/kubernetes-sigs/controller-tools/blob/455fb28296658dc93252700128c1e28992e7d38e/pkg/crd/schema.go#L288-L296

In my case, I have an object that mirrors a `ConfigMap` with the following field:

```
BinaryData map[string][]byte `json:"binaryData,omitempty"`
```

abursavich:
@DirectXMan12

DirectXMan12:
good catch, we'll want to fix that.

/kind bug

damemi:
@DirectXMan12 based on your comments in https://github.com/kubernetes-sigs/controller-tools/pull/234/ will this only be allowed for byte slices? eg, if I have a `map[string][]string` that needs to have a named type?

DirectXMan12:
talked directly a bit  -- don't want to break existing folks, so we'll add string slice support too.  Don't really want to encourage maps though, as they're generally discouraged for "complex" datatypes (`[]byte` being an exception because its kinda-of a primitive in our APIs): https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

PR #234, opened 2019-06-27
Add support for arrays as map values into pkg/crd

haiyanmeng:
Fix: https://github.com/kubernetes-sigs/controller-tools/issues/213

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -291,6 +291,12 @@ func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *v1beta1.JSONSchemaPr
 		valSchema = localNamedToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
 	case *ast.SelectorExpr:
 		valSchema = namedToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
+	case *ast.ArrayType:
+		valSchema = arrayToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
+		if valSchema.Type == "array" && valSchema.Items.Schema.Type != "string" {
+			ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map values must be a named type, not %T", mapType.Value), mapType.Value))
+			return &v1beta1.JSONSchemaProps{}
+		}
 	default:
 		ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map values must be a named type, not %T", mapType.Value), mapType.Value))
 		return &v1beta1.JSONSchemaProps{}
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -78,6 +78,12 @@ type CronJobSpec struct {
 	// This is a pointer to distinguish between explicit zero and not specified.
 	// +optional
 	FailedJobsHistoryLimit *int32 `json:"failedJobsHistoryLimit,omitempty"`
+
+	// This tests byte slices are allowed as map values.
+	ByteSliceData map[string][]byte `json:"byteSliceData,omitempty"`
+
+	// This tests string slices are allowed as map values.
+	StringSliceData map[string][]string `json:"stringSliceData,omitempty"`
 }
 
 // use an explicit type marker to verify that apply-first markers generate properly
@@ -85,8 +91,13 @@ type CronJobSpec struct {
 // +kubebuilder:validation:Type=string
 // TotallyABool is a bool that serializes as a string.
 type TotallyABool bool
+
 func (t TotallyABool) MarshalJSON() ([]byte, error) {
-	if t { return []byte(`"true"`), nil } else { return []byte(`"false"`), nil}
+	if t {
+		return []byte(`"true"`), nil
+	} else {
+		return []byte(`"false"`), nil
+	}
 }
 func (t *TotallyABool) UnmarshalJSON(in []byte) error {
 	switch string(in) {
@@ -99,7 +110,6 @@ func (t *TotallyABool) UnmarshalJSON(in []byte) error {
 		return fmt.Errorf("bad TotallyABool value %q", string(in))
 	}
 }
-
 
 // ConcurrencyPolicy describes how the job will be handled.
 // Only one of the following concurrent policies may be specified.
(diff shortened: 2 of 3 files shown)


DirectXMan12:
PR & commit message need more description -- when I look back via `git commit`, I should be able to figure enough historical context w/o having to jump to github

DirectXMan12:
This is a bit tricky -- we generally *don't* allow this, except with byte slices.  I think we should just special-case this to preserve the effect that output types are conformant with kubernetes API conventions.

haiyanmeng:
@DirectXMan12 , limited the support only to `byte slices`. PTAL.

pkg/crd/schema.go
```diff
@@ -291,6 +291,12 @@ func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *v1beta1.JSONSchemaPr
 		valSchema = localNamedToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
 	case *ast.SelectorExpr:
 		valSchema = namedToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
+	case *ast.ArrayType:
+		valSchema = arrayToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
+		if valSchema.Type != "string" {
```
DirectXMan12: this needs a comment on why string works here, since it's not immediately obvious

DirectXMan12: also we'll want to add support for string slices, as per an addendum to the issue.

haiyanmeng: > this needs a comment on why string works here, since it's not immediately obvious
Would you mind of sharing the reason why string works here? Thanks!

pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
```diff
@@ -4452,6 +4458,25 @@ spec:
...
+                                  admission controller will reject Pod create requests
+                                  which have the overhead already set. If RuntimeClass
+                                  is configured and selected in the PodSpec, Overhead
+                                  will be set to the value defined in the corresponding
+                                  RuntimeClass, otherwise it will remain unset and
+                                  treated as zero. More info: https://git.k8s.io/enhancements/keps/sig-node/20190226-pod-overhead.md
+                                  This field is alpha-level as of Kubernetes v1.16,
+                                  and is only honored by servers that enable the PodOverhead
+                                  feature.'
+                                type: object
```
haiyanmeng: To be honest, I don't know why this got be added.

pkg/crd/testdata/cronjob_types.go
```diff
@@ -78,15 +78,26 @@ type CronJobSpec struct {
...
 // TotallyABool is a bool that serializes as a string.
 type TotallyABool bool
+
 func (t TotallyABool) MarshalJSON() ([]byte, error) {
-	if t { return []byte(`"true"`), nil } else { return []byte(`"false"`), nil}
+	if t {
+		return []byte(`"true"`), nil
+	} else {
+		return []byte(`"false"`), nil
+	}
```
haiyanmeng: This is just a reformatting automatically done by my vim.

haiyanmeng:
> can you add a test case for this in `pkg/crd/testdata`, please

Done.

haiyanmeng:
> can you add a test case for this in `pkg/crd/testdata`, please

This test was not happy with my change. However, the diff in https://travis-ci.org/kubernetes-sigs/controller-tools/jobs/553494323 is too complex to decode.

I also `$GOPATH/bin/controller-gen  crd paths=. output:dir=.` on the master branch, and it has diffs, which seems suprising:
```
haiyanmeng@haiyanmeng:testdata$ $GOPATH/bin/controller-gen  crd paths=. output:dir=.
haiyanmeng@haiyanmeng:testdata$ git diff
diff --git a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
index dbf4bd03..910e2665 100644
--- a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
+++ b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
@@ -4452,6 +4452,25 @@ spec:
                                   must match a node''s labels for the pod to be scheduled
                                   on that node. More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/'
                                 type: object
+                              overhead:
+                                additionalProperties:
+                                  type: string
+                                description: 'Overhead represents the resource overhead
+                                  associated with running a pod for a given RuntimeClass.
+                                  This field will be autopopulated at admission time
+                                  by the RuntimeClass admission controller. If the
+                                  RuntimeClass admission controller is enabled, overhead
+                                  must not be set in Pod create requests. The RuntimeClass
+                                  admission controller will reject Pod create requests
+                                  which have the overhead already set. If RuntimeClass
+                                  is configured and selected in the PodSpec, Overhead
+                                  will be set to the value defined in the corresponding
+                                  RuntimeClass, otherwise it will remain unset and
+                                  treated as zero. More info: https://git.k8s.io/enhancements/keps/sig-node/20190226-pod-overhead.md
+                                  This field is alpha-level as of Kubernetes v1.16,
+                                  and is only honored by servers that enable the PodOverhead
+                                  feature.'
+                                type: object
                               preemptionPolicy:
                                 description: PreemptionPolicy is the Policy for preempting
                                   pods with lower priority. One of Never, PreemptLowerPriority.
```

DirectXMan12:
That looks like we pulled in a newer version of kubernetes somehow.  Did you have `GO111MODULE=on` and were in the testdata directory?

haiyanmeng:
> That looks like we pulled in a newer version of kubernetes somehow. Did you have `GO111MODULE=on` and were in the testdata directory?

Thanks for pointing this out. Fixed it!

pkg/crd/schema.go
```diff
@@ -291,6 +291,12 @@ func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *v1beta1.JSONSchemaPr
 		valSchema = localNamedToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
 	case *ast.SelectorExpr:
 		valSchema = namedToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
+	case *ast.ArrayType:
+		valSchema = arrayToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
+		if valSchema.Type == "array" && valSchema.Items.Schema.Type != "string" {
```
mengqiy: Is it possible `valSchema.Items.Schema` may be nil?
If possible, we should check it. Otherwise, you can ignore this comment.

haiyanmeng: It is not possible if `valSchema.Type != "array"`

PR #235, opened 2019-06-27
Merge rules in config/rbac/role.yaml

haiyanmeng:
To minimize the generated rules in role.yaml, the following optimizations are added:
* remove duplicates from every field of rbac.Rule
* sort items in every field of rbac.Rule

This should fix https://github.com/kubernetes-sigs/kubebuilder/issues/743

### DIFF
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -23,6 +23,9 @@ limitations under the License.
 package rbac
 
 import (
+	"sort"
+	"strings"
+
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -44,8 +47,57 @@ type Rule struct {
 	URLs      []string `marker:"urls,optional"`
 }
 
+// ruleKey represents the resources and non-resources a Rule applies.
+type ruleKey struct {
+	Groups    string
+	Resources string
+	URLs      string
+}
+
+// key normalizes the Rule and returns a ruleKey object.
+func (r *Rule) key() ruleKey {
+	r.normalize()
+	return ruleKey{
+		Groups:    strings.Join(r.Groups, "&"),
+		Resources: strings.Join(r.Resources, "&"),
+		URLs:      strings.Join(r.URLs, "&"),
+	}
+}
+
+// addVerbs adds new verbs into a Rule.
+// The duplicates in `r.Verbs` will be removed, and then `r.Verbs` will be sorted.
+func (r *Rule) addVerbs(verbs []string) {
+	r.Verbs = removeDupAndSort(append(r.Verbs, verbs...))
+}
+
+// normalize removes duplicates from each field of a Rule, and sorts each field.
+func (r *Rule) normalize() {
+	r.Groups = removeDupAndSort(r.Groups)
+	r.Resources = removeDupAndSort(r.Resources)
+	r.Verbs = removeDupAndSort(r.Verbs)
+	r.URLs = removeDupAndSort(r.URLs)
+}
+
+// removeDupAndSort removes duplicates in strs, sorts the items, and returns a
+// new slice of strings.
+func removeDupAndSort(strs []string) []string {
+	set := make(map[string]bool)
+	for _, str := range strs {
+		if _, ok := set[str]; !ok {
+			set[str] = true
+		}
+	}
+
+	var result []string
+	for str := range set {
+		result = append(result, str)
+	}
+	sort.Strings(result)
+	return result
+}
+
 // ToRule converts this rule to its Kubernetes API form.
-func (r Rule) ToRule() rbacv1.PolicyRule {
+func (r *Rule) ToRule() rbacv1.PolicyRule {
 	// fix the group names first, since letting people type "core" is nice
 	for i, group := range r.Groups {
 		if group == "core" {
@@ -68,20 +120,33 @@ type Generator struct {
 func (Generator) RegisterMarkers(into *markers.Registry) error {
 	return into.Register(RuleDefinition)
 }
+
 func (g Generator) Generate(ctx *genall.GenerationContext) error {
-	var rules []rbacv1.PolicyRule
+	rules := make(map[ruleKey]*Rule)
 	for _, root := range ctx.Roots {
 		markerSet, err := markers.PackageMarkers(ctx.Collector, root)
 		if err != nil {
 			root.AddError(err)
 		}
 
-		for _, rule := range markerSet[RuleDefinition.Name] {
-			rules = append(rules, rule.(Rule).ToRule())
+		for _, markerValue := range markerSet[RuleDefinition.Name] {
+			rule := markerValue.(Rule)
+			key := rule.key()
+			if _, ok := rules[key]; !ok {
+				rules[key] = &rule
+				continue
+			}
+			rules[key].addVerbs(rule.Verbs)
 		}
 	}
 
-	if len(rules) == 0 {
+	var policyRules []rbacv1.PolicyRule
+	for _, rule := range rules {
+		policyRules = append(policyRules, rule.ToRule())
+
+	}
+
+	if len(policyRules) == 0 {
 		return nil
 	}
 
@@ -93,7 +158,7 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name: g.RoleName,
 		},
-		Rules: rules,
+		Rules: policyRules,
 	}); err != nil {
 		return err
 	}


pkg/rbac/parser.go
```diff
@@ -68,20 +126,36 @@ type Generator struct {
...
+	rules := make(map[string]*Rule)
 	for _, root := range ctx.Roots {
 		markerSet, err := markers.PackageMarkers(ctx.Collector, root)
 		if err != nil {
 			root.AddError(err)
 		}
 
-		for _, rule := range markerSet[RuleDefinition.Name] {
-			rules = append(rules, rule.(Rule).ToRule())
+		for _, marketValues := range markerSet[RuleDefinition.Name] {
```
DirectXMan12: market?

DirectXMan12: suddenly kubebuilder is fintech :-P

pkg/rbac/parser.go
```diff
@@ -44,8 +48,62 @@ type Rule struct {
 	URLs      []string `marker:"urls,optional"`
 }
 
+// target generates a string representing the resources and/or non-resources a Rule applies.
+func (r *Rule) target() string {
+	r.cleanUp()
+	s := []string{
```
DirectXMan12: I don't think this is sufficient -- I think we should spearately groups from resources from urls with a different character than we used to separate the individual items.

pkg/rbac/parser.go
```diff
@@ -44,8 +48,62 @@ type Rule struct {
 	URLs      []string `marker:"urls,optional"`
 }
 
+// target generates a string representing the resources and/or non-resources a Rule applies.
```
DirectXMan12: this godoc should be a bit more specific on what's used and why.

pkg/rbac/parser.go
```diff
@@ -44,8 +48,62 @@ type Rule struct {
...
+	return strings.Join(s, ",")
+}
+
+// addVerbs adds new verbs into a Rule
+func (r *Rule) addVerbs(verbs []string) {
+	r.Verbs = removeDup(append(r.Verbs, verbs...))
+}
+
+// cleanUp removes duplicates from each field of a Rule, and sorts each field.
+func (r *Rule) cleanUp() {
```
DirectXMan12: nit: cleanup is insufficiently descriptive.  Maybe `normalize` or something?

pkg/rbac/parser.go
```diff
@@ -68,20 +126,36 @@ type Generator struct {
 func (Generator) RegisterMarkers(into *markers.Registry) error {
 	return into.Register(RuleDefinition)
 }
+
 func (g Generator) Generate(ctx *genall.GenerationContext) error {
-	var rules []rbacv1.PolicyRule
+	// the map key is the target of a Rule (Rule.target())
+	rules := make(map[string]*Rule)
```
DirectXMan12: make the key a separate type -- it's clearer, and you can have `ruleKey { groups string; resources string; url string }`

pkg/rbac/parser.go
```diff
@@ -44,8 +48,62 @@ type Rule struct {
...
+	sort.Strings(r.Groups)
+	sort.Strings(r.Resources)
+	sort.Strings(r.Verbs)
+	sort.Strings(r.URLs)
+}
+
+// merge merges another Rule into the existing Rule
+// Only the Rules with the same Rule.target() can be merged.
+func (r *Rule) merge(other Rule) error {
+	if r.target() != other.target() {
```
DirectXMan12: why does this need an extra check?  Maybe just get rid of this method.

pkg/rbac/parser.go
haiyanmeng: Should be "marker". Fixed it!

pkg/rbac/parser.go
haiyanmeng: Removed this method.

pkg/rbac/parser.go
haiyanmeng: using ruleKey as the map key is a good suggestion. Fixed it!

pkg/rbac/parser.go
haiyanmeng: `normalize` is definitely better. That word didn't come to me yesterday!
Fixed it

pkg/rbac/parser.go
haiyanmeng: Removed this method, and use `ruleKey` as the map key.

pkg/rbac/parser.go
haiyanmeng: Removed this method and use ruleKey as the map key.

haiyanmeng:
@DirectXMan12 , thanks for the suggestions! PTAL!

pkg/rbac/parser.go
```diff
@@ -44,8 +47,54 @@ type Rule struct {
 	URLs      []string `marker:"urls,optional"`
 }
 
+// ruleKey represents the resources and non-resources a Rule applies.
+type ruleKey string
```
DirectXMan12: this can just be a struct:

```go
// ruleKey allows deduping rbac rules by groups/resources/urls combinations
// the keys are strings joined by `,`s to allow for use as a map key
type ruleKey struct {
  Groups string
  Resources string
  URLs string
}
```

pkg/rbac/parser.go
```diff
@@ -68,20 +117,33 @@ type Generator struct {
...
+	rules := make(map[ruleKey]*Rule)
 	for _, root := range ctx.Roots {
 		markerSet, err := markers.PackageMarkers(ctx.Collector, root)
 		if err != nil {
 			root.AddError(err)
 		}
 
-		for _, rule := range markerSet[RuleDefinition.Name] {
-			rules = append(rules, rule.(Rule).ToRule())
+		for _, markerValues := range markerSet[RuleDefinition.Name] {
```
DirectXMan12: nit: `markerValue`

pkg/rbac/parser.go
```diff
@@ -68,20 +117,33 @@ type Generator struct {
...
+			rule := markerValues.(Rule)
+			key := rule.key()
+			if _, ok := rules[key]; !ok {
```
DirectXMan12: this would be a lot clearer as `rules[rule.key()] = append(rules[rule.key()], rule.Verbs)`, and then dedup later.

pkg/rbac/parser.go
haiyanmeng: Done

pkg/rbac/parser.go
haiyanmeng: `Append in the first round + dedup in a second round` does have one issue: in the case when some marker values have lots of duplicates, the rule values may have lots of dups.

The current method is clear to me in that it only needs one round to get things done.

pkg/rbac/parser.go
haiyanmeng: I still don't see why using a struct here is better than a string.
We just need something to identify a rule.

DirectXMan12: it's a bit clearer as to how it's composed, and a bit harder to mess up + one fewer join operation.

pkg/rbac/parser.go
DirectXMan12: ¯\_(ツ)_/¯

I suppose it's fine either way, but to your point: this method creates a lot more intermediate objects.  In either case, the concession is relatively small, since this probably isn't a hot path.

pkg/rbac/parser.go
```diff
@@ -44,8 +47,54 @@ type Rule struct {
...
+	r.Resources = removeDupAndSort(r.Resources)
+	r.Verbs = removeDupAndSort(r.Verbs)
+	r.URLs = removeDupAndSort(r.URLs)
+}
+
+// removeDupAndSort removes duplicates in s, sorts the items, and returns a
+// new slice of strings.
+func removeDupAndSort(s []string) []string {
+	set := make(map[string]bool)
+	for _, str := range s {
```
DirectXMan12: we try to avoid single-letter variables for anything but method receivers, especially in cases where you've got a lot of similar types.

haiyanmeng: Fixed it! Thanks!

pkg/rbac/parser.go
haiyanmeng: Fixed it! Thanks!

pkg/rbac/parser.go
haiyanmeng: Thanks for sharing your insight on this! Then I would leave this as it is.

pkg/rbac/parser.go
```diff
@@ -44,8 +47,57 @@ type Rule struct {
...
+	Groups    string
+	Resources string
+	URLs      string
+}
+
+// key normalizes the Rule and returns a ruleKey object.
+func (r *Rule) key() ruleKey {
+	r.normalize()
+	return ruleKey{
+		strings.Join(r.Groups, "&"),
```
DirectXMan12: use struct keys

surprised `go vet` didn't complain.

haiyanmeng: Done

Issue #238, opened 2019-07-02
Race condition in (*TypeChecker).check()


jgustie:
I am using the 0.2.0-beta.2 release (darwin-amd64) and occasionally see a panic that seems to stem from a race condition in the `TypeChecker.check` method; I rebuilt `controller-gen` with `-race` and used that to `make manifests` in a Kubebuilder tutorial project (v2 scaffolding), it produces data race warnings generally like:

```
==================
WARNING: DATA RACE
Write at 0x00c000898810 by goroutine 50:
  runtime.mapassign_fast64()
      /usr/local/go/src/runtime/map_fast64.go:92 +0x0
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check()
      .../sigs.k8s.io/controller-tools/pkg/loader/refs.go:247 +0x21b
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check.func1()
      .../sigs.k8s.io/controller-tools/pkg/loader/refs.go:239 +0x6c

Previous write at 0x00c000898810 by goroutine 47:
  runtime.mapassign_fast64()
      /usr/local/go/src/runtime/map_fast64.go:92 +0x0
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check()
      .../sigs.k8s.io/controller-tools/pkg/loader/refs.go:247 +0x21b
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check.func1()
      .../sigs.k8s.io/controller-tools/pkg/loader/refs.go:239 +0x6c

Goroutine 50 (running) created at:
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check()
      .../sigs.k8s.io/controller-tools/pkg/loader/refs.go:237 +0x1b3
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check.func1()
      .../sigs.k8s.io/controller-tools/pkg/loader/refs.go:239 +0x6c

Goroutine 47 (finished) created at:
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check()
      .../sigs.k8s.io/controller-tools/pkg/loader/refs.go:237 +0x1b3
  .../sigs.k8s.io/controller-tools/pkg/loader.(*TypeChecker).check.func1()
[...]

PR #241, opened 2019-07-08
Add a `sync.Mutex` field into TypeChecker

haiyanmeng:
This avoids race condition which may happen when multiple goroutines try
to modify the `checkedPackages` field of TypeChecker simultaneously.

Fix: https://github.com/kubernetes-sigs/controller-tools/issues/238

**Performance concerns:**
I measured the time consumed by the `TypeChecker.Check` method on `kubebuilder/cronjob/api/v1` package with and without this change. In both cases, the time consumption is about 130 ms. One explanation for the similar performance is that the race condition does not happen at all during my test.

### DIFF
--- a/pkg/loader/refs.go
+++ b/pkg/loader/refs.go
@@ -187,6 +187,7 @@ func allReferencedPackages(pkg *Package, filterNodes NodeFilter) []*Package {
 type TypeChecker struct {
 	checkedPackages map[*Package]struct{}
 	filterNodes     NodeFilter
+	sync.Mutex
 }
 
 // Check type-checks the given package and all packages referenced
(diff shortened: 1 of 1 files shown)


pkg/loader/refs.go
```diff
@@ -244,5 +245,7 @@ func (c *TypeChecker) check(root *Package) {
 	// ...then, we can safely type-check ourself
 	root.NeedTypesInfo()
 
+	c.Lock()
+	defer c.Unlock()
```
mengqiy: Do we want to synchronize writes only or reads and writes?
There is a read in the same function.

DirectXMan12: yep, need to synchronize reads as well

haiyanmeng: Fixed it!

PR #247, opened 2019-07-10
 🐛Add simple +required marker

damemi:
If an API is set to default optional with `// +kubebuilder:validation:Optional`, the `// +required` marker isn't respected (because it doesn't exist, however `// +kubebuilder:validation:Required` does work). This adds the simple form of the required marker for parity with existing expectations


mengqiy:
Can you please add some tests?

DirectXMan12:
closing this for bookkeeping purposes, esp since we already have an equivalent marker

PR #248, opened 2019-07-11
:sparkles: Help!

DirectXMan12:
:notes: ... I need somebody! Help! Not just anybody! Help! You know I need someone, heeeeelp :notes:

This adds support for fetch more-or-less detailed help text, and adds help for all available markers.
From a user-facing perspective, this still looks like the `-w` and `-h` flags, except that both can now be passed more than once to get more details.

Passing either 4 times yields JSON output, which we can use to generate HTML documentation for the book.

There are also a number of utility changes along the way, as well as a slight change in default behavior -- previously, if no generators were passed, controller-gen attempted to invoke all of them, and then failed on rbac-gen's missing arguments.  Now it just errors out with a message about specifying generators.


DirectXMan12:
All apologies to the Beatles.  I may be a bit punchy.

DirectXMan12:
Example output (not showing color/formatting, which exists -- try it out yourself):

```bash
$ controller-gen -h

Usage:
  controller-gen [flags]

Examples:
        # Generate RBAC manifests and crds for all types under apis/,
        # outputting crds to /tmp/crds and everything else to stdout
        controller-gen rbac:roleName=<role name> crd paths=./apis/... output:crd:dir=/tmp/crds output:stdout

        # Generate deepcopy implementations for a particular file
        controller-gen deepcopy paths=./apis/v1beta1/some_types.go

        # Run all the generators for a given project
        controller-gen paths=./apis/...


Flags:
  -h, --detailed-help count   print out more detailed help (passing more times yields more detailed information about options)
      --help                  print out usage and a summary of options summary
  -w, --which-markers count   print out all markers available with the requested generators (passing more times yields more detailed information)


Options


generators

+webhook                                       package  generates (partial) {Mutating,Validating}WebhookConfiguration objects.
+rbac:roleName=<string>                        package  generates ClusterRole objects.
+object[:headerFile=<string>][:year=<string>]  package  generates code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+crd[:trivialVersions=<bool>]                  package  generates CustomResourceDefinition objects.


generic

+paths=<[]string>  package  represents paths and go-style path patterns to use as package roots.


output rules (optionally as output:<generator>:...)

+output:artifacts[:code=<string>]:config=<string>  package  outputs artifacts to different locations, depending on whether they're package-associated or not.
+output:dir=<string>                               package  outputs each artifact to the given directory, regardless of if it's package-associated or not.
+output:none                                       package  skips outputting anything.
+output:stdout                                     package  outputs everything to standard-out, with no separation.

```

```bash
$ controller-gen crd -www


CRD


+groupName package
	=<string>  specifies the API group name for this package.

+kubebuilder:printcolumn type
	adds a column to "kubectl get" output for this CRD.

	JSONPath=<string>
		specifies the jsonpath expression used to extract the value of the column.
	[description=<string>]
		specifies the help/description for this column.
	[format=<string>]
		specifies the format of the column. 
		 It may be any OpenAPI data format corresponding to the type, listed at https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types.

	name=<string>
		specifies the name of the column.
	[priority=<int>]
		indicates how important it is that this column be displayed. 
		 Lower priority (*higher* numbered) columns will be hidden if the terminal width is too small.

	type=<string>
		indicates the type of the column. 
		 It may be any OpenAPI data type listed at https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types.


+kubebuilder:resource type
	configures naming and scope for a CRD.

	[categories=<[]string>]
		specifies which group aliases this resource is part of. 
		 Group aliases are used to work with groups of resources at once. The most common one is "all" which covers about a third of the base resources in Kubernetes, and is generally used for "user-facing" resources.

	path=<string>
		specifies the plural "resource" for this CRD. 
		 It generally corresponds to a plural, lower-cased version of the Kind. See https://book.kubebuilder.io/cronjob-tutorial/gvks.html.

	[scope=<string>]
		overrides the scope of the CRD (cluster vs namespaced). 
		 Scope defaults to "namespaced".  Cluster-scoped ("cluster") resources don't exist in namespaces.

	[shortName=<[]string>]
		specifies aliases for this CRD. 
		 Short names are often used when people have work with your resource over and over again.  For instance, "rs" for "replicaset" or "crd" for customresourcedefinition.

	[singular=<string>]
		overrides the singular form of your resource. 
		 The singular form is otherwise defaulted off the plural (path).


+kubebuilder:storageversion type
	marks this version as the "storage version" for the CRD for conversion. 
	 When conversion is enabled for a CRD (i.e. it's not a trivial-versions/single-version CRD), one version is set as the "storage version" to be stored in etcd.  Attempting to store any other version will result in conversion to the storage version via a conversion webhook.

+kubebuilder:subresource:scale type
	enables the "/scale" subresource on a CRD.

	[selectorpath=<string>]
		specifies the jsonpath to the pod label selector field for the scale's status. 
		 The selector field must be the *string* form (serialized form) of a selector. Setting a pod label selector is necessary for your type to work with the HorizontalPodAutoscaler.

	specpath=<string>
		specifies the jsonpath to the replicas field for the scale's spec.
	statuspath=<string>
		specifies the jsonpath to the replicas field for the scale's status.

+kubebuilder:subresource:status type
	enables the "/status" subresource on a CRD.

+versionName package
	=<string>  overrides the API group version for this package (defaults to the package name).

CRD validation


+kubebuilder:validation:Enum type
	=<[]any>  specifies that this (scalar) field is restricted to the *exact* values specified here.

+kubebuilder:validation:ExclusiveMaximum type
	=<bool>  indicates that the maximum is "up to" but not including that value.

+kubebuilder:validation:ExclusiveMinimum type
	=<bool>  indicates that the minimum is "up to" but not including that value.

+kubebuilder:validation:Format type
	=<string>  specifies additional "complex" formatting for this field. 
		 For example, a date-time field would be marked as "type: string" and "format: date-time".

+kubebuilder:validation:MaxItems type
	=<int>  specifies the maximum length for this list.

+kubebuilder:validation:MaxLength type
	=<int>  specifies the maximum length for this string.

+kubebuilder:validation:Maximum type
	=<int>  specifies the maximum numeric value that this field can have.

+kubebuilder:validation:MinItems type
	=<int>  specifies the minimun length for this list.

+kubebuilder:validation:MinLength type
	=<int>  specifies the minimum length for this string.

+kubebuilder:validation:Minimum type
	=<int>  specifies the minimum numeric value that this field can have.

+kubebuilder:validation:MultipleOf type
	=<int>  specifies that this field must have a numeric value that's a multiple of this one.

+kubebuilder:validation:Optional field
	specifies that this field is optional, if fields are required by default.

+kubebuilder:validation:Optional package
	specifies that all fields in this package are optional by default.

+kubebuilder:validation:Pattern type
	=<string>  specifies that this string must match the given regular expression.

+kubebuilder:validation:Required field
	specifies that this field is required, if fields are optional by default.

+kubebuilder:validation:Required package
	specifies that all fields in this package are required by default.

+kubebuilder:validation:Type type
	=<string>  overrides the type for this field (which defaults to the equivalent of the Go type). 
		 This generally must be paired with custom serialization.  For example, the metav1.Time field would be marked as "type: string" and "format: date-time".

+kubebuilder:validation:UniqueItems type
	=<bool>  specifies that all items in this list must be unique.

+nullable field
	marks this field as allowing the "null" value. 
	 This is often not necessary, but may be helpful with custom serialization.

+optional field
	specifies that this field is optional, if fields are required by default.

```

DirectXMan12:
Part 1 of a two part saga -- next up, the KubeBuilder book

Issue #239, opened 2019-07-03
Make path optional


mengqiy:
From https://github.com/kubernetes-sigs/kubebuilder/pull/763#discussion_r296004286.

haiyanmeng:
I will take a look at this

PR #252, opened 2019-07-11
Make the `Path` field of the "+kubebuilder:resource" marker optional

haiyanmeng:
Fix: https://github.com/kubernetes-sigs/controller-tools/issues/239

This PR also adds `-race` flag into `go test`

This should unblock https://github.com/kubernetes-sigs/kubebuilder/pull/763

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -179,7 +179,7 @@ func (s PrintColumn) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, versio
 
 // Resource defines "+kubebuilder:resource"
 type Resource struct {
-	Path       string
+	Path       string   `marker:",optional"`
 	ShortName  []string `marker:",optional"`
 	Categories []string `marker:",optional"`
 	Singular   string   `marker:",optional"`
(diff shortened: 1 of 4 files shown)


pkg/crd/markers/crd.go
```diff
@@ -179,7 +179,7 @@ func (s PrintColumn) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, versio
 
 // Resource defines "+kubebuilder:resource"
 type Resource struct {
-	Path       string
+	Path       string   `marker:",optional"`
```
droot: I think we will also have to tweak the apply method for this marker to ignore path if empty. Right ?

haiyanmeng: You are right! Fixed it!

haiyanmeng:
> I wonder how easy it is to add a test to verify this change. If it is easy, then pl. consider adding the test.

I added a naked `kubebuilder:resource` marker into `testdata`, and regenerated `testdata.kubebuilder.io_cronjobs.yaml`. 
Thanks for pointing this out.

haiyanmeng:
FYI, I added another commit into the PR which adds `-race` flag into `go test`

Issue #251, opened 2019-07-11
Role ordering is not stable


jgustie:
Using 0.2.0-beta.3 the RBAC generator is modifying files on every pass. I think this is due to the #235 changes: the ultimate ordering of the policy rules slice is now determined by a map iteration.

I'm still transitioning from Java to Go, but I think the fix is to aggregate the `rules` keys into another slice, sort the ruleKey slice and iterate over that when building the list of policy rules.

DirectXMan12:
they should be being sorted @haiyanmeng FYI

PR #253, opened 2019-07-11
Avoid reordering Rules

haiyanmeng:
Currently, the `role.yaml` generated by multiple runs of controller-gen
may have different Rule orders. This commit makes the Rule order in
`role.yaml` stable.

Fix: https://github.com/kubernetes-sigs/controller-tools/issues/251


pkg/rbac/parser.go
```diff
@@ -121,45 +121,64 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
 }
 
-func (g Generator) Generate(ctx *genall.GenerationContext) error {
-	rules := make(map[ruleKey]*Rule)
+// GenerateClusterRole generates a rbacv1.ClusterRole object
+func GenerateClusterRole(ctx *genall.GenerationContext, roleName string) (*rbacv1.ClusterRole, error) {
+	// Using a slice here to preserve the order of Rules
+	var ruleSlice []Rule
+	// ruleKeyToIndexMap maps a ruleKey to its index in ruleSlice
+	ruleKeyToIndexMap := make(map[ruleKey]int)
```
mengqiy: You don't need a map now.

haiyanmeng: I kept the map, because the map is useful to track whether a Rule with a certain key has existed or not.

pkg/rbac/parser_integration_test.go
```diff
@@ -0,0 +1,57 @@
+package rbac_test
+
+import (
+	"io/ioutil"
+	"os"
+
+	"github.com/google/go-cmp/cmp"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/yaml"
```
mengqiy: nit: don't group k8s imports with non-k8s imports.

pkg/rbac/parser_integration_test.go
```diff
@@ -0,0 +1,57 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+
+	"sigs.k8s.io/controller-tools/pkg/genall"
+	"sigs.k8s.io/controller-tools/pkg/loader"
+	"sigs.k8s.io/controller-tools/pkg/markers"
+	"sigs.k8s.io/controller-tools/pkg/rbac"
+)
+
+var _ = Describe("ClusterRole generated by the RBAC Generator", func() {
+	It("should match the expected result", func() {
```
mengqiy: We may want to run the tests several times to ensure output is stable and no regression in the future.

haiyanmeng: Added a retry loop for 5 times. Thanks!

pkg/rbac/parser_integration_test.go
haiyanmeng: Fixed

pkg/rbac/parser.go
```diff
@@ -140,26 +154,46 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
-	var policyRules []rbacv1.PolicyRule
-	for _, rule := range rules {
-		policyRules = append(policyRules, rule.ToRule())
+	if len(rules) == 0 {
+		return nil, nil
+	}
 
+	// sort the Rules in rules according to their ruleKeys
+	keys := make([]ruleKey, 0)
+	for key, _ := range rules {
```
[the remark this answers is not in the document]
haiyanmeng: Fixed

Issue #216, opened 2019-05-30
Adjust for structural schema KEP


DirectXMan12:
A new kep was introduces to restrict how we generate OpenAPI validation for CRDs.  It has a number of special extensions that we'll want to support, plus restrictions on validating objectmeta. We'll need some specially handling here, since people might want a full metadata openapi spec for other reasons.

See also https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190425-structural-openapi.md

DirectXMan12:
cc @sttts

sttts:
> since people might want a full metadata openapi spec for other reasons.

Which are those reasons? 

Note that the spec is most probably incomplete when defined by the developer because ObjectMeta is defined by the cluster at hand. In contrast to anything else in a CRD, the ObjectMeta life cycle is not the one the operator builder controls. When the cluster adds a meta feature (server side apply, more powerful GC, etc.), every resource will inherit it.

DirectXMan12:
We have a contributor who initially wanted to use this code for generating openapi specs to be consumed by editor tooling.

jimmidyson:
Any thoughts on how to tackle this @DirectXMan12?

pires:
@vincepri @ncdc can you describe better what was the problem embedded CRD ObjectMeta was raising? Seems to be related to this.

sttts:
@pires With structural schema we have `x-kubernetes-embedded-resource: true` which does everything around `ObjectMeta` validation.

jimmidyson:
I've worked around this with a kustomize patch of the crd manifest to only include name and generateName in ObjectMeta properties, but still with ObjectMeta in original struct (nothing gets generated if this isn't present). We could just do that on generation of validation spec, just restrict to name and generateName properties in metadata.

sttts:
@jimmidyson you will be able to have `x-kubernetes-embedded-resource: true` and in parallel your custom restrictions on name and generateName.

pires:
@jimmidyson @sttts IIUC the approach Cluster API folks picked up was to [remove ObjectMeta and explicitly specify the attributes Jimmi mentioned](https://github.com/kubernetes-sigs/cluster-api/pull/1062).  As I was reading the updates for CRD in 1.15, including `x-kubernetes-embedded-resource: true`, I figured it could be a _better_ solution. Hopefully, I'm not wrong 🤞

sttts:
[...]

PR #254, opened 2019-07-12
:bug: special-casing metav1.ObjectMeta.

mengqiy:
We now only generate the validation schema for name and generateName
fields.

This PR also alligns with https://github.com/kubernetes-sigs/controller-tools/issues/216


DirectXMan12:
Other parts we'll need to cover: `x-kubernetes-object` and overriding this validation for specific name/generateName constraints.

/lgtm
/approve

DirectXMan12:
This is actually slightly incorrect -- it's fine for top-level, but embedded object meta is now generated incorrectly -- you actually want the "full" object meta there, or using `x-kubernetes-embedded-resource`.

DirectXMan12:
we should revert this before release

Issue #225, opened 2019-06-17
Support generating Role


mengqiy:
Moved to CT from https://github.com/kubernetes-sigs/kubebuilder/issues/401.

kubebuilder currently only support generating ClusterRole, but there are use cases that we may only want a Role which restrict the permission to certain namespaces.

haiyanmeng:
I can work on this.

PR #260, opened 2019-07-17
Add a new `namespace` field into the RBAC marker

haiyanmeng:
This makes is possible to generate both ClusterRole and Role(s) from the RBAC marker

Fix: https://github.com/kubernetes-sigs/controller-tools/issues/225


pkg/rbac/parser.go
```diff
@@ -146,69 +147,110 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
-			rules[key].addVerbs(rule.Verbs)
+			ruleMap[key].addVerbs(rule.Verbs)
 		}
-	}
 
-	if len(rules) == 0 {
-		return nil, nil
-	}
+		// sort the Rules in rules according to their ruleKeys
+		keys := make([]ruleKey, 0)
```
mengqiy: Change it to `make([]ruleKey, len(ruleMap))` to avoid resizing when appending.

pkg/rbac/parser.go
```diff
@@ -146,69 +147,110 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
 
-	var policyRules []rbacv1.PolicyRule
-	for _, key := range keys {
-		policyRules = append(policyRules, rules[key].ToRule())
+		}
+		return policyRules
+	}
 
+	// collect all the namespaces and sort them
+	var namespaces sort.StringSlice
```
mengqiy: It seems it'll be simpler to just use `sort.Strings()`
https://golang.org/pkg/sort/#Strings

pkg/rbac/parser.go
```diff
@@ -146,69 +147,110 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
+		namespaces = append(namespaces, ns)
+	}
+	namespaces.Sort()
+
+	// process the items in rulesByNS by the order specified in `namespaces` to make sure that the Role order is stable
+	var objs []interface{}
+	for _, ns := range namespaces {
+		rules := rulesByNS[ns]
+		var policyRules []rbacv1.PolicyRule
+		policyRules = NormalizeRules(rules)
```
mengqiy: Why not just using `policyRules := NormalizeRules(rules)`?

pkg/rbac/parser_integration_test.go
```diff
@@ -43,19 +45,29 @@ var _ = Describe("ClusterRole generated by the RBAC Generator", func() {
...
 			Expect(err).NotTo(HaveOccurred())
 
 			By("loading the desired YAML")
 			expectedFile, err := ioutil.ReadFile("role.yaml")
 			Expect(err).NotTo(HaveOccurred())
 
 			By("parsing the desired YAML")
-			var role rbacv1.ClusterRole
-			Expect(yaml.Unmarshal(expectedFile, &role)).To(Succeed())
+			for i, expectedRoleBytes := range bytes.Split(expectedFile, []byte("\n---\n"))[1:] {
```
mengqiy: It's a little too complex for a simple unit test.
Since we have sorted everything. We can just compare the 2 objects in serialized format, can't we? We return the 1st place that mismatches.

mengqiy:
Currently we use `+rbac:roleName=<string>` to specify the `ClusterRole` name.
I guess it's OK we use this to set the names for both `Role` and `ClusterRole`.
We may want to have 2 different markers in the future if the users request.

pkg/rbac/parser.go
```diff
@@ -146,69 +147,110 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
 	return nil
 }
 
-// GenerateClusterRole generates a rbacv1.ClusterRole object
-func GenerateClusterRole(ctx *genall.GenerationContext, roleName string) (*rbacv1.ClusterRole, error) {
-	rules := make(map[ruleKey]*Rule)
+// GenerateRoles generate a slice of objs representing either a ClusterRole or a Role object
+// The order of the objs in the returned slice is stable and determined by their namespaces.
+func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{}, error) {
```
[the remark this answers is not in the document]
haiyanmeng: I would keep this as it. Because `WriteYAML expects `...interface{}`:
```
func (g GenerationContext) WriteYAML(itemPath string, objs ...interface{}) error
```

pkg/rbac/parser.go
```diff
@@ -51,7 +51,8 @@ type Rule struct {
 	// Verbs specifies the (lowercase) kubernetes API verbs that this rule encompasses.
 	Verbs []string
 	// URL specifies the non-resource URLs that this rule encompasses.
-	URLs []string `marker:"urls,optional"`
+	URLs      []string `marker:"urls,optional"`
+	Namespace string   `marker:",optional"`
```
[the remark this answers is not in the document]
haiyanmeng: Done

pkg/rbac/parser.go
haiyanmeng: Done

pkg/rbac/parser.go
haiyanmeng: Done

pkg/rbac/parser.go
haiyanmeng: Done

pkg/rbac/parser_integration_test.go
haiyanmeng: `rbac.GenerateRoles` generate a list of objs, to serialize these objects into a single file, we need to combine all the objects and  add line break `\n---\n` between objects. This does not seem to much simpler than the existing code in this PR. Another advantage of the existing code is that we can give users more specific diff.

haiyanmeng:
> Currently we use `+rbac:roleName=<string>` to specify the `ClusterRole` name.
> I guess it's OK we use this to set the names for both `Role` and `ClusterRole`.
> We may want to have 2 different markers in the future if the users request.

+1

pkg/rbac/testdata/out
```diff
@@ -0,0 +1,83 @@
+Error: /usr/bin/gopackagesdriver: exit status 2: runtime: running GoogleExitFunction
```
mengqiy: Do we need this?

pkg/rbac/testdata/1
```diff
@@ -0,0 +1,51 @@
+Error: unknown option "2"
```
mengqiy: Do we need this?

pkg/rbac/testdata/out
haiyanmeng: Oops. Checked this in by accident. 
Removed it.

pkg/rbac/testdata/1
haiyanmeng: No. Removed it

mengqiy:
I just recall that the generated marker help is not updated in this PR: https://github.com/kubernetes-sigs/controller-tools/blob/46695abfb89bb6fcd972edf4f925915434726008/pkg/rbac/zz_generated.markerhelp.go#L50-L67

IIRC you can run `go run ./cmd/helpgen` to generate it. I might be wrong. It is recently added.

haiyanmeng:
> added

haiyanmeng:
Updated the generated marker help.

PR #263, opened 2019-07-17
:sparkles: metadata field changes for structural schema

droot:
This PR handles generating schema for metadata as per structural schema.
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190425-structural-openapi.md

Partial fix for https://github.com/kubernetes-sigs/controller-tools/issues/216


pkg/crd/flatten.go
```diff
@@ -439,3 +446,49 @@ func (f *flattenVisitor) Visit(baseSchema *apiext.JSONSchemaProps) SchemaVisitor
...
+	metav1Pkg := "k8s.io/apimachinery/pkg/apis/meta/v1"
+	switch {
+	// root level ObjectMeta should not be specified in the validation schema
+	// because it is managed by kube-apiserver. Refer to the KEP specified
+	// above.
+	case flattenCtx.level == 1 /* root level metadata */ &&
+		ident.Package.ID == metav1Pkg && ident.Name == "ObjectMeta":
+		return &apiext.JSONSchemaProps{
+			Type: "object",
+			Properties: map[string]apiext.JSONSchemaProps{
```
sttts: these two are possible, but not needed in a structural schema.

pkg/crd/flatten.go
```diff
@@ -439,3 +446,49 @@ func (f *flattenVisitor) Visit(baseSchema *apiext.JSONSchemaProps) SchemaVisitor
...
+					Type: "string",
+				},
+			},
+		}
+	case flattenCtx.level > 1 /* ObjectMeta being used at a non-root level */ &&
+		ident.Package.ID == metav1Pkg && ident.Name == "ObjectMeta":
+		return nil
+		// TODO(droot): enable it once we move k8s 1.15
+		// return &apiext.JSONSchemaProps{
+		// 	// XEmbeddedResource: true,
```
sttts: for this @DirectXMan12 brought up the point of partial ObjectMeta. There is no requirement that the metadata is complete for an embedded resource. /cc @liggitt 

We either want a way to mark an embedded ObjectMeta as partial, or we relax the validation we do today (if that is possible now in beta). @liggitt wdyt?

liggitt: what is currently required that we would want to relax?

liggitt: as far as I remember, only kind/apiVersion were required

sttts: It's about ObjectMeta: name, namespace as far as I see.

liggitt: name must be optional in openapi because of generateName, and iirc our server-side implementation put a dummy name in place when validating if no name was provided

namespace must already be optional (both server-side and in openapi) because of cluster-scoped resources

sttts: ok, I think I misunderstood @DirectXMan12 in Slack. He was talking about `TypeMeta` (and I remember the dummy name value we put it to work around the check).

So the question is whether without TypeMeta it is really an embedded object.

droot:
@sttts @DirectXMan12 @mengqiy I reevaluated my approach and thought of another to approach the solution:

By default, do not generate any validation for ObjectMeta (irrespective of it being a root or embedded). Let users specify marker ` // +kubebuilder:validation:embedded ` if we want enable validation for embedded objects. This is how users do it for enabling any validation for the fields in the spec today, so it's inline with that. This will also avoid the partial objectMeta issue that @stts commented above where in current flattenVisitor it doesn't have a good way to track if objectMeta has sibling node TypeMeta or not.  Thoughts ?

DirectXMan12:
sgtm on the marker

DirectXMan12:
that doesn't actually resolve the problem though -- the `x-kubernetes-embedded`, IIRC requires `apiVersion` and `kind` to be specified too, so won't work for things like the `template` field in `Deployment`/`CronJob`/etc.  

@sttts @liggitt 

but we can tackle that later

droot:
closing this in favor of https://github.com/kubernetes-sigs/controller-tools/pull/266

PR #266, opened 2019-07-18
:warning: drop validation schema for objectmeta

droot:
As per structural schema KEP
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190425-structural-openapi.md,
dropping support objectmeta schema generation.


droot:
About adding marker for `x-kubernetes-embedded: true`, we need to update controller-tools to 1.15, so that we can refer to new CRD jsonschema definition. So that will be added in a future PR.

Issue #236, opened 2019-06-28
Versions list in the generated CRD should be sorted by version name


droot:
Looks like versions list generated CRD is not stable. So if the ordering doesn't match the one persisted in the etcd, server will throw an error.

You may see an error similar to the following:

```
The CustomResourceDefinition <YOUR CRD> is invalid: spec.version: Invalid value: "v1": must match the first version in spec.versions
```

mengqiy:
Both `spec.version` and `spec.versions` are optional.
IMO if we have more than one versions, we should just alway clear `spec.version` field.

mengqiy:
Ref: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#version-priority

DirectXMan12:
The actual issue here, I believe, is not the sorting, but the defaulting of `spec.version` using the sorted version.

Having a stable sort is nice, but what we actually need to do here is that we're consistently setting `version`.

haiyanmeng:
> The actual issue here, I believe, is not the sorting, but the defaulting of `spec.version` using the sorted version.
> 
> Having a stable sort is nice, but what we actually need to do here is that we're consistently setting `version`.

Thanks for pointing out, Solly.

In summary, what we really need are two things:
1) make the order of CRD versions stable;
2) explicitly set `crd.Spec.Version` to the first item of the sorted CRD version list.

PR #267, opened 2019-07-18
Sort the CRD Versions in the generated CRD and set `crd.Spec.Version` explicitly

haiyanmeng:
According to the kubernetes CRD validation, we need to make sure that the `Version` field always matches the first version in the versions list.  If that changes, we have to explicitly set the version field, since the defaulting will have pointed it at a different version on initial creation.

Fixes: https://github.com/kubernetes-sigs/controller-tools/issues/236

### DIFF
--- a/pkg/crd/spec.go
+++ b/pkg/crd/spec.go
@@ -17,6 +17,7 @@ package crd
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/gobuffalo/flect"
@@ -118,6 +119,11 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind) {
 	if len(crd.Spec.Versions) == 0 {
 		return
 	}
+
+	// it is necessary to make sure the order of CRD versions in crd.Spec.Versions is stable and explicitly set crd.Spec.Version.
+	// Otherwise, crd.Spec.Version may point to different CRD versions across different runs.
+	sort.Slice(crd.Spec.Versions, func(i, j int) bool { return crd.Spec.Versions[i].Name < crd.Spec.Versions[j].Name })
+	crd.Spec.Version = crd.Spec.Versions[0].Name
 
 	// make sure we have *a* storage version
 	// (default it if we only have one, otherwise, bail)
--- a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
+++ b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
@@ -11,6 +11,7 @@ spec:
     kind: CronJob
     plural: cronjobs
   scope: Namespaced
+  version: v1
   versions:
   - name: v1
     schema:


pkg/crd/spec.go
```diff
@@ -36,6 +39,102 @@ type SpecMarker interface {
 	ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, version string) error
 }
 
+type VersionType string
```
DirectXMan12: needs godoc

pkg/crd/spec.go
```diff
@@ -36,6 +39,102 @@ type SpecMarker interface {
...
+
+type VersionName string
+
+// Value creates a float64 type of value to represent a VersionName
+func (v VersionName) Value() float64 {
+	// remove the first character, `v`
+	name := string(v)[1:]
+	name = strings.Replace(name, "beta", ".", -1)
+	name = strings.Replace(name, "alpha", ".", -1)
+	result, _ := strconv.ParseFloat(name, 64)
```
DirectXMan12: eeek.  versions are not floats, and treating them as such is a bad practice.  For example: beta1 is not the same as beta10, but in float terms they would be.  You'll need to find a better way to do this.

pkg/crd/spec.go
```diff
@@ -36,6 +39,102 @@ type SpecMarker interface {
...
+	return result
+}
+
+// VersionNames implements sort.Interface for []VersionName based on VersionName.Value
+// all the items in VersionNames should have the same VersionType
+type VersionNames []VersionName
+
+func (v VersionNames) Len() int           { return len(v) }
```
DirectXMan12: you can use `sort.Slice` for this

pkg/crd/spec.go
```diff
@@ -36,6 +39,102 @@ type SpecMarker interface {
...
+// Strings creates a slice of string where each string corresponds to one VersionName
+func (v VersionNames) Strings() []string {
+	result := []string{}
+	for _, item := range v {
+		result = append(result, string(item))
+	}
+	return result
+}
+
+func sortVersionNames(versionNames []string) []string {
```
DirectXMan12: this needs godoc

pkg/crd/spec.go
```diff
@@ -36,6 +39,102 @@ type SpecMarker interface {
...
+	patterns := map[VersionType]string{
```
DirectXMan12: this can be generalized into `^v([0-9]+)((alpha|beta)([0-9]*))?$`, at which point you can numerically sort groups 1 and 4, and string (lexigraphically) sort group 3, IIRC.  The regex looks a bit more complicated, but the sorting logic is less complicated

pkg/crd/spec.go
```diff
@@ -72,23 +171,26 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind) {
...
+			Name:   verName,
 			Served: true,
 			Schema: &apiext.CustomResourceValidation{
 				OpenAPIV3Schema: fullSchema,
 			},
 		}
-		crd.Spec.Versions = append(crd.Spec.Versions, ver)
 	}
 
+	crd.Spec.Versions = SortCRDVersions(versionMap)
```
DirectXMan12: you could also simplify this a bit by calling `sort.Slice` and putting parsed versions in a lookaside map, potentially.

haiyanmeng: Using `sort.Strings` instead

pkg/crd/spec.go
haiyanmeng: removed

pkg/crd/spec.go
haiyanmeng: removed

pkg/crd/spec.go
haiyanmeng: removed

pkg/crd/spec.go
haiyanmeng: removed

pkg/crd/spec.go
haiyanmeng: removed

haiyanmeng:
@DirectXMan12 , fixed it! PTAL!

pkg/crd/spec.go
```diff
@@ -36,6 +37,24 @@ type SpecMarker interface {
 	ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, version string) error
 }
 
+// SortCRDVersions sort values in VersionMap by version names using the `sort.Strings` function.
```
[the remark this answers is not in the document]
haiyanmeng: Removed this func. Use `sort.Slice` instead.

pkg/crd/spec.go
```diff
@@ -72,23 +91,26 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind) {
...
 		typeIdent := TypeIdent{Package: pkg, Name: groupKind.Kind}
 		typeInfo := p.Types[typeIdent]
 		if typeInfo == nil {
 			continue
 		}
 		fullSchema := FlattenEmbedded(p.flattener.FlattenType(typeIdent), pkg)
-		ver := apiext.CustomResourceDefinitionVersion{
-			Name:   p.GroupVersions[pkg].Version,
+		verName := p.GroupVersions[pkg].Version
+		versionMap[verName] = apiext.CustomResourceDefinitionVersion{
```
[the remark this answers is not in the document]
haiyanmeng: Done! Thanks for pointing this out! Very neat!

pkg/crd/spec.go
```diff
@@ -122,6 +144,7 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind) {
 	// make sure we have *a* storage version
 	// (default it if we only have one, otherwise, bail)
 	if len(crd.Spec.Versions) == 1 {
+		crd.Spec.Version = crd.Spec.Versions[0].Name
```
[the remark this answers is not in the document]
haiyanmeng: Fixed it

DirectXMan12:
/hold

needs a commit message -- you can just use the PR message I added, if you want.

otherwise LGTM

haiyanmeng:
> /hold
> 
> needs a commit message -- you can just use the PR message I added, if you want.
> 
> otherwise LGTM

Done. I just used the PR msg you added.

PR #268, opened 2019-07-19
🐛 Handle the error type in crd gen

coderanger:
So this fixes a very niche parsing issue:

```go
type noNotify struct{ error }
```

That parses as a field, but it has no package because it's a core type I think? Not 100% sure but it definitely causes a segfault :)

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -148,6 +148,11 @@ func findKubeKinds(parser *Parser, metav1Pkg *loader.Package) []schema.GroupKind
 				// ObjectMeta and TypeMeta are named types
 				continue
 			}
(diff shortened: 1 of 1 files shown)


munnerz:
Huh, how come you're embedding an `error` type into your API? What're you trying to achieve with this?

Avoiding segfaults sounds like a good plan regardless, but I'm curious as to why you're doing this 😅

/lgtm

coderanger:
@munnerz It's a extension to pkg/errors to basically tag an error as "please don't generate a notification for this" so when it propagates up to my error handlers, they can check and not push a Slack message.

DirectXMan12:
ah, the universe types.  They're weird an need explicit handling, but the rest are builtins.  `error` is weird.

Issue #270, opened 2019-07-19
metav1.MicroTime set as CRD type: object instead of format: date-time, type: string


jwoglom:
metav1.Time is properly parsed as a date-time, while metav1.MicroTime is not.

e.g. a field definition ``` LastUpdateTime metav1.MicroTime `json:"lastUpdateTime"` ``` results in a CRD generation of:
```
properties:
  lastResetTime:
    format: date-time
    type: string
```
Changing the field to `metav1.MicroTime` results in:
```
properties:
  lastResetTime:
    type: object
```
resulting in kubebuilder errors such as:
```
/path/to/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/apis/meta/v1/micro_time.go:34:2: enountered struct field "" without JSON tag in type "MicroTime"
```

PR #271, opened 2019-07-19
✨ Correctly format metav1.MicroTime as date-time

jwoglom:
This brings the handling of k8s.io/apimachinery/pkg/apis/meta/v1/MicroTime in sync with k8s.io/apimachinery/pkg/apis/meta/v1/Time.

This resolves issue #270.

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -35,6 +35,10 @@ var KnownPackages = map[string]PackageOverride{
 			Type:   "string",
 			Format: "date-time",
 		}
+		p.Schemata[TypeIdent{Name: "MicroTime", Package: pkg}] = apiext.JSONSchemaProps{
+			Type:   "string",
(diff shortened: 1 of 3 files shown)


pkg/crd/known_types.go
```diff
@@ -35,6 +35,10 @@ var KnownPackages = map[string]PackageOverride{
 			Type:   "string",
 			Format: "date-time",
 		}
+		p.Schemata[TypeIdent{Name: "MicroTime", Package: pkg}] = apiext.JSONSchemaProps{
+			Type:   "string",
+			Format: "date-time",
+		}
```
droot: pl. add test in the test project in `pkg/crd`

jwoglom: Done.

droot:
Looks good. Can you pl. squash the commits, then it's good to go. Thanks.

PR #275, opened 2019-07-23
:bug: A myriad of object-gen fixes, plus tests

### LINKED ISSUE #262: Wrong deepcopy generation when using type aliased arrays or maps
I'm hitting an issue when generating `deepcopy` files with `controller-gen`. If I have something like:
```golang
// IngressRules is a slice of AWS ingress rules for security groups.
type IngressRules []*IngressRule
```
It generates:
```golang
func (in IngressRules) DeepCopyInto(out *IngressRules) {
    in := &in
    *out = make(IngressRules, len(*in))
    for i := range *in {
        if (*in)[i] != nil {
            in, out := &(*in)[i], &(*out)[i]
            *out = new(IngressRule)
            (*in).DeepCopyInto(*out)
        }
    }
}
```
Which fails compilation with `cannot use &in (type *IngressRules) as type IngressRules in assignment`.

The same issue goes for `type Tags map[string]string`, where we alias Tags to a map.

Tried both latest `master` and `beta.3`

DirectXMan12:
This contains a myriad of fixes for corner cases in object-gen, plus a fairly comprehensive integration test based partially on the cronjob tutorial and partially on the legacy deepcopy-gen integration tests.

Fixes #262


DirectXMan12:
side note: I somehow have managed to come out of this disliking gengo even more than before :-/

DirectXMan12:
Just to be triple, sure, I've run this against `k8s.io/api` as well.

vincepri:
LGTM

PR #276, opened 2019-07-23
:warning: Update Kubernetes client libraries to 1.15.0 release

munnerz:
This is #231 by @tamalsaha rebased onto master.

I'm not 100% sure how dependency files are meant to be managed in this repo as there's no `hack/` directory, and the repo also includes both Gopkg.* files as well as go.{mod,sum}.

/cc @DirectXMan12

PR #277, opened 2019-07-23
🐛 Add apiextensions JSON type to known_types.go

munnerz:
Adds support for the apiextensions-apiserver JSON type.

Without this patch I get the following error:

```
/Users/james/go/src/github.com/jetstack/cert-manager/vendor/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1/types_jsonschema.go:98:2: enountered struct field "Raw" without JSON tag in type "JSON"
```

/cc @DirectXMan12

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -84,6 +84,17 @@ var KnownPackages = map[string]PackageOverride{
 		}
 		// No point in calling AddPackage, this is the sole inhabitant
 	},
+
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1": func(p *Parser, pkg *loader.Package) {
+		p.Schemata[TypeIdent{Name: "JSON", Package: pkg}] = apiext.JSONSchemaProps{
+			XPreserveUnknownFields: boolPtr(true),
+		}
+		p.AddPackage(pkg) // get the rest of the types
+	},
+}
+
+func boolPtr(b bool) *bool {
+	return &b
 }
 
 // AddKnownTypes registers the packages overrides in KnownPackages with the given parser.


pkg/crd/known_types.go
```diff
@@ -73,6 +73,13 @@ var KnownPackages = map[string]PackageOverride{
 		}
 		// No point in calling AddPackage, this is the sole inhabitant
 	},
+
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1": func(p *Parser, pkg *loader.Package) {
+		p.Schemata[TypeIdent{Name: "JSON", Package: pkg}] = apiext.JSONSchemaProps{
+			Type: "string",
```
munnerz: I'm not 100% sure if the type here should be `string` or something else - the existing Golang code doesn't actually specify.

cc @sttts any ideas?

sttts: type string is wrong. This is arbitrary JSON, like RawExtension but without Type/ObjectMeta.

munnerz: After some testing, I've updated this to `object`. I think this is correct, although it *may* not accept data which is a string whereas it probably should given JSON is meant to be completely unstructured.

What's the proper way to set this? It looks like the Kubernetes APIServer exposes an `object` with a field named `Raw` of type `string` in the OpenAPI schema for RawExtension which I think would *also* be incorrect?

sttts: Also object is wrong. This is pure JSON, any type. In other words, no type should be set at all.

But you have to set `x-kubernetes-preserve-unknown-fields: true`.

sttts:
```
apiext.JSONSchemaProps{
  XPreserveUnknownFields: true,
}
```

munnerz: 👍 thanks, so it seems like this requires #276 then!

munnerz:
I've updated our repo (https://github.com/jetstack/cert-manager/pull/1946) to use this, however now run into an issue when installing these CRDs on anything *older* than Kubernetes 1.15, presumably because the new field on the CRD does not exist before 1.15.

With preserveUnknownFields eventually flipping to `true` by default in future, this is going to be a big PITA for us as we'll have to ship manifests specifically for k8s <= 1.14. Just a heads up really, and I assume it's accounted for already in the 'flip the default' plan. I'd hope that 1.14 would be EOL before this default flips 😄

For controller-tools specifically, I'm not sure how best to proceed. It'd be great if this sort of thing can be handled, but in reality it can't be handled properly *yet*, so maybe this patch should also be skipped until 1.14 is EOL 🙄

pkg/crd/known_types.go
sttts: yes

sttts:
> however now run into an issue when installing these CRDs on anything older than Kubernetes 1.15, presumably because the new field on the CRD does not exist before 1.15.

What is the issue? Unsupported fields are ignored by the API server.

DirectXMan12:
unsupported fields are ignored by the server, but hated by kubectl due to client-side schema validation, which is on by default.

Everyone always seems to forget that, and then it causes headaches down the line.

sttts:
kubebuilder needs a mode for pruned CRDs. In that case, the `x-kubernetes-preserve-unknown-fields: true` field should be set.

DirectXMan12:
We have 1.15 types now, so this can be updated

/hold

until then

sttts:
lgtm

munnerz:
Looks like this is ready for merge now, just needs lgtm and approved labels! 😄 

/hold cancel

pkg/crd/known_types.go
```diff
@@ -84,6 +85,13 @@ var KnownPackages = map[string]PackageOverride{
 		}
 		// No point in calling AddPackage, this is the sole inhabitant
 	},
+
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1": func(p *Parser, pkg *loader.Package) {
+		p.Schemata[TypeIdent{Name: "JSON", Package: pkg}] = apiext.JSONSchemaProps{
+			XPreserveUnknownFields: pointer.BoolPtr(true),
```
DirectXMan12: can you just inline this?  I see very little reason to pull in a library for this

munnerz: done

munnerz:
Updated to inline the function, mind taking another look @DirectXMan12?

PR #282, opened 2019-07-25
🐛 Set singular name in generated CRD yaml

tamalsaha:

Singular name passed by annotation was not used or set in the generated CRD yamls.


pkg/crd/markers/crd.go
```diff
@@ -268,6 +268,9 @@ func (s Resource) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, version s
 	if s.Path != "" {
 		crd.Names.Plural = s.Path
 	}
+	if s.Singular != "" {
+		crd.Names.Singular = s.Singular
+	}
```
droot: Can we pl. add a test for this.  pkg/crd has a project in testdata that we are using for tests. 

otherwise, LGTM to me.

Issue #257, opened 2019-07-16
CRD yaml file generated is too verbose


skonto:
Is there an option to create a shorter generated CRD yaml? Is there a way to declare defaults?
I tried to create the equivalent of this : https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/manifest/spark-operator-crds.yaml and it was ~5361 lines. Are there any flags to set to make it have a reasonable size? Also I noticed inconsistencies:
eg. 
```
            pythonVersion:
              description: This sets the major Python version of the docker image
                used to run the driver and executor containers. Can either be 2 or
                3, default 2. Optional.
              type: string
```
while in the original it is:
```
            pythonVersion:
              enum:
              - "2"
              - "3"
```

PR #283, opened 2019-07-25
🏃 Remove status from generated CRD yaml

tamalsaha:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,7 @@
 
 
 [[projects]]
-  digest = "0:"
+  digest = "1:865079840386857c809b72ce300be7580cb50d3d3129ce11bf9aa6ca2bc1934a"
   name = "github.com/fatih/color"
   packages = ["."]
   pruneopts = "UT"
@@ -10,7 +10,7 @@
   version = "v1.7.0"
 
 [[projects]]
-  digest = "0:"
+  digest = "1:0e638432a31fe74a82a8efb3f73015041548990c47a78f6a37397755ae41ab39"
   name = "github.com/gobuffalo/flect"
   packages = ["."]
   pruneopts = "UT"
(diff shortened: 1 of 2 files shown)


tamalsaha:
Currently generated CRD yamls have an empty status section. This pr will remove that.

crd-status

tamalsaha:
This same pipeline can be used to remove the description fields to solve https://github.com/kubernetes-sigs/controller-tools/issues/288 and https://github.com/kubernetes-sigs/controller-tools/issues/257

Once we have `map[string]interface{}`, the description fields can be yanked.

munnerz:
+1 to what you're doing here, but I wonder why the regular `omitempty` tags aren't already taking care of this.. 🙄 

Aside from my own curiosity though, this lgtm 😄

/lgtm

DirectXMan12:
Those fields aren't properly marked as optional/omitempty  We actually *need* have them present for some kubernetes versions, otherwise client-side validation barfs.

tamalsaha:
> We actually need have them present for some kubernetes versions, otherwise client-side validation barfs.

@DirectXMan12 , can you please tell me which version causes problem? I did not know that.

DirectXMan12:
I don't recall off the top of my head, but I've definitely seen kubectls that exhibit the issue.  If you can't find one, I'll double check tomorrow.

DirectXMan12:
hmm... it's also possible I'm misremembering, and it's that status, when present, gets generated incorrectly by default.  I'll play around a bit.

liubog2008:
It's useful for server-side apply. Now status will be conflicted.

See http://k8s.io/docs/reference/using-api/api-concepts/#conflicts
Apply failed with 3 conflicts: conflicts with "kube-apiserver" using apiextensions.k8s.io/v1:
- .status.conditions
- .status.acceptedNames.kind
- .status.acceptedNames.plural

liubog2008:
@DirectXMan12 @tamalsaha

PR #290, opened 2019-07-31
:bug: Collapse Common CRD Version Info

DirectXMan12:
The Kube API server requires that information common to all versions of
a CRD be collapsed into the top-level fields in v1beta1, since not doing
so would create ambiguity when round-tripping through the upcoming v1
version (which doesn't have top-level fields).  This makes sure we
collapse things.

### DIFF
--- a/pkg/crd/spec.go
+++ b/pkg/crd/spec.go
@@ -22,6 +22,7 @@ import (
 	"github.com/gobuffalo/flect"
 
 	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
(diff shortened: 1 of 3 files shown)


pkg/crd/spec.go
```diff
@@ -36,6 +37,86 @@ type SpecMarker interface {
...
+func mergeIdenticalSchemata(crd *apiext.CustomResourceDefinition) {
+	schema := crd.Spec.Versions[0].Schema
+	for _, ver := range crd.Spec.Versions {
+		if ver.Schema == nil || !equality.Semantic.DeepEqual(schema, ver.Schema) {
+			// either all nil, or not identical
+			return
+		}
+	}
+
+	// things are identical if we've gotten this far, so move the subresources up
```
mengqiy: s/subresources/schema

pkg/crd/spec.go
```diff
@@ -36,6 +37,86 @@ type SpecMarker interface {
...
+func mergeIdenticalPrinterColumns(crd *apiext.CustomResourceDefinition) {
+	cols := crd.Spec.Versions[0].AdditionalPrinterColumns
+	for _, ver := range crd.Spec.Versions {
+		if len(ver.AdditionalPrinterColumns) == 0 || !equality.Semantic.DeepEqual(cols, ver.AdditionalPrinterColumns) {
+			// either all nil, or not identical
+			return
+		}
+	}
+
+	// things are identical if we've gotten this far, so move the subresources up
```
mengqiy: s/subresources/printerColumns

mengqiy:
It seems changes to `Gopkg.lock` is required to make travis happy.

pkg/crd/spec.go
```diff
@@ -36,6 +37,86 @@ type SpecMarker interface {
 	ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, version string) error
 }
 
+// mergeIdenticalSubresources checks to see if subresources are identical across
+// all versions, and if so, merges them into a top-level version.
+//
+// This assumes you're not using trivial versions.
```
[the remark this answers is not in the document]
DirectXMan12: in the case of trivial versions, this is effectively a noop, since you can't have non-top-level fields anyway

PR #298, opened 2019-08-06
✨ Support for oneOf Groups for Discriminate Types

seglberg:
These changes add support for the field marker: `kubebuilder:validation:OneOf` which will generate a `oneOf` block in the CRD structural definition[[1](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema)]. This could be expanded in the future to include more complex oneOf objects, however for now, it works to simply define fields which are mutually exclusive:

```go
// Discriminator Test Spec
type TestSpec struct {

	// +kubebuilder:validation:OneOf
	Cat map[string]string `json:"cat,omitempty"`

	// +kubebuilder:validation:OneOf
	Dog map[string]string `json:"dog,omitempty"`
}
```

becomes

```yaml
        spec:
          description: Discriminator Test Spec
          oneOf:
          - properties:
              cat: {}
            required:
            - cat
          - properties:
              dog: {}
            required:
            - dog
          properties:
            cat:
              additionalProperties:
                type: string
              type: object
            dog:
              additionalProperties:
                type: string
              type: object
          type: object
```

This then allows the CRD to enforce a mutually exclusive type:

```yaml
apiVersion: "inflow.arroyo.io/v1"
kind: Test
metadata:
  name: test-both
spec:
  cat: {"says": "meow"}
  dog: {"says": "woof"}
```

produces `"spec" must validate one and only one schema (oneOf). Found 2 valid alternatives`

```yaml
apiVersion: "inflow.arroyo.io/v1"
kind: Test
metadata:
  name: test-none
spec: {}
```

produces `"spec" must validate one and only one schema (oneOf). Found none valid`

---

[1] [https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema)

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -70,6 +70,9 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 	must(markers.MakeDefinition("optional", markers.DescribesField, struct{}{})).
 		WithHelp(markers.SimpleHelp("CRD validation", "specifies that this field is optional, if fields are required by default.")),
 
+	must(markers.MakeDefinition("kubebuilder:validation:OneOf", markers.DescribesField, struct{}{})).
+		WithHelp(markers.SimpleHelp("CRD validation", "specifies that this field is part of a oneOf group")),
+
 	must(markers.MakeDefinition("nullable", markers.DescribesField, Nullable{})).
 		WithHelp(Nullable{}.Help()),
 }
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -350,6 +350,10 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *v1beta1.JSO
 		fieldName := jsonOpts[0]
 		inline = inline || fieldName == "" // anonymous fields are inline fields in YAML/JSON
 
+		fieldMarkedOptional := (field.Markers.Get("kubebuilder:validation:Optional") != nil || field.Markers.Get("optional") != nil)
+		fieldMarkedRequired := (field.Markers.Get("kubebuilder:validation:Required") != nil)
+		fieldMarkedOneOf := (field.Markers.Get("kubebuilder:validation:OneOf") != nil)
+
 		// if no default required mode is set, default to required
 		defaultMode := "required"
 		if ctx.PackageMarkers.Get("kubebuilder:validation:Optional") != nil {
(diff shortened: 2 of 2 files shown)


DirectXMan12:
/hold

a) you'll need to sign the CLA

b) I'd like to see a more comprehensive design for this

c) this might be a foot-gun considering some uses of one-of could produce "non-structural" schemata (https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema)

d) we generally discourage this pattern, preferring unions instead (we'll need to support the `x-kubernetes-union` tag, though)

seglberg:
I'll gladly sign the CLA if this ends up going anywhere significant since it sounds like I may have gone down the wrong road here.

I didn't seem to come across any concepts of a Union type when digging at the current sources. I had been modeling my APIs after Kubernetes 1.15 types such as VolumeSource. Is there more information or docs you could point me towards so that I may formulate a more comprehensive design for this?

seglberg:
Ah. So I just ran across https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190425-structural-openapi.md

I'm still getting my bearing around the Kubernetes project structure... but it looks like the Union extensions have yet to land?

DirectXMan12:
The union thing is a new feature that codifies a pattern that we've had for a while.  And yeah, it's hard to navigate those some times.

union appears to have half-landed, but I think we can produce the equivalent openapi off of the standard union markers introduced in the union kep linked in that doc (`+union`, `+unionDeprecated`, and `+unionDiscriminator`)

@sttts

DirectXMan12:
also @apelisse @jennybuckley might know more about the state of the union bits, but I don't see XUnion as part of the JSON schema types yet, but it's in kube-openapi already.

seglberg:
Producing (or "unfolding" as one of the docs put it) the equivalent openapi schema shouldn't be too difficult until there is full support for the extension. That begs the question. In the example:

```yaml
x-kubernetes-unions:
- discriminator: type
  fields:
    foo: Foo
    bar: Bar
oneOf:
  - required: ["foo"]
  - required: ["bar"]
properties:
  type:
    type: string
    enum:
    - Foo
    - Bar
```

should the discriminator inside the oneOf be set to a narrowed enum? 

Until the apiserver supports setting the discriminator field for us, we need to lock it down so users cannot do something like setting the field `foo` and set the type to `Bar`, since this will currently pass validation.

For now, I would think a more correct expansion would be:
```yaml
oneOf:
  - required: ["foo"]
    properties:
        type:
            enum:
            - Foo
  - required: ["bar"]
    properties:
        type:
            enum:
            - Bar
```

DirectXMan12:
There's an OpenAPI concept of a discriminator.  That should be sufficient to prevent that case.

apelisse:
+1 on what @DirectXMan12 says, we don't know if it's going to land in 1.16, we can't really agree on all the semantics for union, but we're working on it. Also, I don't think I'm opposed to have the "oneOf" field being generated by kube-builder. The "union" marker that we're proposing in the [KEP](https://github.com/kubernetes/enhancements/issues/1027) is orthogonal and still requires validation.

sttts:
I think unfolding does not harm. The first one looks good:

```
oneOf:
  - required: ["foo"]
  - required: ["bar"]
properties:
  type:
    type: string
    enum:
    - Foo
    - Bar
```

But the other one is wrong because `{"foo": 1, "bar": 2, "type":"Foo"}` passes that spec:

```
oneOf:
  - required: ["foo"]
    properties:
        type:
            enum:
            - Foo
  - required: ["bar"]
    properties:
        type:
            enum:
            - Bar
```

displague:
This approach seems like it could block adopting an approach that permits multiple `oneOf` required groups within a type as https://github.com/kubernetes-sigs/controller-tools/issues/461#issuecomment-654524139 seeks.

fieldA or fieldB are required and
fieldC or fieldD are required

PR #299, opened 2019-08-07
Limiting description of fields in CRD's OpenAPI Schema

droot:
This PR adds support for limiting the description length for fields in OpenAPI Schema. It introduces a parameter `maxDescLen` for CRD generator. 

- Default value (0) means keep the description as-is for backward compatibility
- Negative value (-1) means drop the description field completely
- Positive (n) value will truncate the description to closest sentence boundary if description exceeds `maxDescLen`.

fixes https://github.com/kubernetes-sigs/kubebuilder/issues/906

### DIFF
--- a/pkg/crd/desc_visitor.go
+++ b/pkg/crd/desc_visitor.go
@@ -1,0 +1,78 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crd
+
+import (
+	"strings"
+	"unicode"
+
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+)
+
+// TruncateDescription truncates the description of fields in given schema if it
+// exceeds maxLen.
+// It tries to chop off the description at the closest sentence boundary.
+func TruncateDescription(schema *apiext.JSONSchemaProps, maxLen int) {
+	EditSchema(schema, descVisitor{maxLen: maxLen})
+}
+
+// descVisitor recursively visits all fields in the schema and truncates the
+// description of the fields to specified maxLen.
+type descVisitor struct {
+	// maxLen is the maximum allowed length for decription of a field
+	maxLen int
+}
+
+func (v descVisitor) Visit(schema *apiext.JSONSchemaProps) SchemaVisitor {
+	if schema == nil {
+		return v
+	}
+	if v.maxLen < 0 {
+		return nil /* no further work to be done for this schema */
+	}
+	if v.maxLen == 0 {
+		schema.Description = ""
+		return v
+	}
+	if len(schema.Description) > v.maxLen {
+		schema.Description = truncateString(schema.Description, v.maxLen)
+		return v
+	}
+	return v
+}
+
+// truncateString truncates given desc string if it exceeds maxLen. It may
+// return string with length less than maxLen even in cases where original desc
+// exceeds maxLen because it tries to chop off the desc at the closest sentence
+// boundary to avoid incomplete sentences.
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.6 // indirect
 	github.com/kr/pretty v0.1.0 // indirect
+	github.com/markbates/inflect v1.0.4
```
mengqiy: We no longer use this.
Now we use `github.com/gobuffalo/flect`.
I can't any place in your code that is using it. So drop this line and then run `go mod tidy`.

pkg/crd/desc_visitor.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+)
+
+// TruncateDescription truncates the description of fields in given schema if it
+// exceeds maxLen.
+// It tries to chop off the description at the closest sentence boundary.
+func TruncateDescription(schema *apiext.JSONSchemaProps, maxLen int) *apiext.JSONSchemaProps {
+	outSchema := schema.DeepCopy()
```
mengqiy: Do we really need to deepcopy it? The schema may be huge.

pkg/crd/gen.go
```diff
@@ -39,6 +39,12 @@ type Generator struct {
 	// Kubernetes API servers.  The storage version's schema will be used as
 	// the CRD's schema.
 	TrivialVersions bool `marker:",optional"`
+
+	// MaxDescLen limits the description length of each field in CRD's OpenAPI schema.
+	//
+	// 0 value (default) indicates no limit on description of fields.
+	// -1 value indicates drop the description for all fields.
+	MaxDescLen int `marker:",optional"`
```
mengqiy: IMO it'd better to change it to a pointer:
- nil means no limit (default)
- 0 means drop description completely
- n means at most n characters

@DirectXMan12 Pointer should work here, right? If not, we probably should fix it.

pkg/crd/desc_visitor_test.go
```diff
@@ -0,0 +1,94 @@
...
+limitations under the License.
+*/
+
+package crd_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	// . "github.com/onsi/gomega/gstruct"
```
mengqiy: Drop this

pkg/crd/desc_visitor.go
DirectXMan12: eh.  It's probably fine.  It's not like we're doing this at scale.  Although, I'd just move this out of the call, and make `TruncateDescription` not return a value, and then the caller can choose

pkg/crd/desc_visitor.go
```diff
@@ -0,0 +1,72 @@
...
+		return v
+	}
+	if len(schema.Description) > v.maxLen {
+		schema.Description = truncateString(schema.Description, v.maxLen)
+		return v
+	}
+	return v
+}
+
+func truncateString(desc string, maxLen int) string {
```
DirectXMan12: godoc, please

pkg/crd/desc_visitor.go
```diff
@@ -0,0 +1,72 @@
...
+	desc = desc[0:maxLen]
+
+	// Trying to chop off at closest sentence boundary.
+	if n := strings.LastIndexAny(desc, "."); n > 0 {
```
DirectXMan12: ". " (period-space) is safer for sentences, but we should actually probably use unicode.SentenceTerminal or something similar.

pkg/crd/gen.go
DirectXMan12: agree with @mengqiy here

droot: Good suggestion, refactored it to take ptr.

pkg/crd/desc_visitor.go
droot: refactored it as suggested by @DirectXMan12

pkg/crd/desc_visitor.go
droot: Done.

pkg/crd/desc_visitor.go
droot: I couldn't find some quick helper for SentenceTerminal in `unicode` pkg --> https://golang.org/pkg/unicode/. Quick search suggest, I can create a helper fn using chars listed in this page --> https://www.unicode.org/review/pr-23.html. Do you think it is worth the effort here ?

PR #303, opened 2019-08-09
Treat unstructured.Unstructured type as known type.

### LINKED ISSUE #301: CRD cannot be generated for Unstructured type
We got a type that has `*unstructured.Unstructured` as [one of its fields](https://github.com/crossplaneio/crossplane/blob/master/pkg/apis/workload/v1alpha1/types.go#L150) and we're generating CRD off of it. I am working on transition to controller-tools v0.2.0 and running into the following error:
```
github.com/crossplaneio/crossplane/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/unstructured.go:45:2: encountered struct field "Object" without JSON tag in type "Unstructured"
```

Looked into the source code and there really isn't a tag for `Object` field, however, even if I manually write the json tag, it returns the following:
```
github.com/crossplaneio/crossplane/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/unstructured.go:45:20: map values must be a named type, not *ast.InterfaceType
```

Is there a way to have controller-tools handle this? I imagine it's more complicated due to having `interface{}` as value but I'd appreciate if we can find a way for it to work. Note that it worked with version 0.1.9

You can see the ongoing work here https://github.com/crossplaneio/crossplane/pull/665

You can see the versions used in `Gopkg.toml`.

To reproduce, checkout the PR and run `make manifests`

You can also try to have `*unstructured.Unstructured` typed field in any struct and try to generate CRD.

muvaf:
I think this fixes #301 . Basically treat `unstructured.Unstructured` type exactly same with `RawExtension`. 

Known packages are referred as:
```
// KnownPackages overrides types in some comment packages that have custom validation
// but don't have validation markers on them (since they're from core Kubernetes).
```
And I think we might be able to treat Unstructured like core k8s types because of the statement in  https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/unstructured.go#L31 

But I don't really have a strong opinion on this.

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -68,6 +68,13 @@ var KnownPackages = map[string]PackageOverride{
 		p.AddPackage(pkg) // get the rest of the types
 	},
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured": func(p *Parser, pkg *loader.Package) {
+		p.Schemata[TypeIdent{Name: "Unstructured", Package: pkg}] = apiext.JSONSchemaProps{
+			Type: "object",
+		}
+		p.AddPackage(pkg) // get the rest of the types
+	},
+
 	"k8s.io/apimachinery/pkg/util/intstr": func(p *Parser, pkg *loader.Package) {
 		p.Schemata[TypeIdent{Name: "IntOrString", Package: pkg}] = apiext.JSONSchemaProps{
 			AnyOf: []apiext.JSONSchemaProps{


DirectXMan12:
Generally, you should be using `runtime.RawExtension` here instead

DirectXMan12:
since unstructured also requires you have objectmeta

muvaf:
The way that `RawExtension` is handled is basically representing it as `type: object` in the CRD. I see that `Unstructured` could be considered as a subset of `RawExtension` type (please correct me if I'm wrong). So, unless more specific properties we can add to `Unstructured` here in this file, treating it like `RawExtension` should be fine in my opinion.

Why not just have everyone switch to use `RawExtension`? Dropping support for a type in CRDs (in this case `Unstructured`) is very hard for the consumers of the package because it means that the consumer has to change their interface to comply, which can exponentially affect upstream consumer tree of the package. For example, in [Crossplane](https://github.com/crossplaneio/crossplane), we've got stack definitions that provision the infrastructure and handles the installation of the applications. People have their own stacks as well (imagine a bunch of CR YAMLs). If we have to change the type of the object in the CRD, we'd need to communicate this to everyone and change just too much stuff. 

That could even be a blocker for migration to v0.2.0. We had to use our own fork for this fix in https://github.com/crossplaneio/crossplane/pull/675

DirectXMan12:
>  If we have to change the type of the object in the CRD, we'd need to communicate this to everyone and change just too much stuff.

Ack, that seems reasonable, but we should add a warning.

AFAIK, `unstructured.Unstructured` is not intended to be part of a public API -- it's a generic deserialization target for unknown types.  You can see this if you look at the public API.  So it's related to Unstructured, but neither a superset or a subset, AIUI.

muvaf:
That is a strong argument for not using `unstructured.Unstructured`.

Is there a good way to tell whether a struct is a part of the public API and intended to be used externally?

DirectXMan12:
talked to API machinery folks a bit to see if I could get a better answer, and what I got back was "it might occasionally be acceptable to use unstructured, but generally RawExtension is preferred in non-internal version usecases".

A number of things here are just "look at what Kubernetes does" if it's not explicitly listed in the api-conventions guidelines document, unfortunately.

muvaf:
@DirectXMan12 Done. Thank you!

pkg/crd/known_types.go
```diff
@@ -68,6 +68,14 @@ var KnownPackages = map[string]PackageOverride{
 		p.AddPackage(pkg) // get the rest of the types
 	},
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured": func(p *Parser, pkg *loader.Package) {
+		p.Schemata[TypeIdent{Name: "Unstructured", Package: pkg}] = apiext.JSONSchemaProps{
+			// TODO(directxman12): regexp validation for this (or get kube to support it as a format value)
```
DirectXMan12: get rid of this comment, and we're good to go (this comment doesn't make sense for this override)

PR #306, opened 2019-08-13
⚠Change autogenerated / do not edit comment in deepcopy files

vincepri:

This PR changes the comment warning users to `DO NOT EDIT` and makes it similar to other k8s generation tools. I stumbled upon this while running verify-boilerplate from k/k, which looks for "DO NOT EDIT" in a file to consider it autogenerated.


chuckha:
For context: https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.py#L211

This is copied around to other projects and has this regex hardcoded.

DirectXMan12:
can you change helpgen (cmd/helpgen) too?  It's just for internal use, might might as well be consistent.  Otherwise LGTM.

vincepri:
@DirectXMan12 Done! I also ran go generate to regen the other files

Issue #307, opened 2019-08-16
Feedback on creating conversion webhook CRDs using controller-tools


munnerz:
I've been spending the last few days implementing a conversion webhook for cert-manager.

For context, cert-manager does *not* use Kubebuilder, however it *does* use controller-tools to generate CustomResourceDefinition resources. We are also utilising the upstream `conversion-gen` tool to generate our conversions, as opposed to defining one of our external versions as the 'hub version' (as described in the kubebuilder documentation).

In the process of doing this, I ran into (and in some cases, fixed) a number of issues. I thought I'd open this issue in order to track improving support for multi-version CRDs in controller-gen 😄 

1) When `trivialVersions=false`, there's currently no way to *exclude* a particular API version from the generated CRD, meaning my own internal version is also listed (and schema generated) on my CRD under the `versions` field. I've put together a patch that adds a `skipversion` marker that can be added to a struct definition to have that particular version *removed* from the generated CRD

2) Improper output when schemas/subresources/additionalPrinterColumns are the same between versions: Kubernetes will reject CRD manifests that define multiple versions *if* the schema, subresource or additionalPrinterColumn definitions between the two versions are identical. Currently, controller-gen will output `version` stanzas that *all* contain these fields, even if they are identical, which means the resource cannot be applied.

3) No way to add a `conversion` stanza to the generated CRD. I *think* this is because Kubebuilder itself recommends using a kustomize overlay to add this extra stanza, which makes a lot of sense. On the other hand, we don't currently use Kustomize which makes this problematic. Could we investigate adding some additional markers to allow configuring this? For what its worth, our configuration here is super predictable, so this mechanism would actually be really feasible in the cert-manager case:

```yaml
  conversion:
    strategy: Webhook
    webhookClientConfig:
      service:
        name: kubernetes
        namespace: default
        path: /apis/conversion.certmanager.k8s.io/v1beta1/certificaterequests
```

That said, I imagine you don't want to overload the configuration surface with too many markers, so this may need some thought 😄 

I will put in a PR for (1), but I am unsure the best way to implement (2), and I'm even moreso unsure whether we even want to implement (3) 😄 

/cc @DirectXMan12 @sttts @JoshVanL

DirectXMan12:
> 1 (exclude an api version)

Will take a look at the patch, seems reasonable

> 2 (improperly merged output)

there's a fix for 2 in #290.  Should be in the most recent release (v0.2.0-rc.0)

[...]

PR #308, opened 2019-08-16
✨ Add skipversion marker to selectively exclude CRD versions

munnerz:
Adds a new marker that allows excluding particular CRD versions from the generated CRD output.

This is specifically useful to me in order to support having a stable 'internal' version that is used for a basis for conversion (we prefer to define this instead of using an external version as our 'hub' as we don't want to have to rewrite older conversion functions as we deprecate APIs 😄).

For more context, see #307


DirectXMan12:
/lgtm
/approve

Can you follow up with a change to the integration test that tests this?

PR #310, opened 2019-08-22
:bug: Write YAML stream with correct start marker

mcristina422:
I came across an issue reading `role.yaml` using a package in k8s.io/apimachinery/pkg/util/yaml. The blank new line at the beginning of the file made the decoder error. I _think_ this violates the YAML spec, according to https://yaml.org/spec/1.2/spec.html#id2800132, but I'm definitely not an expert. 

This changes the YAML generator to separate objects simply with `---\n`. 

I was curious about how the internals of https://github.com/kubernetes-sigs/kustomize printed mutli object yaml, but I didn't dig deep enough to find the implementation.


pkg/genall/genall.go
```diff
@@ -103,7 +103,7 @@ func (g GenerationContext) WriteYAML(itemPath string, objs ...interface{}) error
 		if err != nil {
 			return err
 		}
-		n, err := out.Write(append([]byte("\n---\n"), yamlContent...))
+		n, err := out.Write(append([]byte("---\n"), yamlContent...))
```
DirectXMan12: is `yaml.Marshal` guaranteed to produce a newline at the end?

DirectXMan12: (i.e. is this safe to not have the extra newline?)

DirectXMan12: I think we actually have a helper burried in the k8s libraries for this

DirectXMan12: (nope, just a decoder)

mcristina422: I think `yaml.Marshal` is guaranteed to produce a newline

PR #311, opened 2019-08-26
:sparkles: pkg/crd/spec.go: set default ListKind and Singular CRD names

estroz:
A CRD's ListKind and Singular names are not added by default, but are easy to infer from group data and therefore should be set by default.


DirectXMan12:
fair enough.  This matches what the API server does

/lgtm
/approve

PR #314, opened 2019-09-06
📖Update builtinToType comments to explain support for int32 and int64

daneharrigan:
This PR is to address the request made the following:
- https://github.com/kubernetes-sigs/kubebuilder/issues/921#issuecomment-527563284
- https://github.com/kubernetes-sigs/kubebuilder/pull/975#issuecomment-527563813

I didn't mention `resource.Quantity` in the changes because I saw no sign of it in `builtinToType`. Please let me know if I've missed something and thanks!

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -389,8 +389,8 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *v1beta1.JSO
 }
 
 // builtinToType converts builtin basic types to their equivalent JSON schema form.
-// It *only* handles types allowed by the kubernetes API standards, meaning
-// no floats (technically we shouldn't allow int64 or plain int either, but :shrug:).
+// It *only* handles types allowed by the kubernetes API standards. Floats are not
+// allowed.
 func builtinToType(basic *types.Basic) (typ string, format string, err error) {
 	// NB(directxman12): formats from OpenAPI v3 are slightly different than those defined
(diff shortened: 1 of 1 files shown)


DirectXMan12:
> I didn't mention resource.Quantity in the changes because I saw no sign of it in builtinToType

Yeah -- it's not a built-in from the Go perspective, so it's handed elsewhere (in overrides, technically).

pkg/crd/schema.go
```diff
@@ -389,8 +389,9 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *v1beta1.JSO
 }
 
 // builtinToType converts builtin basic types to their equivalent JSON schema form.
-// It *only* handles types allowed by the kubernetes API standards, meaning
-// no floats (technically we shouldn't allow int64 or plain int either, but :shrug:).
+// It *only* handles types allowed by the kubernetes API standards. Numbers can be
+// int32 and int64. Floats are not allowed. (technically we shouldn't allow
+// int64 or plain int either, but :shrug:).
```
DirectXMan12: you can drop the `int64` from the list of `shouldn't allow`

daneharrigan: I think this is more along the lines of what you're thinking. And thanks for the direction

DirectXMan12:
P.S. it's generally not a good idea to at-mention people in PR descriptions, as they'll get pinged whenever that merge commit moves around.  I've fixed it for you this time.

PR #317, opened 2019-09-12
:sparkles: add support for pointers as map values

knight42:
for the time being `controller-gen` does not accept pointers as map values and IMHO my patch should fix this problem as shown below.

```
$ tree .
.
├── api.go
├── go.mod
└── go.sum

0 directories, 3 files
```

```
$ cat api.go
package crd

import (
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

// +kubebuilder:object:generate=true
// +groupName=example.org

type Env struct {
	A int    `json:"a"`
	B string `json:"b"`
}

type App struct {

	Envs map[string]*Env `json:"envs,omitempty"`
}
```

using the latest controller-gen(https://github.com/kubernetes-sigs/controller-tools/commit/9d85d107df07a210d8b23d2b612365fd66bbe187)

```
$ /tmp/controller-gen-9d85d107d crd paths=./... output:crd:stdout

// invalid crd omitted

/private/tmp/crd/api.go:20:18: map values must be a named type, not *ast.StarExpr
run `controller-gen crd paths=./... output:crd:stdout -w` to see all available markers, or `controller-gen crd paths=./... output:crd:stdout -h` for usage
```

my patch:

```
$ /tmp/controller-gen-fix crd paths=./... output:crd:stdout
```

<summary>
<details>
<pre>
---
metadata:
  annotations:
  creationTimestamp: null
  name: apps.example.org
spec:
  group: example.org
  names:
    kind: App
    listKind: AppList
    plural: apps
    singular: app
  scope: Namespaced
  validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          type: string
        envs:
          additionalProperties:
            properties:
              a:
                type: integer
              b:
                type: string
            required:
            - a
            - b
            type: object
          type: object
        kind:
          type: string
        metadata:
          type: object
      type: object
  version: crd
  versions:
  - name: crd
    served: true
    storage: true
status:
  acceptedNames:
    kind: ""
    plural: ""
  conditions: []
  storedVersions: []
</pre>
</details>
<summary>

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -297,6 +297,8 @@ func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *apiext.JSONSchemaPro
 			ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map values must be a named type, not %T", mapType.Value), mapType.Value))
 			return &apiext.JSONSchemaProps{}
 		}
+	case *ast.StarExpr:
+		valSchema = typeToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
 	default:
 		ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map values must be a named type, not %T", mapType.Value), mapType.Value))
 		return &apiext.JSONSchemaProps{}
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -91,6 +91,9 @@ type CronJobSpec struct {
 	// This tests string slices are allowed as map values.
 	StringSliceData map[string][]string `json:"stringSliceData,omitempty"`
 
+	// This tests pointers are allowed as map values.
+	PtrData map[string]*string `json:"ptrData,omitempty"`
+
 	// This tests that markers that are allowed on both fields and types are applied to fields
 	// +kubebuilder:validation:MinLength=4
 	TwoOfAKindPart0 string `json:"twoOfAKindPart0"`
--- a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
+++ b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
@@ -5038,6 +5038,11 @@ spec:
                 description: This tests that pattern validator is properly applied.
                 pattern: ^$|^((https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))$
                 type: string
+              ptrData:
+                additionalProperties:
+                  type: string
+                description: This tests pointers are allowed as map values.
+                type: object
               schedule:
                 description: The schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
                 type: string
--- a/pkg/webhook/zz_generated.markerhelp.go
+++ b/pkg/webhook/zz_generated.markerhelp.go
@@ -57,11 +57,11 @@ func (Config) Help() *markers.DefinitionHelp {
 				Details: "",
 			},
 			"Name": markers.DetailedHelp{
-				Summary: "indicates the name of this webhook configuration.",
+				Summary: "indicates the name of this webhook configuration. Should be a domain with at least three segments separated by dots",
 				Details: "",
 			},
 			"Path": markers.DetailedHelp{
-				Summary: "specifies that path that the API server should connect to this webhook on.",
+				Summary: "specifies that path that the API server should connect to this webhook on. Must be prefixed with a '/validate-' or '/mutate-' depending on the type, and followed by $GROUP-$VERSION-$KIND where all values are lower-cased and the periods in the group are substituted for hyphens. For example, a validating webhook path for type batch.tutorial.kubebuilder.io/v1,Kind=CronJob would be /validate-batch-tutorial-kubebuilder-io-v1-cronjob",
 				Details: "",
 			},
 		},


yuchengwu:
any progress? or is there a way to workaround this situation

knight42:
@yuchengwu just declare a new type for the pointer type.

SimonCqk:
Does this version release in v0.2.2? I have the same issue when using controller-gen.

SimonCqk:
@knight42 PR has to fix conflicts.

erdrix:
Would be great to merge it, for dependent tools : https://github.com/operator-framework/operator-sdk/issues/2485

knight42:
rebased

DirectXMan12:
This is not a fix for the listed issue, and needs more elaboration

knight42:
@DirectXMan12 I wonder why this is not a fix? I think the problem of the listed issue is that for the time being `controller-gen` does not accept pointers as map values and IMHO my patch should fix this problem as shown below.

```
$ tree .
.
├── api.go
├── go.mod
└── go.sum

0 directories, 3 files
```

```
$ cat api.go
package crd

import (
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

// +kubebuilder:object:generate=true
// +groupName=example.org

type Env struct {
	A int    `json:"a"`
	B string `json:"b"`
}

type App struct {

	Envs map[string]*Env `json:"envs,omitempty"`
}
```

using the latest controller-gen(https://github.com/kubernetes-sigs/controller-tools/commit/9d85d107df07a210d8b23d2b612365fd66bbe187)

```
$ /tmp/controller-gen-9d85d107d crd paths=./... output:crd:stdout

// invalid crd omitted

/private/tmp/crd/api.go:20:18: map values must be a named type, not *ast.StarExpr
run `controller-gen crd paths=./... output:crd:stdout -w` to see all available markers, or `controller-gen crd paths=./... output:crd:stdout -h` for usage
```

my patch:

```
$ /tmp/controller-gen-fix crd paths=./... output:crd:stdout

---
metadata:
  annotations:
  creationTimestamp: null
  name: apps.example.org
spec:
  group: example.org
  names:
    kind: App
    listKind: AppList
    plural: apps
    singular: app
  scope: Namespaced
  validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          type: string
        envs:
          additionalProperties:
            properties:
              a:
                type: integer
              b:
                type: string
            required:
            - a
            - b
            type: object
          type: object
        kind:
          type: string
        metadata:
          type: object
      type: object
  version: crd
  versions:
  - name: crd
    served: true
    storage: true
status:
  acceptedNames:
    kind: ""
    plural: ""
  conditions: []
  storedVersions: []

```

DirectXMan12:
The listed issue is that controller-gen is trying to generate a CRD definition for a type not used in the CRD.  It's breaking because it's not types we accept in a CRD, but the actually issue is that the user didn't intend for that type to be parsed

rajathagasthya:
@DirectXMan12 This needs to be fixed regardless, right? This is a case I am hitting too where I refer to a pointer to a struct defined in the CRD in a map and `controller-gen` complains. Much like the example in https://github.com/kubernetes-sigs/controller-tools/pull/317#issuecomment-582252963. 

Or is there a strict guideline that map values cannot be pointers?

knight42:
@DirectXMan12 well... you're right, I have misunderstood, sorry about that. I've removed the "Fix ..." line in the description, but I think that pointers should be accepted as map values nevertheless.

cscetbon:
Hey guys, any news regarding that PR ? we had to fork operator-sdk in order to use that PR https://github.com/Orange-OpenSource/casskop/blame/master/documentation/development.md#L62

DirectXMan12:
> Or is there a strict guideline that map values cannot be pointers?

In generally, you're pretty much never supposed to use maps in k8s.  When you do, it's pretty much exclusively `map[string-ish]string-ish`.  There's a couple of exceptions for legacy cases, but in general, if you're using a pointer, it's an indication that you might be doing something against k8s practices.

DirectXMan12:
More specifically, `map[string-ish]struct-ish` is canonically *not* supposed to be a thing.

rajathagasthya:
Yeah, I read in the API conventions that maps are not preferred. But then there is https://github.com/kubernetes/kubernetes/issues/8190#issuecomment-538605819 :)

Anyway, we have a bunch of code (based on operator-sdk which we are moving to kubebuilder v2) that uses maps (unfortunately) which we cannot change right away (without bumping apiversion and writing conversion). So we are not able to use controller-tools to generate a CRD :(

cscetbon:
If you read the next comment it says the opposite https://github.com/kubernetes/kubernetes/issues/8190#issuecomment-539975062. And yeah same here, without this PR we can't use it.

DirectXMan12:
the key there is "non-standard keys", I believe.  See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

seglberg:
We are also interested in this. While we don't use maps in the spec section of our CRDs, we do use them extensively in our status fields.

The biggest benefit for us would be that this would significantly clean up our duplicated types. We already work around the map issue by re-defining our pointer types, so whether or not this gets accepted, we will still use maps. I think the biggest benefit of this PR is that it eliminates the need of an (ugly) workaround.

rajathagasthya:
We don't use it in spec too, just in status. So the reason for "list are preferred over maps" that it's hard to distinguish between magic keywords and just strings (as described in https://github.com/kubernetes/kubernetes/issues/2004#issuecomment-60641437) is not _really_ applicable I think. But yeah, we can workaround it for now as this is presumably a bigger API decision not to support maps.

rajathagasthya:
@seglberg Is the workaround just to declare a new type for the pointer to old type like `type PtrFoo *Foo`? I tried that, but deepcopy seems to have issues when you do that and causes build failures. 

```
api/v1beta1/zz_generated.deepcopy.go:285:10: (*in).DeepCopyInto undefined (type PtrNodeStatus has no field or method DeepCopyInto)
```

knight42:
@rajathagasthya Did you run `make generate`?

rajathagasthya:
@knight42 Yep, I don't see deepcopy funcs generated for the new type.

Jeffwan:
Any updates on this issue? Is the change acceptable? @DirectXMan12  @knight42 

I am using kubebuilder2 and also have this issue. The worst thing is I am using some common packages in my API definition, the dependency has map pointer usage as well. Without the change, I can not move forward..

ifilonenko:
any status on this? I too am using kubebuilderv2 and have bumped into this issue

DirectXMan12:
At this point, I've gotten enough feedback that this is seriously blocking folks that I'm gonna reconsider and approve this.  I want to follow up with a "warn about this thing" PR, though, since we still don't recommend this, but there's enough cases (esp around older CRs, embedded types from other systems, etc) that I think I was a bit too firm on blocking this earlier.  Apologies for the delays/troubles I've caused with holding this up.

liubog2008:
Could you release a new version contains this fix？ @DirectXMan12 @ifilonenko

cscetbon:
@DirectXMan12 when can we expect a new version like asked 👆

PR #321, opened 2019-09-18
:sparkles: Add support for CRD v1 to crdschema generator

marun:
This PR adds support to crdschema generator for CRD v1 manifests. ~The schema generated for a v1 CRD is also validated for [structural compatibility](https://kubernetes.io/blog/2019/06/20/crd-structural-schema/).~

Upon the advice of @sttts I have removed the structural schema validation since the implementation requires the use of code intended to be internal to the kube apiserver. Schema construction is intended to be correct, but ultimately validation can only be authoritatively performed on submission to an apiserver.  

cc: @sttts

### DIFF
--- a/pkg/schemapatcher/gen.go
+++ b/pkg/schemapatcher/gen.go
@@ -22,10 +22,12 @@ import (
 	"path/filepath"
 
 	"gopkg.in/yaml.v3"
-	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextv1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog"
 	kyaml "sigs.k8s.io/yaml"
 
 	crdgen "sigs.k8s.io/controller-tools/pkg/crd"
@@ -146,21 +148,26 @@ func (g Generator) Generate(ctx *genall.GenerationContext) (result error) {
 			existingInfo.Versions[ver] = struct{}{}
 		}
 
-		allSame := true
-		firstSchema := existingInfo.NewSchemata[someVer]
-		for ver := range existingInfo.Versions {
-			otherSchema, hasSchema := existingInfo.NewSchemata[ver]
-			if !hasSchema || !equality.Semantic.DeepEqual(firstSchema, otherSchema) {
-				allSame = false
-				break
+		globalSchemaWritten := false
+		if existingInfo.WriteGlobalSchema {
+			allSame := true
+			firstSchema := existingInfo.NewSchemata[someVer]
+			for ver := range existingInfo.Versions {
+				otherSchema, hasSchema := existingInfo.NewSchemata[ver]
+				if !hasSchema || !equality.Semantic.DeepEqual(firstSchema, otherSchema) {
+					allSame = false
+					break
+				}
+			}
+			if allSame {
+				if err := existingInfo.setGlobalSchema(); err != nil {
+					return fmt.Errorf("failed to set global firstSchema for %s: %v", existingInfo.GroupKind, err)
+				}
+				globalSchemaWritten = true
 			}
 		}
 
-		if allSame {
-			if err := existingInfo.setGlobalSchema(); err != nil {
-				return fmt.Errorf("failed to set global firstSchema for %s: %v", existingInfo.GroupKind, err)
-			}
-		} else {
+		if !globalSchemaWritten {
 			if err := existingInfo.setVersionedSchemata(); err != nil {
 				return fmt.Errorf("failed to set versioned schemas for %s: %v", existingInfo.GroupKind, err)
 			}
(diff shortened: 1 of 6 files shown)


marun:
Ready for review

pkg/schemapatcher/gen.go
```diff
@@ -356,13 +384,75 @@ func crdsFromDirectory(ctx *genall.GenerationContext, dir string) (map[schema.Gr
...
+}
+
+// validateStructural checks whether the given openapi schema is
+// structurally valid (required for CRD v1).
+func validateStructural(schema *apiextv1b1.JSONSchemaProps) error {
+	marshalledSchema, err := json.Marshal(schema)
+	if err != nil {
+		return err
+	}
+	genericSchema := &apiext.JSONSchemaProps{}
```
marun: @sttts This check for a structural schema is using internal methods (as per discussion on #323). Is there an alternate way to validate that a schema is structural, or is that capability intentionally limited to the k/k repo?

sttts: like all other validations in kube they are supposed to be internal. 

Before calling NewStructural, you are supposed to validate with apiextensions-apiserver/pkg/apis/apiextensions/validation. That's of course tricky as you don't have a CRD here yet.

tl/dr: I am still not convinced we should validate anything here about the schema. We should rather make schemas structural by construction medium term. Any failing validation here would be a controller-tools bug, not a user error.

marun: The latest iteration removes the validation.

pkg/schemapatcher/gen.go
```diff
@@ -330,24 +341,24 @@ func crdsFromDirectory(ctx *genall.GenerationContext, dir string) (map[schema.Gr
...
-		if err := kyaml.Unmarshal(rawContent, &actualCRD); err != nil {
-			continue
+		if typeMeta.APIVersion == apiextv1APIVersion {
+			groupKind, versions = unmarshalCRDV1(rawContent)
+			// v1 CRD does not have a global schema
+		} else if typeMeta.APIVersion == apiextv1b1APIVersion {
+			groupKind, versions = unmarshalCRDV1beta1(rawContent)
+			// v1beta1 CRD has a global schema
+			writeGlobalSchema = true
 		}
```
sttts:
```
else {
fatal error? Or at least a bold warning?
}
```

marun: Done

pkg/schemapatcher/gen.go
```diff
@@ -330,24 +341,26 @@ func crdsFromDirectory(ctx *genall.GenerationContext, dir string) (map[schema.Gr
...
+		} else if typeMeta.APIVersion == apiextv1b1APIVersion {
+			groupKind, versions = unmarshalCRDV1beta1(rawContent)
+			// v1beta1 CRD has a global schema
+			writeGlobalSchema = true
 		} else {
-			versions = make(map[string]struct{}, len(actualCRD.Spec.Versions))
-			for _, ver := range actualCRD.Spec.Versions {
-				versions[ver.Name] = struct{}{}
-			}
+			return nil, fmt.Errorf("Unknown CRD apiVersion %q", typeMeta.APIVersion)
```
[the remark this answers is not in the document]
sttts: API version

pkg/schemapatcher/gen.go
```diff
@@ -330,24 +341,26 @@ func crdsFromDirectory(ctx *genall.GenerationContext, dir string) (map[schema.Gr
...
+		}
+		if groupKind == nil {
+			continue
```
sttts: can this happen now?

marun: Yes - both unmarshal functions returns `nil, nil` if unmarshaling fails (silent fail). This was the previous behavior so I figured it should be retained.  Though maybe klog should be in play to provide output when this happens?

sttts: Would expect a warning, yes.

sttts: Why is there no error return value in unmarshalCRDV1/v1beta1?

munnerz:
Taking a quick look through this PR, I don't see any way to choose the output version when invoking the generator? It'd be great to continue to support `v1beta1` so controller authors targetting Kubernetes <1.16 can use the latest controller-tools features whilst still generating compatible CRDs 😄

sttts:
@munnerz this PR is not about the crd generator, but the schema patcher. I agree with your request, but that's a follow-up as it is unrelated. PR welcome :)

pkg/schemapatcher/gen.go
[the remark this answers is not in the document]
marun: Done

pkg/schemapatcher/gen.go
marun: Updated to return error values and log them as warnings.

pkg/schemapatcher/gen.go
```diff
@@ -330,24 +342,29 @@ func crdsFromDirectory(ctx *genall.GenerationContext, dir string) (map[schema.Gr
...
+		if typeMeta.APIVersion == apiextv1APIVersion {
+			groupKind, versions, err = unmarshalCRDV1(rawContent)
+			if err != nil {
+				klog.Warningf("failed to unmarshal to CRD v1: %v", err)
+			}
+			// v1 CRD does not have a global schema
+		} else if typeMeta.APIVersion == apiextv1b1APIVersion {
+			groupKind, versions, err = unmarshalCRDV1beta1(rawContent)
+			if err != nil {
+				klog.Warningf("failed to unmarshal to CRD v1beta1: %v", err)
```
sttts: continue

pkg/schemapatcher/gen.go
```diff
@@ -330,24 +342,29 @@ func crdsFromDirectory(ctx *genall.GenerationContext, dir string) (map[schema.Gr
...
-			versions = map[string]struct{}{actualCRD.Spec.Version: struct{}{}}
-		} else {
-			versions = make(map[string]struct{}, len(actualCRD.Spec.Versions))
-			for _, ver := range actualCRD.Spec.Versions {
-				versions[ver.Name] = struct{}{}
+		// collect the group-kind and versions from the actual structured form
+		if typeMeta.APIVersion == apiextv1APIVersion {
+			groupKind, versions, err = unmarshalCRDV1(rawContent)
+			if err != nil {
+				klog.Warningf("failed to unmarshal to CRD v1: %v", err)
```
sttts: continue

marun: *sigh* done

sttts:
@DirectXMan12 can we merge this?

DirectXMan12:
Heh, I completely missed this before I wrote #356.  I'll take a look at this tomorrow, and hopefully we can unify.

sttts:
> Heh, I completely missed this before I wrote #356

This moment you spent a day writing something and then check open pull requests :) But I guess you support also crd-gen v1 output.

marun:
Superseded by #356

PR #322, opened 2019-09-19
:bug: promote version-specific CRD fields when only one version

joelanford:
This fixes an issue where version-specific fields are not properly promoted to their top-level `.spec` equivalents when there is only one version defined in `.spec.versions`.

### DIFF
--- a/pkg/crd/crd_spec_test.go
+++ b/pkg/crd/crd_spec_test.go
@@ -27,6 +27,36 @@ import (
 var _ = Describe("CRD Generation", func() {
 	Describe("Utilities", func() {
 		Describe("MergeIdenticalVersionInfo", func() {
+			It("should replace per-version schemata with a top-level schema if only one version", func() {
+				spec := &apiext.CustomResourceDefinition{
+					Spec: apiext.CustomResourceDefinitionSpec{
+						Versions: []apiext.CustomResourceDefinitionVersion{
+							{
+								Name:    "v1",
+								Storage: true,
+								Schema: &apiext.CustomResourceValidation{
+									OpenAPIV3Schema: &apiext.JSONSchemaProps{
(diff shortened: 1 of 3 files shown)


DirectXMan12:
This seems to check out with what we were seeing earlier on slack.  Can you test against a couple different k8s versions to check that it's compatible with them?

joelanford:
@DirectXMan12 with `controller-gen` built from this branch, I confirmed that the CRD generated from an API with just a single API version with `trivialVersions=false`:
1. Correctly promotes the version-specific `subresources`, `schema`, and `additionalPrinterColumns` to the respective top-level fields.
2. Is correctly applied on Kubernetes clusters `1.13.9`, `1.14.6`, and `1.15.2`

On all versions, I can:
1. See all of the top-level fields with `kubectl describe crd apps.testgroup.my.domain`
2. See printer columns with `kubectl get apps.testgroup.my.domain app-sample`
3. Get validation errors when applying invalid resources (and as expected with structural schemas, I noticed stricter validation in 1.15)

prydonius:
Can confirm this works with my 1.15 and 1.14 clusters. Thanks @joelanford!

DirectXMan12:
Awesome, thanks!

/lgtm
/approve

pires:
This doesn't work if `crdVersions=v1` is set even if `trivialVersions=false`.

DirectXMan12:
What issues are you seeing?  v1 *only* has per-version schemata.

pires:
Ah, if that's the case then `additionalPrinterColumns` will always show up per version, right? If that's the case, then I misunderstood and I have no issues.

DirectXMan12:
yes, it should always show up per version in v1

pires:
Thank you!

PR #323, opened 2019-09-20
:sparkles: Add a marker to support field defaulting in crd schema

marun:
**NOTE:** This PR is not ready for review until #329 merges.

Kube 1.16 adds beta support [for providing default values for fields](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting) in a CRD schema. This PR adds a marker that supports specifying defaults for all supported types.
 
cc: @sttts

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -72,6 +72,9 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("nullable", markers.DescribesField, Nullable{})).
 		WithHelp(Nullable{}.Help()),
+
+	must(markers.MakeAnyTypeDefinition("kubebuilder:validation:Default", markers.DescribesField, Default{})).
+		WithHelp(Default{}.Help()),
 }
 
 func init() {
@@ -158,6 +161,19 @@ type Type string
 //
 // This is often not necessary, but may be helpful with custom serialization.
 type Nullable struct{}
+
+// +controllertools:marker:generateHelp:category="CRD validation"
+// Default sets the default value for this field.
+//
+// A default value will be accepted as any value valid for the
+// field. Formatting for common types include: boolean: `true`, string:
+// `Cluster`, numerical: `1.24`, array: `{1,2}`, object: `{policy:
+// "delete"}`). Defaults should be defined in pruned form, and only best-effort
+// validation will be performed. Full validation of a default requires
+// submission of the containing CRD to an apiserver.
+type Default struct {
+	Value interface{}
+}
 
 func (m Maximum) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
 	if schema.Type != "integer" {
@@ -282,5 +298,15 @@ func (m Type) ApplyFirst() {}
 
 func (m Nullable) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
 	schema.Nullable = true
+	return nil
+}
+
+// Defaults are only valid CRDs created with the v1 API
+func (m Default) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
+	marshalledDefault, err := json.Marshal(m.Value)
+	if err != nil {
+		return err
+	}
+	schema.Default = &v1beta1.JSON{Raw: marshalledDefault}
 	return nil
 }
--- a/pkg/crd/markers/zz_generated.markerhelp.go
+++ b/pkg/crd/markers/zz_generated.markerhelp.go
@@ -24,6 +24,22 @@ import (
 	"sigs.k8s.io/controller-tools/pkg/markers"
 )
 
+func (Default) Help() *markers.DefinitionHelp {
+	return &markers.DefinitionHelp{
+		Category: "CRD validation",
+		DetailedHelp: markers.DetailedHelp{
+			Summary: "sets the default value for this field. ",
+			Details: " A default value will be accepted as any value valid for the field. Formatting for common types include: boolean: `true`, string: `Cluster`, numerical: `1.24`, array: `{1,2}`, object: `{policy: \"delete\"}`). Defaults should be defined in pruned form, and only best-effort validation will be performed. Full validation of a default requires submission of the containing CRD to an apiserver.",
+		},
+		FieldHelp: map[string]markers.DetailedHelp{
+			"Value": markers.DetailedHelp{
+				Summary: "",
+				Details: "",
+			},
+		},
+	}
+}
+
 func (Enum) Help() *markers.DefinitionHelp {
 	return &markers.DefinitionHelp{
 		Category: "CRD validation",
--- a/pkg/crd/markers/doc.go
+++ b/pkg/crd/markers/doc.go
@@ -21,7 +21,7 @@ limitations under the License.
 //
 // Validation Markers
 //
-// Validation markers have values that imprlement ApplyToSchema
+// Validation markers have values that implement ApplyToSchema
 // (crd.SchemaMarker).  Any marker implementing this will automatically
 // be run after the rest of a given schema node has been generated.
 // Markers that need to be run before any other markers can also
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -93,6 +93,27 @@ type CronJobSpec struct {
 
 	// This tests that markers that are allowed on both fields and types are applied to types
 	TwoOfAKindPart1 LongerString `json:"twoOfAKindPart1"`
+
+	// This tests that primitive defaulting can be performed.
+	// +kubebuilder:validation:Default=forty-two
+	DefaultedString string `json:"defaultedString"`
+
+	// This tests that slice defaulting can be performed.
+	// +kubebuilder:validation:Default={a,b}
+	DefaultedSlice []string `json:"defaultedSlice"`
+
+	// This tests that object defaulting can be performed.
+	// +kubebuilder:validation:Default={{nested: {foo: "baz", bar: true}},{nested: {bar: false}}}
+	DefaultedObject []RootObject `json:"defaultedObject"`
+}
+
+type NestedObject struct {
+	Foo string `json:"foo"`
+	Bar bool   `json:"bar"`
+}
+
+type RootObject struct {
+	Nested NestedObject `json:"nested"`
 }
 
 // +kubebuilder:validation:MinLength=4
--- a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
+++ b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
@@ -59,6 +59,42 @@ spec:
               - Forbid
               - Replace
               type: string
+            defaultedObject:
+              default:
+              - nested:
+                  bar: true
+                  foo: baz
+              - nested:
+                  bar: false
+              description: This tests that object defaulting can be performed.
+              items:
+                properties:
+                  nested:
+                    properties:
+                      bar:
+                        type: boolean
+                      foo:
+                        type: string
+                    required:
+                    - bar
+                    - foo
+                    type: object
+                required:
+                - nested
+                type: object
+              type: array
+            defaultedSlice:
+              default:
+              - a
+              - b
+              description: This tests that slice defaulting can be performed.
+              items:
+                type: string
+              type: array
+            defaultedString:
+              default: forty-two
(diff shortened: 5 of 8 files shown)


sttts:
Also for arbitrary `x-kubernetes-preserve-unknown-fields` fields.

DirectXMan12:
we'll need to extend the marker syntax a bit for arbitrary objects, but it should be doable

DirectXMan12:
arrays of primitives or other arrays should be doale today

DirectXMan12:
Ah, nvm.  I see, you're just using strings.  I think we might want to use structured parsing (you can use the "any" type here) so you can get parsing errors when generating config.

marun:
@DirectXMan12 

> Ah, nvm. I see, you're just using strings. I think we might want to use structured parsing (you can use the "any" type here) so you can get parsing errors when generating config.

I've updated the code to validate the default according to the provided schema rather than manually validating based on the field type. This supports all field types and errors will result if the supplied value doesn't match the schema.  Is that sufficient?

> we'll need to extend the marker syntax a bit for arbitrary objects, but it should be doable

Assuming continuing to use strings is acceptable, what are your thoughts on allowing json-formatted slices (`[1,2]`) and objects (`{"foo": 1, "bar": 2}`) to be specified? This would require allowing commas to appear in strings when enclosed in square or curly braces instead of having a comma appearing outside a quoted string be treated as an argument separator.

marun:
Latest iteration includes preliminary support for json-formatted strings. The integration test is failing due a default object value not being serialized to yaml in a stable order.

marun:
Ready for review

pkg/crd/markers/validation.go
```diff
@@ -284,3 +294,54 @@ func (m Nullable) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
 	schema.Nullable = true
 	return nil
 }
+
+// TODO(marun) Defaults should only be generated for v1 crds
```
sttts: and rejected or ignored for v1beta1? Tending to the former.

pkg/crd/markers/validation.go
```diff
@@ -159,6 +161,14 @@ type Type string
 // This is often not necessary, but may be helpful with custom serialization.
 type Nullable struct{}
 
+// +controllertools:marker:generateHelp:category="CRD validation"
+// Default sets the default value for this field.
+//
+// A default value will be accepted as a json-formatted string (e.g. `true`,
+// `"Cluster"`, `1.24`, `[1,2]`, `{"policy": "delete"}`) and validated as the
+// type indicated for the field.
+type Default string
```
sttts: as we discussed, would `type Default struct { interface{} }` work?

pkg/crd/markers/validation.go
```diff
@@ -284,3 +294,55 @@ func (m Nullable) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
...
 }
+
+// Defaults are only valid CRDs created with the v1 API
+func (m Default) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
+	// Convert versioned schema to unversioned form
+	marshalledSchema, err := json.Marshal(schema)
+	if err != nil {
+		return err
+	}
+	unversionedSchema := &apiext.JSONSchemaProps{}
```
sttts: the internal types are not meant for public consumption and might become inaccessible to this code base any time in the future. Why do we need that?

pkg/crd/markers/validation.go
```diff
@@ -284,3 +294,55 @@ func (m Nullable) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
...
+	}
+
+	// Initialize validator for schema
+	crValidation := &apiext.CustomResourceValidation{
+		OpenAPIV3Schema: unversionedSchema,
+	}
+	validator, _, err := validation.NewSchemaValidator(crValidation)
+	if err != nil {
+		return err
+	}
```
sttts: if you really want to validate the value here, you can just unmarshal the schema bytes into a go-openapi spec and call go-openapi's validation directly, without accessing apiextensions-apiserver internals.

pkg/crd/markers/validation.go
```diff
@@ -284,3 +294,55 @@ func (m Nullable) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
...
+	err = json.Unmarshal([]byte(value), &obj)
+	if err != nil {
+		return err
+	}
+
+	// Validate the default with the field's schema.
+	result := validator.Validate(obj)
+	if len(result.Errors) > 0 {
+		return fmt.Errorf("default fails validation: %v", result.Errors)
+	}
```
sttts: note that this is by far not complete. This lacks 
- the apiextensions/v1 validation
- a check for structurality of the provided schema
- a check that we actually have a v1 CRD
- validation that the value is pruned
- logic to cope with `x-kubernetes-preserve-unknown-fields`
- logic to handle defaults under ObjectMeta correctly.

sttts: We don't want to and cannot do this validation here.

pkg/crd/markers/validation.go
marun: That is not a valid golang type definition. It would be necessary to define the empty interface as a named field rather than embedding it, requiring either that the field name be provided (e.g. `Default=Value=<value>`) or that the parser be special-cased to not require the field name. `AnyType` parsing currently lacks support for object parsing (e.g. `{"foo": "bar"}`), and I'm not sure it's worth implementing for defaulting since the raw argument can be trivially json unmarshalled.

Instead, I defined `Default` as `markers.RawArguments` (`[]byte`) and revised marker construction to enable the use of `RawType` parsing. The previous strategy to detect when a field should be parsed as `RawType` relied upon comparing the field type to `RawArguments`. This works for a nested type (e.g. slice element or struct field) since the type can be identical, but a user-defined type fails that check (e.g. `reflect.TypeOf(Default) != reflect.TypeOf(RawArguments)`).

pkg/crd/markers/validation.go
marun: I didn't realize these exported methods were intended to be internal. How do I determine whether any given functionality provided by a dependency is fair game for use?

marun: *sigh* I guess the use of unversioned API types would be my first clue.

pkg/crd/markers/validation.go
marun: No good reason, just conforming to what was required by the methods I was trying to call.

pkg/crd/markers/validation.go
marun: I'm going to treat this as an educational opportunity. :)  

> note that this is by far not complete. This lacks
> 
> * the apiextensions/v1 validation

What is this and how is it relevant to default validation?

> * a check for structurality of the provided schema

This method is validating a default value, not a schema. Am I missing something?

> * a check that we actually have a v1 CRD

As previously discussed, the current architecture of schema generation will likely require removal of generated defaults via post-processing.

> * validation that the value is pruned

Is this something I should just ignore then?

> * logic to cope with `x-kubernetes-preserve-unknown-fields`

Hmm, I was optimistically hoping that the validation method I was calling would be considering this extension.  So the reason my test is passing is the lack of a field type and the absence of pruning? 

> * logic to handle defaults under ObjectMeta correctly.

Special-casing presumably - not all fields can be defaulted?

pkg/crd/markers/validation.go
sttts: They might go away any time. If we finally move internal packages to `internal`, there won't be a replacement.

pkg/crd/markers/validation.go
sttts: > I'm going to treat this as an educational opportunity. :)
> > note that this is by far not complete. This lacks
> > 
> > * the apiextensions/v1 validation
> 
> What is this and how is it relevant to default validation?

pkg/apiextensions/validation

It determines whether a default is allowed at all. It forbid certain schema fields although they show up in the schema types. For example we forbid defaults for fields in metadata.

> > > * a check for structurality of the provided schema
> 
> This method is validating a default value, not a schema. Am I missing something?

Only structural schemas allow defaults. But I agree: if you do the structural check elsewhere, we can probbaly assume it here.

> > > * a check that we actually have a v1 CRD
> As previously discussed, the current architecture of schema generation will likely require removal of generated defaults via post-processing.
> 
> > * validation that the value is pruned
> > Is this something I should just ignore then?

I think you can only ignore this requirement with a reasonable amount of code. It is tricky in combination with metadata and x-kubernetes-embedded-object. In other words, we have to be aware that the validation here is incomplete. This can be fine.

> > > * logic to cope with `x-kubernetes-preserve-unknown-fields`
> > Hmm, I was optimistically hoping that the validation method I was calling would be considering this extension. So the reason my test is passing is the lack of a field type and the absence of pruning?

I meant the pruning check here. The validation you call does not check pruning.

> 
> > * logic to handle defaults under ObjectMeta correctly.
> 
> Special-casing presumably - not all fields can be defaulted?

`<-- snip -->`

tl/dr: we can call validation via go-openapi here. That's fine and might catch errors from users. But it will be incomplete. I.e. you will be able to define APIs with markers which will result in CRD manifests which are rejected. Be clear about that in a Go comment at the top that this is best-effort.

After accepting that it is incomplete, we have to make sure the validation is correct, i.e. that you don't forbid defaults which would be valid in a CRD. I believe this is the case as we don't drop anything from the CRD schema, but only add more restriction through the `x-kubernetes-*` fields.

marun:
@sttts Rebased on top of #329, which required a change in format for how defaults are specified in the marker (no longer strict json).  I'm working on validating defaults with the schema post-flattening rather than at time of application.

pkg/crd/markers/validation.go
sttts: I am fine with this now.

pkg/crd/markers/zz_generated.markerhelp.go
```diff
@@ -24,6 +24,22 @@ import (
 	"sigs.k8s.io/controller-tools/pkg/markers"
 )
 
+func (Default) Help() *markers.DefinitionHelp {
+	return &markers.DefinitionHelp{
+		Category: "CRD validation",
+		DetailedHelp: markers.DetailedHelp{
+			Summary: "sets the default value for this field. ",
+			Details: " A default value will be accepted as any valid default value (e.g. boolean: `true`, string: `Cluster`, numerical: `1.24`, array: `{1,2}`, object: `{policy: \"delete\"}`) and validated as the type indicated for the field.",
```
sttts: leading space?

sttts: Maybe document that this must be pruned and validate against the schema.

marun: The leading space seems to be common to all generated help. @DirectXMan12 is this a bug?

marun: > Maybe document that this must be pruned and validate against the schema.

Done

marun:
Ready for review

pkg/crd/markers/validation.go
```diff
@@ -159,6 +163,13 @@ type Type string
 // This is often not necessary, but may be helpful with custom serialization.
 type Nullable struct{}
 
+// +controllertools:marker:generateHelp:category="CRD validation"
+// Default sets the default value for this field.
+//
+// A default value will be accepted as a string and validated as the type
+// indicated for the field.
+type Default string
```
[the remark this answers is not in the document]
marun: I'm guessing this is an old comment?

marun:
> Can you update the marker syntax in the kubebulider book's reference section to mention the map syntax?

Will do!

pkg/crd/markers/validation.go
```diff
@@ -72,6 +72,9 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("nullable", markers.DescribesField, Nullable{})).
 		WithHelp(Nullable{}.Help()),
+
+	must(markers.MakeAnyTypeDefinition("kubebuilder:validation:Default", markers.DescribesField, Default{})).
```
DirectXMan12: @marun before we release a release with this, do we want `:validation` in the middle, since this isn't actually validation (even though it's in the same place as validation)?  Not sure if that's confusing

marun: I'll remove.

marun: PR: https://github.com/kubernetes-sigs/controller-tools/pull/337

Issue #291, opened 2019-07-31
controller-gen doesn't work with k8s.io/apiextensions-apiserver


jiachengxu:
Hi folks,

I use controller-gen integrated with Kubebuilder to generate RBAC, webhook and CRDs for my project, there is a command called `make manifests` to run:
 `controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases`   
to generate everything, but it doesn't work for me since I have `k8s.io/apiextensions-apiserver` in my `go.mod` as a dependency, the following error messages are poping out:
`/Users/maxwell/go/pkg/mod/k8s.io/apiextensions-apiserver@v0.0.0-20190409022649-727a075fdec8/pkg/apis/apiextensions/v1beta1/types_jsonschema.go:63:2: enountered struct field "Raw" without JSON tag in type "JSON"
/Users/maxwell/go/pkg/mod/k8s.io/apiextensions-apiserver@v0.0.0-20190409022649-727a075fdec8/pkg/apis/apiextensions/v1beta1/types_jsonschema.go:29:24: unsupported type "float64"
/Users/maxwell/go/pkg/mod/k8s.io/apiextensions-apiserver@v0.0.0-20190409022649-727a075fdec8/pkg/apis/apiextensions/v1beta1/types_jsonschema.go:31:24: unsupported type "float64"
/Users/maxwell/go/pkg/mod/k8s.io/apiextensions-apiserver@v0.0.0-20190409022649-727a075fdec8/pkg/apis/apiextensions/v1beta1/types_jsonschema.go:39:24: unsupported type "float64"
/Users/maxwell/go/pkg/mod/k8s.io/apiextensions-apiserver@v0.0.0-20190409022649-727a075fdec8/pkg/apis/apiextensions/v1beta1/types_jsonschema.go:85:2: enountered struct field "Schema" without JSON tag in type "JSONSchemaPropsOrArray"
/Users/maxwell/go/pkg/mod/k8s.io/apiextensions-apiserver@v0.0.0-20190409022649-727a075fdec8/pkg/apis/apiextensions/v1beta1/types_jsonschema.go:86:2: enountered struct field "JSONSchemas" without JSON tag in type "JSONSchemaPropsOrArray"
[...]

PR #326, opened 2019-09-23
🐛Add apiextension JSONSchemaPropsOrArray, JSONSchemaPropsOrBool, JSONSchemaOrStringArray to known_types.go

jiachengxu:
Fix: #291 
Ref: #277 
cc @thetechnick
cc @munnerz @sttts @DirectXMan12 

This PR adds the `JSONSchemaPropsOrArray`, `JSONSchemaPropsOrBool`, `JSONSchemaOrStringArray` types to known_types.go, but we haven't figured out the `Type` for the `JSONSchemaProps`, and we are keeping it to `object` in this PR, but if it is not correct, we can still change it.

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -84,6 +84,42 @@ var KnownPackages = map[string]PackageOverride{
 		}
 		// No point in calling AddPackage, this is the sole inhabitant
 	},
+
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1": func(p *Parser, pkg *loader.Package) {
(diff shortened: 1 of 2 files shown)


sttts:
I don’t think kubebuilder should have special support for these types. It should rather have the mechanisms to handle these correctly with some user provided markers if necessary. JSONSchemaProps and friends are not fundamental types like RawExtensions or runtime.Object.

DirectXMan12:
agree.  We mostly have overrides for types that you can't apply markers to and that are fairly common.  It's not clear to me what the usecase is here...

PR #329, opened 2019-09-27
support maps in marker marker parsing

DirectXMan12:
Maps in markers always have string keys, and may have any argument type, including
nested maps, slices, etc.
"Guessed" maps are always non-uniform (i.e. key of AnyType), since
they're frequently used to represent objects in JSON form.

They syntax looks like:

```
//+this:is:a:map={"text": "abc", len: 33}
```


DirectXMan12:
/hold

this is partial -- whether we merge this depends on how it works with #323

DirectXMan12:
There's one caveat: we'll never "guess" an empty map to be a map.  I *think* this is okay for now, but if anyone sees a way around it lmk.

marun:
I've rebased #323 on this PR and it works like a charm.  LGTM

Issue #259, opened 2019-07-17
Add a version command or option to `controller-gen`


alexkreidler:
It would be great to know which version of `controller-gen` we're using in our kubebuilder projects to better diagnose errors and help the community.

Please add a `version` command or `--version` or `-v` flags to controller-gen to help us with this. Thanks a ton!

jiachengxu:
Yes, this would be helpful, since I encountered a similar issue with kubebuilder :https://github.com/kubernetes-sigs/kubebuilder/issues/909. I could try to work on this if help is needed.

DirectXMan12:
/kind feature
/help-wanted

You can more-or-less copy the ldflags/git approach from other projects or bits of kubernetes.  If you've done something similar before, or are willing to do a bit of bash/makefile spelunking, this is a reasonable starter issue.

DirectXMan12:
(we'd probably add this as a `--version` flag)

jiachengxu:
Hi @DirectXMan12, if help is needed, I am willing to work on this issue, any more hints on this?

DirectXMan12:
~~kustomize uses goreleaser, so it's not 1-to-1, but you can get the gist of the general technique here: https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/goreleaser.yaml#L11 / https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/commands/misc/version.go~~

~~Instead of goreleaser, we'd use a script to grab the version info.  The general trick is to run something like `git describe --tags --abbrev=14` and pull the version and commit out of that.
metrics-server uses a giant makefile for doing it, but has extra fancy footwork to stay compatible with what's done in kubernetes/kubernetes: https://github.com/kubernetes-incubator/metrics-server/blob/master/hack/Makefile.buildinfo.  I don't think we need to get that complicated, though.~~

On further thought, since we're generally go-installing, we probably need to make use of https://golang.org/pkg/runtime/debug/#BuildInfo to grab the version info at runtime.  Then we can print that out.

PR #332, opened 2019-10-01
:sparkles: Add a --version option to controller-gen

dlespiau:
It's useful to know the version of a `controller-gen` binary lying around. We could write checks in the kubebuilder `Makefile` to upgrade `controller-gen` when needed, or at least warn about it.

Fixes: #259 

I checked that this indeed worked with main modules when using `go get`: https://github.com/dlespiau/go-main-module-version.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,34 @@
+package version
+
+import (
+	"fmt"
+	"runtime/debug"
+)
+
+// version returns the version of the main module
+func version() string {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		// binary has not been built with module support
+		return "(unknown)"
+	}
+	return info.Main.Version
+}
+
+// Print prints the main module version on stdout.
+//
+// Print will display either:
+//
+// - "Version: v0.2.1" when the program has been compiled with:
+//
+//   $ go get github.com/controller-tools/cmd/controller-gen@v0.2.1
+//
+//   Note: go modules requires the usage of semver compatible tags starting with
+//        'v' to have nice human-readable versions.
+//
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -0,0 +1,26 @@
...
+		return "", errors.New("binary has not been built with module support")
+	}
+	return info.Main.Version, nil
+}
+
+// Print prints the main module version on stdout
+func Print() error {
+	v, err := version()
+	if err != nil {
+		return err
```
droot: Looks like nothing will get printed for the end-user. We should probably print `Version: Unknown` in this case.

droot:
Also, can you pl. update the commit and PR as per our template --> https://raw.githubusercontent.com/kubernetes-sigs/controller-tools/master/.github/pull_request_template.md

droot:
I took another look and feel that we can simplify the version function by not returning error. In case of error, we simply return "unknown" version.

dlespiau:
The error was propagated and displayed by the main program explaining why the version coud not be displayed (not compiled with go modules). But I see the point: may as well always display a "Version: .*" string that can be parsed in always the same way.

One detail is that the go runtime with return version `(devel)` when compiled from a local checkout. I chose to follow that lead and display `(unknown)` instead of the suggested `Unknown` so we have a stronger invariant. Also documented all the possible behaviours of `Print`.

droot:
>One detail is that the go runtime with return version (devel) when compiled from a local checkout. I chose to follow that lead and display (unknown) instead of the suggested Unknown so we have a stronger invariant. Also documented all the possible behaviours of Print.

Sounds good to me.

PR #337, opened 2019-10-05
:sparkles: s/kubebuilder:validation:Default/kubebuilder:Default/

marun:


pkg/crd/markers/validation.go
```diff
@@ -73,7 +73,7 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 	must(markers.MakeDefinition("nullable", markers.DescribesField, Nullable{})).
 		WithHelp(Nullable{}.Help()),
 
-	must(markers.MakeAnyTypeDefinition("kubebuilder:validation:Default", markers.DescribesField, Default{})).
+	must(markers.MakeAnyTypeDefinition("kubebuilder:Default", markers.DescribesField, Default{})).
```
droot: Had an offline chat with @DirectXMan12 about it. We should lowercase the `Default` to keep it consistent with `kubebuilder:validation` marker and eventually we can make these marker parsing case insensitive.

Otherwise, it looks good.

marun: Done

DirectXMan12:
/lgtm
/approve

awesome

PR #344, opened 2019-10-16
 ✨ Do not create empty webhook/manifests.yaml file

dsbrng25b:
Even if there are no webhooks an empty webhook/manifests.yaml file is created.
The PR first checks if there are webhooks and only then calls WriteYAML.

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -242,8 +242,8 @@ func (Generator) Generate(ctx *genall.GenerationContext) error {
 
 	}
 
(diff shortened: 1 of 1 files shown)


alexeldeib:
I noticed this today and was thinking the same change should be made :) 

/lgtm

DirectXMan12:
cc @mengqiy for final signoff

mengqiy:
There was a concern that if the `manifests.yaml` file doesn't exist, `config/webhook` is no longer a valid kustomize package. But if there is not webhook objects in `manifests.yaml`, the users may not use `config/webhook` at all.
So I guess this change is fine.

Issue #281, opened 2019-07-25
Marker for x-kubernetes-preserve-unknown-fields


hasheddan:
Similar to #278, a marker should be created to specify pruning behavior as was introduced in [v1.15](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#pruning-versus-preserving-unknown-fields). There would also need to be an integration test case for this marker.

Proposal for format of marker: `+kubebuilder:validation:PreserveUnknownFields`

DirectXMan12:
seems reasonable to me :-)

liggitt:
note that `spec.preserveUnknownFields: true` will not be allowed in v1, but must be specified in the schema itself, for example:
```yaml
versions:
- name: v1
  schema:
    type: "object"
    x-kubernetes-preserve-unknown-fields: true
- name: v2
  schema:
    type: "object"
    properties:
      myproperty:
        type: "object"
        x-kubernetes-preserve-unknown-fields: true
```

Also note that use of `x-kubernetes-preserve-unknown-fields: true` to avoid specifying any schema properties is discouraged, since it limits the amount of useful information available to clients. Use for specific subtrees that can contain arbitrary JSON is the intended use case.

DirectXMan12:
yep, the marker would set the x- value.

hasheddan:
Fixed by #345 

/close

PR #345, opened 2019-10-17
Add x-kubernetes-{embedded-resource,preserve-unknown-fields} support

sttts:
Marker syntax:
- `kubebuilder:pruning:PreserveUnknownFields`
- `kubebuilder:validation:EmbeddedResource`

Fixing https://github.com/kubernetes-sigs/controller-tools/issues/281

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -58,6 +58,8 @@ var ValidationMarkers = mustMakeAllWithPrefix("kubebuilder:validation", markers.
 	Enum(nil),
 	Format(""),
 	Type(""),
+	XPreserveUnknownFields{},
+	XEmbeddedResource{},
 )
 
 // FieldOnlyMarkers list field-specific validation markers (i.e. those markers that don't make
@@ -75,6 +77,11 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 
 	must(markers.MakeAnyTypeDefinition("kubebuilder:default", markers.DescribesField, Default{})).
 		WithHelp(Default{}.Help()),
+
+	must(markers.MakeDefinition("kubebuilder:pruning:PreserveUnknownFields", markers.DescribesField, XPreserveUnknownFields{})).
+		WithHelp(XPreserveUnknownFields{}.Help()),
+	must(markers.MakeDefinition("kubebuilder:validation:EmbeddedResource", markers.DescribesField, XEmbeddedResource{})).
+		WithHelp(XEmbeddedResource{}.Help()),
 }
 
 func init() {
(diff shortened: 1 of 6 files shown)


pkg/crd/markers/validation.go
```diff
@@ -310,3 +338,13 @@ func (m Default) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
 	schema.Default = &v1beta1.JSON{Raw: marshalledDefault}
 	return nil
 }
+
+func (m XPreserveUnknownFields) ApplyToSchema(schema *v1beta1.JSONSchemaProps) error {
+	schema.XPreserveUnknownFields = pointer.BoolPtr(true)
```
DirectXMan12:
```suggestion
        defTrue := true
	schema.XPreserveUnknownFields = &defTrue
```

This is the silliest library, and I'd rather not have to deal with something else in it conflicting with a dependency.

pkg/crd/markers/validation.go
```diff
@@ -175,6 +183,26 @@ type Default struct {
 	Value interface{}
 }
 
+// +controllertools:marker:generateHelp:category="CRD pruning"
+// PreserveUnknownFields stops the apiserver from pruning fields which are not specified.
+//
+// By default the apiserver drops unknown fields from the request payload
+// during the decoding step.stops the API server. This marker stop the API
```
DirectXMan12: `during the decoding step.stop the API server` looks like a copy paste typo from the next sentence

sttts: fixed

sttts:
@DirectXMan12 ptal

pkg/crd/markers/validation.go
```diff
@@ -175,6 +182,26 @@ type Default struct {
 	Value interface{}
 }
 
+// +controllertools:marker:generateHelp:category="CRD pruning"
```
DirectXMan12: nit: can you stick this in the validation category?  it just makes it less weird to show in the docs (or we should come up with a general category for non-validation schema things, along with `listType` and such).  cc @apelisse

apelisse: I've used "CRD topology", wdyt?

DirectXMan12: SGTM

DirectXMan12: (or "CRD processing" or "CRD structure" -- whatever you think gets the point across)

apelisse: I thought about structure too. processing is super "imperative", it's such a bad word ;-)

sttts: `CRD processing` fits best. It's not about structure, but about the way we process requests.

sttts: updated

DirectXMan12:
rebased,
re-applying lgtm

Issue #349, opened 2019-10-23
Empty validation field when generating CRDs with multiple versions


sebgl:
I think there's a bug which prevents the trivial versions option to work as expected.

How to reproduce:

* setup an `apis/` pkg using a kubebuilder directory structure, containing multiple versions of the same CRD, with a different spec:
```
* apis/mycrd/v1alpha1
* apis/mycrd/v1beta1
```
* generate the CRD manifests using `crd:trivialVersions=true`
* according to the code, the CRD top-level `validation` field should be filled with the schema of the version which has `storage: true`: 

https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/gen.go#L98-L116

* the resulting CRD has an empty `validation` field, which should not happen, I think?

Unless we explicitly want to have no `validation` set when multiple versions are specified? (Which currently happens, I think, as an undesired side-effect in the code).

sebgl:
Looking at the code, the `validation` and `schema` fields are mutated in `toTrivialVersions()`.
I think we process each individual groupKind as many times as there are corresponding versions, but since we end up mutating the same CRD multiple times, its validation field becomes `nil` starting the second iteration.

sebgl:
/remove-lifecycle stale

I think this is still an issue. https://github.com/kubernetes-sigs/controller-tools/pull/350 is pending review.  https://github.com/kubernetes-sigs/controller-tools/pull/365#discussion_r350404541 adds integration tests to catch this bug.

PR #350, opened 2019-10-23
:bug: Fix trivialVersions validation by processing each groupKind only once

sebgl:
Fixes https://github.com/kubernetes-sigs/controller-tools/issues/349.

FindKubeKinds may return duplicates if there are several versions of the
same groupKind for which we want to generate CRDs.
When combined with the trvialVersions option, which does mutate a
pointer of the corresponding object, we end up overriding the
top-level `validation` field with `nil`, resulting from the already
overridden schema of another version of the same GroupKind.

This commit fixes it by making sure FindKubeKinds does not return
duplicate entries.

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -145,6 +145,8 @@ func FindMetav1(roots []*loader.Package) *loader.Package {
 func FindKubeKinds(parser *Parser, metav1Pkg *loader.Package) []schema.GroupKind {
 	// TODO(directxman12): technically, we should be finding metav1 per-package
 	var kubeKinds []schema.GroupKind
+	// there might be duplicates if some GroupKinds are present in multiple versions
+	found := map[schema.GroupKind]struct{}{}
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Can you add a test for this, if possible?

sebgl:
@DirectXMan12 I'll try to add a test 👍 
(this week is busy, I'll look at it next week)

sebgl:
I'll create a separate PR (for easier review) to add integration tests for CRD generation based on the cronjob example.
Then add a trivial version test covering this PR.

sebgl:
I plan to rebase this PR on https://github.com/kubernetes-sigs/controller-tools/pull/365 once https://github.com/kubernetes-sigs/controller-tools/pull/365 is merged. The rebase will [highlight differences in the integration tests YAML manifests](https://github.com/kubernetes-sigs/controller-tools/pull/365#discussion_r350404541).

PR #351, opened 2019-10-26
:bug: controller-gen panics when encountering an import loop

11xor6:
This fixed an issue where `controller-gen` panics when encountering an import loop. This is especially problematic because when running `make` as part of the default Kubebuilder build `controller-gen` runs before the compiler, thus an import loop can be hard to detect.

### DIFF
--- a/pkg/loader/loader.go
+++ b/pkg/loader/loader.go
@@ -243,6 +243,9 @@ func (l *loader) typeCheck(pkg *Package) {
 
 		// The imports map is keyed by import path.
 		importedPkg := pkg.Imports()[path]
+		if importedPkg == nil {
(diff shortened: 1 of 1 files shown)


DirectXMan12:
thanks for this!  Can you fix the commit user?  Looks like the CLA bot can't link it to your GH account.

11xor6:
I've fixed the commit author however I'm still waiting on my employer for the CLA. It's in the works, but no ETA on when.

DirectXMan12:
alright, just let us know :-)

/lifecycle frozen

^ so that the bot doesn't get cranky

11xor6:
CLA should be resolved.

DirectXMan12:
/lgtm
/approve

awesome!

PR #353, opened 2019-11-01
Extend webhook marker documentation

anyasabo:
Extends the marker documentation for webhook generation for `name` and `path`.

The OpenAPI specs for webhooks require the name is a domain name.

The path is handled specially by controller-runtime, so this adds documentation on how controller-runtime handles it here:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/webhook.go#L158

Otherwise the webhook configuration generated by controller-tools will not match what actually runs with controller-runtime

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -70,9 +70,15 @@ type Config struct {
 	// Versions specifies the API versions that this webhook receives requests for.
 	Versions []string
 
-	// Name indicates the name of this webhook configuration.
+	// Name indicates the name of this webhook configuration. Should be a domain with at least three segments separated by dots
 	Name string
-	// Path specifies that path that the API server should connect to this webhook on.
+
+	// Path specifies that path that the API server should connect to this webhook on. Must be
+	// prefixed with a '/validate-' or '/mutate-' depending on the type, and followed by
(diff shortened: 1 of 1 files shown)


pkg/webhook/parser.go
```diff
@@ -70,9 +70,15 @@ type Config struct {
 	// Versions specifies the API versions that this webhook receives requests for.
 	Versions []string
 
-	// Name indicates the name of this webhook configuration.
+	// Name indicates the name of this webhook configuration. Should be a domain with at least three segments separated by dots
```
mengqiy: > Should be a domain with at least three segments separated by dots

Probably just say:
`Name should be fully qualified, e.g. imagepolicy.kubebuilder.io`.

anyasabo: An [FQDN](https://tools.ietf.org/html/rfc1594#section-5) can also simply be `kubebuilder.io`, so this is more specific about the requirements I think. This is simply using the wording direct from the webhook validation. This is the error you receive if you try to apply a webhook with a name with < 3 labels:

>The ValidatingWebhookConfiguration "validating-webhook-configuration" is invalid: webhooks[0].name: Invalid value: "funfunfun": should be a domain with at least three segments separated by dots

PR #354, opened 2019-11-02
Check pointers before dereferencing

sttts:
On errors in schema generation these would lead to panics.

### DIFF
--- a/pkg/crd/parser.go
+++ b/pkg/crd/parser.go
@@ -187,7 +187,9 @@ func (p *Parser) NeedFlattenedSchemaFor(typ TypeIdent) {
 	partialFlattened := p.flattener.FlattenType(typ)
 	fullyFlattened := FlattenEmbedded(partialFlattened, typ.Package)
 
-	p.FlattenedSchemata[typ] = *fullyFlattened
+	if fullyFlattened != nil {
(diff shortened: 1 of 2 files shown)


DirectXMan12:
/hold

I'd like to discuss this a bit -- I'd consider these to points to be invariant, and if the invariant is violated, I kinda want a panic here.  Is there a case where these are actually *supposed* to be nil?

sttts:
I managed to get them nil somehow and spent long time to fix it. I think it was some wrong marker. A panic is not the right user experience. 

A general complaint during my experiments: error output is often super bad. I had a number of cases where line number and/or file name were missing and I had to guess what was wrong where. Also log output would help a lot, e.g. to understand why files are not processed during schema patching.

DirectXMan12:
> error output is often super bad

it's been a TODO of mine to try and figure out out to track original line number once we lose line source span information.  If you have any good ideas, I'm all ears -- we track it up to a certain point, then it gets lost when we construct the schema.  Having some way to map schema objects back to their source code spans that produced them would be useful error wise.

> error output is often super bad

:+1: agreed.  Can you file issues for both of them?  Log output is prob good-first-issue, at least.

PR #356, opened 2019-11-05
:sparkles: CRD v1 support

### LINKED ISSUE #352: CRD v1 support
We should support (optionally) generating apiextensions/v1 CRDs.  This adds support for defaulting + a handful of other things.  We'll need to keep v1beta1 support around for a while.

DirectXMan12:
This introduces CRD v1 support, as well as support for generating CRDs in multiple versions of the CRD type itself (`apiextensions/v1` and `apiextensions/v1beta1` currently).

crd-gen and schemapatcher both have support for the new format.

The default in crd-gen is now to generate both v1beta1 and v1, with the first specified version (or v1beta1 if default) getting an unsuffixed file name, and the rest getting something that looks like `<old-name>.<crd-version>.yaml`.

This also contains a couple of utility updates, such as fixing up some helpers for regenerating testdata and some missed generator help generation.

Fixes #352


DirectXMan12:
cc @sttts @droot

Gopkg.lock
```diff
@@ -1,382 +0,0 @@
-# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
```
sttts: finally 🎉

pkg/schemapatcher/gen.go
```diff
@@ -125,92 +137,149 @@ func (g Generator) Generate(ctx *genall.GenerationContext) (result error) {
...
-				return fmt.Errorf("failed to set global firstSchema for %s: %v", existingInfo.GroupKind, err)
+			if err := existingSet.setGlobalSchema(); err != nil {
+				return fmt.Errorf("failed to set global firstSchema for %s: %v", existingSet.GroupKind, err)
 			}
 		} else {
-			if err := existingInfo.setVersionedSchemata(); err != nil {
-				return fmt.Errorf("failed to set versioned schemas for %s: %v", existingInfo.GroupKind, err)
+			if err := existingSet.setVersionedSchemata(); err != nil {
+				return fmt.Errorf("failed to set versioned schemas for %s: %v", existingSet.GroupKind, err)
 			}
```
sttts: the top-level field does not exist anymore in v1. You always set schemas, subresources etc. in the versions.

DirectXMan12: yep.  I can probably make it clearer, but if you look at the source for `setGlobalSchema`, it only does so if it's looking at a v1beta1 CRD, and delegates back to `setVersionedSchemata` for the v1 CRD.  We might want to rename the methods...

pkg/crd/gen.go
```diff
@@ -76,16 +89,45 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		return nil
 	}
 
+	crdVersions := g.CRDVersions
+
+	if len(crdVersions) == 0 {
+		// put v1beta1 first to make it the "default" file for compatibility purposes
+		crdVersions = []string{"v1beta1", "v1"}
```
droot: Not sure if this requires major version bump for controller-tools. In a way, it changes the default behavior. I am fine with bumping up `0.3.0`.

pkg/crd/gen.go
```diff
@@ -76,16 +89,45 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
+
 	for _, groupKind := range kubeKinds {
 		parser.NeedCRDFor(groupKind, g.MaxDescLen)
-		crd := parser.CustomResourceDefinitions[groupKind]
+		crdRaw := parser.CustomResourceDefinitions[groupKind]
+		addAttribution(&crdRaw)
+
+		versionedCRDs := make([]interface{}, len(crdVersions))
+		for i, ver := range crdVersions {
+			conv, err := AsVersion(crdRaw, schema.GroupVersion{Group: apiext.SchemeGroupVersion.Group, Version: ver})
```
droot: Summarizing our offline conversation --> so if I am using v1 specific feature (say defaulting), generated v1beta1 CRD will have defaulting specific features enabled, but when submitted to API-Server will result an error.
So current KB projects will run in to this issue when they start using v1 features. Still wondering if there is a better way to handle this transition.

pkg/crd/gen.go
DirectXMan12: I don't think it's a problem because v1beta1 is still in the same named file.

DirectXMan12: we can just not generate v1 at all by default as well

DirectXMan12: let's just do that and eventually change to make this the default

pkg/schemapatcher/gen.go
sttts: yep, looks good

PR #361, opened 2019-11-18
[WIP] :bug: construct CRD from v1beta1

mengqiy:
Top level PreserveUnknownFields is a pointer in v1beta1 contains more information than v1.

Will add tests to cover both v1 and v1beta1.


mengqiy:
Closing this PR since it's not the best way to move forward.

PR #362, opened 2019-11-19
:bug: preserveUnknownFields: true on v1beta1, option

DirectXMan12:
To avoid a breaking change in crd-gen, we need to default to
`preserveUnknownFields: true` if you're just generating v1beta1 CRDs.

In all other cases (multiple versions, v1 only), we default to
`preserveUnknownFields: false`, as this is the reccomended default and
required by v1.

We also introduce an option to crd-gen to allow overriding this
(`crd:preserveUnknownFields`).

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -44,6 +44,15 @@ type Generator struct {
 	// Only works with the v1beta1 CRD version.
 	TrivialVersions bool `marker:",optional"`
 
+	// PreserveUnknownFields indicates whether or not we should turn off pruning.
+	//
+	// Left unspecified, it'll default to true when only a v1beta1 CRD is
+	// generated (to preserve compatibility with older versions of this tool),
+	// or false otherwise.
+	//
+	// It's required to be false for v1 CRDs.
+	PreserveUnknownFields *bool `marker:",optional"`
+
 	// MaxDescLen specifies the maximum description length for fields in CRD's OpenAPI schema.
 	//
 	// 0 indicates drop the description for all fields completely.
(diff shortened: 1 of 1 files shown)


pkg/crd/gen.go
```diff
@@ -117,6 +126,20 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
+		// the sensible default and is required for v1.
+		v1beta1Only := len(crdVersions) == 1 && crdVersions[0] == "v1beta1"
+		switch {
+		case (g.PreserveUnknownFields == nil || *g.PreserveUnknownFields) && v1beta1Only:
+			crd := versionedCRDs[0].(*apiextlegacy.CustomResourceDefinition)
+			crd.Spec.PreserveUnknownFields = nil
+		case g.PreserveUnknownFields == nil, g.PreserveUnknownFields != nil && !*g.PreserveUnknownFields:
+			// it'll be false here (coming from v1) -- leave it as such
+		default:
+			return fmt.Errorf("you may only set PreserveUnknownFields to true with v1beta1 CRDs")
```
mengqiy: It seems current behavior of this piece of code can be captured in the following table:

| initial value  | v1beta1  | v1  |   
|---|---|---|
| true  | nil  | not allowed  |   
| false  | **not allowed**  | false  |   
| nil  | nil  | nil  |   

I don't understand why if the initial value is false, we don't allowed it in v1beta1.
Per https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#pruning-versus-preserving-unknown-fields, it seems we can set it to false in v1beta1.

droot:
Looks good to me other than @mengqiy comment.

pkg/crd/gen.go
DirectXMan12: your chart is off (which maybe means this code needs refactoring), but the second case in the switch statement handles this.

DirectXMan12: basically, this code is roughly:

```rust
match g.PreserveUnknownFields {
  Some(true) if v1beta1Only => { crd.Spec.PreserveUnknownFields = nil }
  None if v1beta1Only => { crd.Spec.PreserveUnknownFields = nil }
  Some(false) => { /* nothing, false by default }
  None => { /* Nothing, false by default */ }
}
```

I'll see if I can make it more obvious

mengqiy: Sorry for misreading it :P

PR #365, opened 2019-11-25
:sparkles: CRD generation integration tests

sebgl:
This PR adds a bunch of integration tests to CRD generation.

Tests cover the following:
- CRD generation from a single API versions (cronjob v1)
    - using `apiextensions.k8s.io/v1` CustomResourceDefinitions
    - using  `apiextensions.k8s.io/v1beta1` CustomResourceDefinitions
- CRD generation from 2 API versions (cronjob v1 and cronjob v2)
    - using `apiextensions.k8s.io/v1` CustomResourceDefinitions
    - using  `apiextensions.k8s.io/v1beta1` CustomResourceDefinitions
    - using both `apiextensions.k8s.io/v1` and `apiextensions.k8s.io/v1beta1`
    - with `maxDescLen` set to a non-default value
    - with `trivialVersions=true`
    - with `preserveUnknownFields=false`

The main goal is to detect present and future regressions and inconsistencies.
 
Implementation details:

* Tests are relying on `testdata` files, pre-generated manually. Similar to the `schemapatcher` integration tests. I chose not to use an `expected` directory but keep the yaml manifests where relevant (either in `v1` for CRD generates from a single package, or one level higher for CRDs covering both cronjob v1 and v2).
* I added an in-memory output writer so we don't have to deal with temporary files.

I'll add a few inline comments in the PR where I feel necessary.


pkg/crd/testdata/Makefile
```diff
@@ -0,0 +1,27 @@
...
+	# Generate separate CRDs for cronjob v1 and v2
+	# - using crd v1
+	$(GEN) crd:crdVersions=v1 paths=./cronjob/v1 > cronjob/v1/cronjob_crdv1.yaml
+	$(GEN) crd:crdVersions=v1 paths=./cronjob/v2 > cronjob/v2/cronjob_crdv1.yaml
+	# - using crd v1beta1
+	$(GEN) crd:crdVersions=v1beta1 paths=./cronjob/v1 > cronjob/v1/cronjob_crdv1beta1.yaml
+	$(GEN) crd:crdVersions=v1beta1 paths=./cronjob/v2 > cronjob/v2/cronjob_crdv1beta1.yaml
+	# Generate a combined CRD of Cronjob v1 & v2
+	# - using crd v1
+	$(GEN) crd:crdVersions=v1 paths=./cronjob/... > cronjob/cronjob_crdv1.yaml
```
sebgl: Whenever we generate yaml manifests covering 2 versions (here `cronjob/v1` and `cronjob/v2`), the same content appears twice in the generated output. In most situations the last content replaces the first one, but this is not the case here when using `output:stdout`.
This should be fixed with https://github.com/kubernetes-sigs/controller-tools/pull/350. These tests at least served to prove that :)

pkg/crd/gen_integration_test.go
```diff
@@ -0,0 +1,224 @@
...
+		By("initializing the generator")
+		var gen genall.Generator = params.generator
+		rt, err := genall.Generators{&gen}.ForRoots(params.rootPaths)
+		Expect(err).NotTo(HaveOccurred())
+
+		outputDir, err := ioutil.TempDir("", "controller-tools-test")
+		Expect(err).NotTo(HaveOccurred())
+		defer os.RemoveAll(outputDir)
+		var buf bytes.Buffer
+		rt.OutputRules.Default = genall.OutputToBuffer{Buffer: &buf}
```
sebgl: If that makes sense we can probably adapt other integration tests to also not use temporary files.

pkg/crd/testdata/cronjob/v2/cronjob_types.go
```diff
@@ -0,0 +1,236 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// CronJobSpec defines the desired state of CronJob
+type CronJobSpec struct {
+	// V2Field is a dummy field that does not exist in v1.
+	V2Field string `json:"v2field"`
```
sebgl: Differences with `v1/cronjob_types`:

- set package name to `v2`
- add `V2Field` to make sure the target CRD is different
- set `kubebuilder:storageversion`

sebgl:
@DirectXMan12 we discussed this briefly on Slack, you may want to have a look :)

kensipe:
@sebgl looks like this needs a rebase

sebgl:
@kensipe this is now rebased on master.
Happy to squash all commits into one if you'd refer (but I guess the "squash and merge" Github feature does the same).

mengqiy:
/assign @DirectXMan12 
per https://github.com/kubernetes-sigs/controller-tools/pull/365#issuecomment-558326967

mcristina422:
@sebgl are you able to rebase this? Or should someone open a new PR?

I think this test is extremely valuable

PR #366, opened 2019-11-27
✨ Add sttts as an approver

soltysh:
/assign @DirectXMan12


DirectXMan12:
let's follow the pattern from CR -- have a `controller-tools-approvers` alias too

Issue #377, opened 2019-12-19
Problems with prepending "---" in when generating YAML


porridge:
There are two issues with [this code](https://github.com/kubernetes-sigs/controller-tools/blob/998668a61585e05bbd875e2fb0d0f536af4cdb58/pkg/genall/genall.go#L110-L115):
1. the `---` dashes separator is needlessly prepended even at the top of files that contain just a single file. This is semantically equivalent to the canonical form or using `---` as a *separator*, however it creates problems with text-level comparisons of output from sources which more closely adhere to the YAML standard (think golden files, etc).
1. the condition guarding `return io.ErrShortWrite` is wrong, because it does not take into account the five extra written bytes. It is possible (however unlikely) that a write would be truncated by a few bytes (e.g. due to disk full condition) but this code would not notice.

PR #378, opened 2019-12-19
:bug: Correct the YAML separator emission logic.

porridge:
Fixes: #377 

See that issue for more details.

Unfortunately we cannot use the nice Encoder API from the
gopkg.io/yaml.v3 package which encapsulates this logic already, because
we need to first emit JSON and then convert it to YAML (I always
wondered why it is so. It turns out because there are no `yaml:` tags on
k8s objects.)

### DIFF
--- a/pkg/genall/genall.go
+++ b/pkg/genall/genall.go
@@ -102,12 +102,21 @@ func (g GenerationContext) WriteYAML(itemPath string, objs ...interface{}) error
 	}
 	defer out.Close()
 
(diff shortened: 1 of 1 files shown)


mcristina422:
There is https://github.com/kubernetes-sigs/controller-tools/pull/310 also open for this

DirectXMan12:
#310 is simpler, and I think should be fine.  We're allowed to start with `---`, just apparently not `\n---`.

Issue #328, opened 2019-09-25
Spec of API machinery's Quantity is wrong (incomplete)


sttts:
https://github.com/kubernetes-sigs/controller-tools/blob/538d4b23662f14d691b7c4c3a9a841a497a4f1e2/pkg/crd/known_types.go#L58 specifies quantities as string, while they allow ints and numbers https://github.com/kubernetes/kubernetes/blob/5713c22eecff4610026643fbd3d37c33a43c168d/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity_test.go#L219. We cannot express int-or-number-or-string in structural schemas, but we have `x-kubernetes-int-or-string`. That would make the spec more complete than today.

DirectXMan12:
huh?

The code in k8s.io/apimachinery says otherwise: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity.go#L393

sttts:
Isn't https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity.go#L646 the relevant line? Quotes are optional. The OpenAPI spec you link to is incomplete (OpenAPI v2 ...).

DirectXMan12:
Interesting... was that intended?  Generally our OpenAPI is less restrictive that what we accept (i.e. we validate more than what our OpenAPI says), but in this case the OpenAPI is *more* restrictive that what we accept.

DirectXMan12:
Technically, neither is right (it should be `int or (string that matches the quantity format)`), but I guess since we accept raw integers in the parsing logic we can move forward with that.

sttts:
IMO the v2 spec is just wrong. A published spec should never be too restrictive. It may be incomplete server side validation will do the final check anyway.

I agree, we could add a regex to specify `string that matches the quantity format` if we want to. Maybe we actually should to avoid unmarshalling errors in Go code.

dthomson25:
@sttts and @DirectXMan12, isn't it true that a quantity can be float (i.e. 1.2 not "1.2")? If that's the case, then isn't a stringOrInt is too restrictive?

DirectXMan12:
[...]

PR #381, opened 2019-12-20
✨ Change Quantity type to int-or-string

arjunrn:
This PR address the issue #328 

We need this because we want to use the `controller-gen` tool to start managing our CRDs. But the generated CRD only allows `string` types for `Quantity`.

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -60,7 +60,7 @@ var KnownPackages = map[string]PackageOverride{
 				{Type: "integer"},
 				{Type: "string"},
 			},
-			Pattern: "^(\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))(([KMGTIE]i)|[mkMGTPE]|((\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))))?$",
+			Pattern: "^(\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))))?$",
 		}
 		// No point in calling AddPackage, this is the sole inhabitant
 	},
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -60,6 +60,7 @@ var KnownPackages = map[string]PackageOverride{
 				{Type: "integer"},
 				{Type: "string"},
 			},
+			Pattern: "^(\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))(([KMGTIE]i)|[mkMGTPE]|((\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))))?$",
 		}
 		// No point in calling AddPackage, this is the sole inhabitant
 	},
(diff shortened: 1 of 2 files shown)


DirectXMan12:
/hold

I'm not sure this is right.  I'm replying in the linked issue.

DirectXMan12:
/hold cancel

as discussed on the main object, I think we can move forward with this.  The OpenAPI in core kubernetes is just wrong.

pkg/crd/known_types.go
```diff
@@ -55,7 +55,11 @@ var KnownPackages = map[string]PackageOverride{
 	"k8s.io/apimachinery/pkg/api/resource": func(p *Parser, pkg *loader.Package) {
 		p.Schemata[TypeIdent{Name: "Quantity", Package: pkg}] = apiext.JSONSchemaProps{
 			// TODO(directxman12): regexp validation for this (or get kube to support it as a format value)
-			Type: "string",
+			XIntOrString: true,
+			AnyOf: []apiext.JSONSchemaProps{
+				{Type: "integer"},
+				{Type: "string"},
```
DirectXMan12: @sttts if we use intorstring, are we allowed to further restrict the variants and still be structural?  i.e.

```yaml
anyOf:
- type: int
- type: string
  pattern: "^(\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))(([KMGTIE]i)|[mkMGTPE]|((\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))))?$"

DirectXMan12: (I think that matches the productions in `k8s.io/apimachinery/pkg/api/resource/quantity.go`)

DirectXMan12: (also, we should really just have `format: quantity` supported in k/k)

sttts: I like that idea about format. We should add that. For old servers though we will need the pattern.

sttts:
Lgtm if the pattern is added.

arjunrn:
@sttts @DirectXMan12 I have added the pattern now. PTAL

DirectXMan12:
Overall looks good.  Did anyone validate that the pattern actually works?  I just did that off the top of my head, so it either works, causes explosions and general misery, or possibly summons something from the bowels of the internet to posses your network card.

arjunrn:
@DirectXMan12 I did not validate it in code till now but here's a [test](https://play.golang.org/p/C9cp32EbDMu) based on all the [valid test cases](https://github.com/kubernetes/kubernetes/blob/5713c22eecff4610026643fbd3d37c33a43c168d/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity_test.go#L219) in k/k. So there are some more types of values which can be specified like exponentials which maybe the regex should support. If we try to support all the accepted values like you said it might summon something. WDYT?

DirectXMan12:
Yeah, the docs for quantity don't match the actual code (shocked, I'm shocked I tell you).  Trying to fix that now.

DirectXMan12:
based on that test, this should be good `"^(\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\\+|-)?(([0-9]+(\\.[0-9]*)?)|(\\.[0-9]+))))?$"` -- was missing some SI suffixes and the `[eE]`

arjunrn:
@DirectXMan12 I've updated the regex now.

PR #384, opened 2019-12-28
:sparkles: make it possible for ldflags to substitute import path in deepcopy

mengqiy:
Make it possible to use ldflags to substitute import path in deepcopy.
This can be helpful for building a controller-gen that works in non-native go build system (e.g. bazel).


pkg/deepcopy/gen.go
```diff
@@ -33,8 +33,9 @@ import (
 // NB(directxman12): markers.LoadRoots ignores autogenerated code via a build tag
 // so any time we check for existing deepcopy functions, we only seen manually written ones.
 
-const (
-	runtimeObjPath = "k8s.io/apimachinery/pkg/runtime.Object"
+var (
```
DirectXMan12: please add a comment explaining these.  Otherwise, LGTM

PR #389, opened 2020-01-10
Update OWNERS_ALIASES

gerred:
I'm super active on controller-runtime and kubebuilder in helping people in Slack and other environments, a reviewer on CR, soon to hopefully be a reviewer on CR, and since CT is more core to everything I want to be very active here as well.


DirectXMan12:
Please add a new alias for reviewer (like in CR and KB), so we can have `reviewer --> approver` flow.  Otherwise adding you as a reviewer here sounds good :-)

Issue #385, opened 2020-01-03
Rethinking metadata schema generation in CRD yamls


tamalsaha:
Currently controller-tools does not generate any properties for metadata fields at any level in CRD yamls to accommodate for structural schema.

https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/known_types.go#L29-L33

My it is my understanding that only top level metadata field is restricted to define name & generatedName fields. Any other nested metadata field can have detailed schema. Some common cases are embedding PodTemplate, PersistentVolumeClaim in CRDs.

Currently we are creating our own `ObjectMeta` types are using them in our CRDs to work around this issue. 

Do you think controller-tools could just strip the properties for top level metadata field? One way to do that will be to remove `ObjectMeta` from known types and then clean the generated yaml before writing to disk.

Any thoughts?

tamalsaha:
/sig api-machinery

cc: @sttts @munnerz @DirectXMan12

sttts:
I would prefer to see a stripped down EmbeddedObjectMeta if possible, or at least let controller-gen do the restriction. ObjectMeta has so many fields which do not make sense for embedded ObjectMeta. Also the schema is super huge.

tamalsaha:
I like the idea of `EmbeddedObjectMeta`. We kind of doing that in our custom type. https://github.com/kmodules/offshoot-api/blob/master/api/v1/pvc.go#L24-L98

DirectXMan12:
I think #395 is actually close to being mergable, it's just a bit out of date.  If someone wants to pick that up, I've left comments on it.

sttts:
@DirectXMan12 see my comment in 395. IMO we should not do that. Instead let's add a sensible EmbeddedObjectMeta to API machinery and make controller-tools warn if that is not used instead of ObjectMeta.

DirectXMan12:
@sttts that doesn't work for when you embed things with ObjectMeta in them already though -- e.g. PodTemplateSpec, which is probably the most common case.  We'd have to swap that out on demand or something.

sttts:
True. So we must add special handling for embedded ObjectMeta to only specifiy sensible fields (more than just `type: object` but considerably less than a full ObjectMeta. It's ugly. As written in the other issue: sin of the past :-/

EronWright:
[...]

PR #395, opened 2020-01-26
:warning: Only prevent validation of top-level ObjectMeta field

arjunrn:
This addresses #385. Currently the validation for `ObjectMeta` is not generated even if it is used somewhere in the schema other than at the top-level. To fix this I mostly reverted the changes from #266 and then after the schema has been generated patched the top-level `metadata` to have no validation. The resulting schema for `ObjectMeta` is still big as @sttts mentions in the issue but I don't see how we can drop only the read-only fields. We can have a hand-written schema with only the required fields but this would need more effort in the long term keeping it in sync with upstream. I'm open to suggestions on how this can be improved.

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -108,6 +108,7 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		parser.NeedCRDFor(groupKind, g.MaxDescLen)
 		crdRaw := parser.CustomResourceDefinitions[groupKind]
 		addAttribution(&crdRaw)
+		FixupMetadata(crdRaw)
 
 		versionedCRDs := make([]interface{}, len(crdVersions))
 		for i, ver := range crdVersions {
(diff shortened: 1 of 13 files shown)


pgier:
FWIW, this patch worked to fix our use case in prometheus-operator (https://github.com/coreos/prometheus-operator/issues/3041)

DirectXMan12:
Title edit: this is a breaking change

DirectXMan12:
@sttts it almost sounds like we want `format: x-kubernetes-metadata` or something here, longterm

DirectXMan12:
I think this approach is fairly reasonable, but it is a breaking change.  We should have a generator option to turn this on or off, I think.  Please add that, and then this is good to go.

DirectXMan12:
(also, I realize this is super old at this point, so if the original submitter is no longer available and someone wants to take this over, please do)

sttts:
I still think it is not a good idea to have full ObjectMeta in an embedded object. The technical work around is have some special EmbeddedObjectMeta with a restricted set of fields. I think that's the route that some operators went by now.

DirectXMan12:
What do we do about `PodTemplateSpec`, then?

sttts:
> What do we do about PodTemplateSpec, then?

Sins of the past.

Do we have any restriction on PodTemplateSpec in contrast to object-top-level meta data? We should have gone the same way there, i.e. EmbeddedObjectMeta.

champak:
For operators that are already deployed re-organizing the CRD to add embeddedObjectMeta may not be feasible for an upgrade. If we could keep at least the name/namespace  (typically these get referenced by other objects) of the ObjectMeta around post conversion, that would help.  https://github.com/kubernetes-sigs/controller-tools/issues/448  Thanks

arjunrn:
Closing in favor of #539

PR #400, opened 2020-02-06
✨ Add kubebuilder:label CRD marker to add labels

vincepri:

/assign @gerred @DirectXMan12

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -45,6 +45,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:skipversion", markers.DescribesType, SkipVersion{})).
 		WithHelp(SkipVersion{}.Help()),
+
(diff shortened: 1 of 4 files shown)


DirectXMan12:
Context: we've resisted doing this so far, since it's something that's generally pretty easy to do with kustomize and friends, and it's not really generally an attribute inherent to the CRD (i.e. it's something you might want to change w/o changing your Go code).

What's the motivating usecase here?

/hold

vincepri:
Adding the marker allows for the CRDs in `crd/bases` to be populated with the label, rather than using kustomize to do add the label (which yeah it's pretty straightforward).

This is useful in `envtest` for example, we require a specific label to be present on the CRD when parsing object references. 

This isn't a release blocker for us, we found another way (hardcode a test CRD).

DirectXMan12:
I think this generally encourages things that are bad patterns IMO, and the workarounds are pretty easy (kustomize, loading manually and setting the field, using a client to set the field, etc), so I'd prefer to not do this.

vincepri:
Sounds good to me, thanks for providing an explanation.

/close

PR #403, opened 2020-02-11
✨ Add kubebuilder:unservedversion CRD marker

rajathagasthya:
Used to set `served=false` for an apiversion.

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -45,6 +45,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:skipversion", markers.DescribesType, SkipVersion{})).
 		WithHelp(SkipVersion{}.Help()),
+
+	must(markers.MakeDefinition("kubebuilder:unservedversion", markers.DescribesType, UnservedVersion{})).
+		WithHelp(UnservedVersion{}.Help()),
 }
 
 // TODO: categories and singular used to be annotations types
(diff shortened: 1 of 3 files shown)


pkg/crd/spec.go
```diff
@@ -151,6 +151,19 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind, maxDescLen *int) {
...
+	for _, ver := range crd.Spec.Versions {
+		if ver.Served {
+			served = true
+			break
+		}
+	}
+	if !served {
+		// just add the error to the first relevant package for this CRD,
+		// since there's no specific error location
+		packages[0].AddError(fmt.Errorf("CRD for %s does not serve any version", groupKind))
```
camilamacedo86: WDYT about something as:

```suggestion
		packages[0].AddError(fmt.Errorf("CRD for %s has the vesion %s which is no longer supported", groupKind, crdVersion ))
```

pkg/crd/spec.go
```diff
@@ -151,6 +151,19 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind, maxDescLen *int) {
 		packages[0].AddError(fmt.Errorf("CRD for %s has no storage version", groupKind))
 	}
 
+	served := false
+	for _, ver := range crd.Spec.Versions {
```
camilamacedo86:
```suggestion
        crdVersion := ""
	for _, ver := range crd.Spec.Versions {
```

DirectXMan12:
drive-by naming nit: this might be clearer as "unserved" or something to that effect

rajathagasthya:
I used `unserveversion` just to be similar to `disableversion` and `skipversion`.

pkg/crd/markers/crd.go
```diff
@@ -290,4 +293,23 @@ func (s Resource) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, version s
 	return nil
 }
 
+// +controllertools:marker:generateHelp:category=CRD
+
+// UnserveVersion does not serve this version.
```
DirectXMan12: nit: this seems a lot better as "unservedversion" (with a d).  Otherwise LGTM

rajathagasthya: Sure, I can change this.

PR #405, opened 2020-02-14
:bug: use map in FindKubeKinds

mengqiy:
It makes more sense to return a hash set instead of a list, since hash set help de-duplicate group-kind.

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -104,7 +104,7 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		crdVersions = []string{"v1beta1"}
 	}
 
-	for _, groupKind := range kubeKinds {
+	for groupKind := range kubeKinds {
 		parser.NeedCRDFor(groupKind, g.MaxDescLen)
 		crdRaw := parser.CustomResourceDefinitions[groupKind]
 		addAttribution(&crdRaw)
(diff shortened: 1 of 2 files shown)


pkg/crd/gen.go
```diff
@@ -206,9 +206,9 @@ func FindMetav1(roots []*loader.Package) *loader.Package {
 // FindKubeKinds locates all types that contain TypeMeta and ObjectMeta
 // (and thus may be a Kubernetes object), and returns the corresponding
 // group-kinds.
-func FindKubeKinds(parser *Parser, metav1Pkg *loader.Package) []schema.GroupKind {
+func FindKubeKinds(parser *Parser, metav1Pkg *loader.Package) map[schema.GroupKind]bool {
 	// TODO(directxman12): technically, we should be finding metav1 per-package
-	var kubeKinds []schema.GroupKind
+	kubeKinds := map[schema.GroupKind]bool{}
```
camilamacedo86: Why the map with key= GroupKind and value=bool to say it was or not found it required. 
Should not more make sense return a list with all found GroupKinds?

mengqiy: `bool` was intended be a placeholder.
I have replaced it with an empty struct.

mengqiy: Map helps de-duplicating group-kind

camilamacedo86: Tks for the clarification.

PR #408, opened 2020-02-26
✨ Add support for MatchPolicy in webhook marker

vincepri:


DirectXMan12:
This needs to be optional, otherwise it's a breaking change

Issue #343, opened 2019-10-15
singular annotation does not get apply to the crd


danigrmartinez:
I wanted to get a specific single and plural name not base on the kind of the CRD. According to the [documentation](https://book.kubebuilder.io/reference/markers/crd.html) can be done using the `+kubebuilder:resource:singular`.

### Reproduce the issue
```
// +kubebuilder:resource:singular=my-singular,path=my-plural,scope=namespaced
```

### Expected CRD result
```
  group: group.com
  names:
    kind: MyKindType
    listKind: MyKindTypeList
    plural: my-plural
    singular: my-singular
  scope: namespaced
```

### Actual CRD result
```
  group: group.com
  names:
    kind: MyKindType
    listKind: MyKindTypeList
    plural: my-plural
    singular: mykindtype
  scope: namespaced
```

This is because on the [ApplyToCRD function](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/crd.go#L298-L313), Singular filed is not use.

I saw some comments about singular but I couldn't understand if it is intentional ignore or it was just forgot.
Thanks!

PR #409, opened 2020-02-26
🐛Fix singular argument for resource marker

hypnoglow:
Singular argument on `+kubebuilder:resource` marker was never actually applied to CRD. This PR fixes it.

Fixes https://github.com/kubernetes-sigs/controller-tools/issues/343

Also fixes a small typo 🙃

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -277,6 +277,9 @@ func (s Resource) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, version s
 	if s.Path != "" {
 		crd.Names.Plural = s.Path
 	}
+	if s.Singular != "" {
+		crd.Names.Singular = s.Singular
+	}
 	crd.Names.ShortNames = s.ShortName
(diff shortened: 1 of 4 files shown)


vincepri:
can you add a test?

DirectXMan12:
o_O yes, test please.  I'm not sure how we missed this :-/

hypnoglow:
@vincepri @DirectXMan12 Can you please guide what test do we want? I see no unit tests or test files in `markers` package, should I create one? Or is there another place where related tests are located?

mengqiy:
You probably add the marker in https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/testdata/cronjob_types.go and then check the generated CRD with the [golden file](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml).

hypnoglow:
@mengqiy Thanks 🙏

I've added the marker and updated the golden file.

PR #411, opened 2020-03-04
:running: Make webhook marker MatchPolicy optional

vincepri:

Sidenote: we need to add verification to the CI jobs

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -60,7 +60,7 @@ type Config struct {
 	// MatchPolicy defines how the "rules" list is used to match incoming requests.
 	// Allowed values are "Exact" (match only if it exactly matches the specified rule)
 	// or "Equivalent" (match a request if it modifies a resource listed in rules, even via another API group or version).
-	MatchPolicy string
+	MatchPolicy string `marker:",optional"`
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/parser.go
```diff
@@ -186,7 +186,7 @@ func (c Config) matchPolicy() *admissionreg.MatchPolicyType {
 	case strings.ToLower(string(admissionreg.Equivalent)):
 		matchPolicy = admissionreg.Equivalent
 	default:
-		matchPolicy = admissionreg.MatchPolicyType(c.MatchPolicy)
+		return nil
```
rajathagasthya: Does the `Unrecognized values are passed through` comment on L180 still hold with this change? IIUC, this will ignore invalid values, which is different behavior from failurePolicy for example.

vincepri: Good point, fixed!

rajathagasthya: @vincepri Are we okay with a different behavior than failurePolicy? i.e. failurePolicy uses any value you give and server returns error if invalid, but matchPolicy ignores invalid values.

vincepri: I think so, we probably want to avoid to set a value that's not supported

PR #412, opened 2020-03-05
Resolve reference in order to remove $ref which is not supported by k8s

### LINKED ISSUE #324: controller-tools creates schema with '$ref' which is not supported in k8s
```golang
type Bar struct{
	A string `json:"a"`
}

// Foo is a Bar
type Foo Bar

type CronJobSpec struct {
	Foo Foo `json:"foo"`
}
```

```yaml
---
metadata:
  creationTimestamp: null
  name: cronjobs.batch.tutorial.kubebuilder.io
spec:
  group: batch.tutorial.kubebuilder.io
  names:
    kind: CronJob
    listKind: CronJobList
    plural: cronjobs
    singular: cronjob
  scope: ""
  subresources:
    status: {}
  validation:
    openAPIV3Schema:
      description: CronJob is the Schema for the cronjobs API
      properties:
        apiVersion:
          type: string
        kind:
          type: string
        metadata:
          type: object
        spec:
          description: CronJobSpec defines the desired state of CronJob CronJobSpec
            defines the desired state of CronJob
          properties:
            foo:
              $ref: '#/definitions/Bar'
              description: Foo is a Bar
          required:
          - foo
          type: object
      type: object
  version: v1
  versions:
  - name: v1
    served: true
    storage: true
status:
  acceptedNames:
    kind: ""
    plural: ""
  conditions: []
  storedVersions: []
```

`kubectl apply` returns a `$ref: Forbidden: $ref is not supported`

/kind bug

tatsuhiro-t:
Fixes #324

### DIFF
--- a/pkg/crd/flatten.go
+++ b/pkg/crd/flatten.go
@@ -248,7 +248,7 @@ func (f *Flattener) loadUnflattenedSchema(typ TypeIdent) (*apiext.JSONSchemaProp
 	if !found {
 		return nil, fmt.Errorf("unable to locate schema for type %s", typ)
 	}
-	return &baseSchema, nil
+	return f.FlattenSchema(baseSchema, typ.Package), nil
 }
 
 // FlattenType flattens the given pre-loaded type, removing any references from it.


pkg/crd/flatten.go
```diff
@@ -248,7 +248,7 @@ func (f *Flattener) loadUnflattenedSchema(typ TypeIdent) (*apiext.JSONSchemaProp
 	if !found {
 		return nil, fmt.Errorf("unable to locate schema for type %s", typ)
 	}
-	return &baseSchema, nil
+	return f.FlattenSchema(baseSchema, typ.Package), nil
```
DirectXMan12: this is absolutely not the right way to fix this

DirectXMan12: given the contract of this method

PR #413, opened 2020-03-07
✨ Add new resourceNames field in RBAC marker

### LINKED ISSUE #401: RBAC generation with specific resourceNames
The regular RBAC generation annotation works great if you want to give permissions over all instances of a certain type. Say you want to give permission to get all statefulsets, that's as easy as

`// +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get`

But what if you want to give permissions only over a few instances? For example, when dealing with PodSecurityPolicies, you'll probably want to generate an RBAC rule that looks like

```
- apiGroups:
  - policy
  resources:
  - podsecuritypolicies
  verbs:
  - use
  resourceNames:
  - psp.1
```

But there's no way to do that with the RBAC generation is there?
`// +kubebuilder:rbac:groups="policy",resources=podsecuritypolicies,verbs=use,resourceNames=psp.1`

ends up generating 

```
- apiGroups:
  - policy
  resources:
  - podsecuritypolicies
  verbs:
  - use
```

So how can you generate RBAC rules for specific resource instances?

djzager:
This makes it possible to generate policy rules restricted to specific
instances.

Fixes #401

### DIFF
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -19,7 +19,7 @@ limitations under the License.
 //
 // The markers take the form:
 //
-//  +kubebuilder:rbac:groups=<groups>,resources=<resources>,verbs=<verbs>,urls=<non resource urls>
+//  +kubebuilder:rbac:groups=<groups>,resources=<resources>,resourceNames=<resource names>,verbs=<verbs>,urls=<non resource urls>
 package rbac
 
 import (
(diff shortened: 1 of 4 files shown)


pkg/rbac/parser.go
```diff
@@ -48,6 +48,10 @@ type Rule struct {
 	Groups []string `marker:",optional"`
 	// Resources specifies the API resources that this rule encompasses.
 	Resources []string `marker:",optional"`
+	// ResourceNames specifies the names of the API resources that this rule encompasses.
+	// Create requests cannot be restricted by resourcename, as the object's name
+	// is not known at authorization time.
```
DirectXMan12: Nit:

```suggestion
	// ResourceNames specifies the names of the API resources that this rule encompasses.
	//
	// Create requests cannot be restricted by resourcename, as the object's name
	// is not known at authorization time.
```

This'll put the extra `Create requests...` in the details field of the docs.

DirectXMan12:
/approve
/hold

unhold when the changes are resolved

DirectXMan12:
Squash down that trivial commit and we're good to go here (my one gripe with the suggestions is that is produces commits like that).

djzager:
> Squash down that trivial commit and we're good to go here (my one gripe with the suggestions is that is produces commits like that).

went to squash and ended up chasing a squirrel. oops

PR #422, opened 2020-03-19
✨Deprecate the duplicated kubebuilder:validation:XPreserveUnknownFields marker.

### LINKED ISSUE #380: duplicate marker: XPreserveUnknownFields vs PreserveUnknownFields
@sttts did you intend to have have preserveunknownfields registered as both `kubebuilder:validation:XPreserveUnknownFields` and `kubebulder:pruning:PreserveUnknownFields` (field only)?

I seem to recall that only the latter was desired.

jiachengxu:
Remove the duplicated kubebuilder:validation:XPreserveUnknownFields marker.
Fixes #380


DirectXMan12:
A quick survey says that this marker is actually still being used, so I'm going to tweak this to deprecate it for the moment then merge.

Issue #398, opened 2020-01-31
Controller-gen creating unwanted CRDs


thackem:
In my API structs I have a spec field like:
```
type MyObjectSpec struct {
	Live bool `json:"live,omitempty"`
	ComplexRef ObjectRef `json:"object,omitempty"`
	SimpleRef *v1.LocalObjectReference `json:"configMapRef"`
}

// no kubebuilder annotation here

type ObjectRef struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
	Foo string `json:"foo"`
}
```

This results in an ObjectRef CRD file being generated in config/crd/bases when I run the default `make generate` target.

DirectXMan12:
/kind feature
/priority important-longterm

/help

this should prob just be a marker to disable this that we can read in `gen.go`

zroubalik:
any update on this please?

DirectXMan12:
I think you should be able to use the `skipversion` marker here.

zroubalik:
@DirectXMan12 but this marker disable CRD generation for all types in the particular version, right?
For example I have `MyTypeA`, `MyTypeB` and `InternalTypeA` in version `v1alpha1` and I'd like to generate CRDs only for `MyTypeA` and `MyTypeB`, not for `InternalTypeA`.
Unfortunately I haven't found a way how to achieve that.

PR #426, opened 2020-03-26
✨ Supported CRD Generation with Embedded Base Types

seglberg:
Allow generating CRDs for types which indirectly embed ObjectMeta and TypeMeta.

Previously, each CRD type needed to directly embed these metav1 types in order to be generated. Now, types will be generated if they embed other types which embed ObjectMeta and TypeMeta.

For example, it will now pick up the following situation:

```go
type ObjectBase struct {
        ...
}

type MyCRDType struct {
	ObjectBase `json:",inline"`
        ...
}
```

and generate a CRD for `MyCRDType`.

This now works because `FindKubeKinds` now recursively checks the fields of a given type if the field itself is embedded. 

The only issue is this produces an unwanted CRD for ObjectBase, however this will be solved by #398 (will tackle that next).

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -221,56 +221,83 @@ func FindMetav1(roots []*loader.Package) *loader.Package {
 func FindKubeKinds(parser *Parser, metav1Pkg *loader.Package) map[schema.GroupKind]struct{} {
 	// TODO(directxman12): technically, we should be finding metav1 per-package
 	kubeKinds := map[schema.GroupKind]struct{}{}
-	for typeIdent, info := range parser.Types {
-		hasObjectMeta := false
-		hasTypeMeta := false
-
-		pkg := typeIdent.Package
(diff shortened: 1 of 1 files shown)


seglberg:
CLA Signed

DirectXMan12:
Sorry for just *completely* missing this.  Can you clarify the usecase a bit here?  It's not clear to me what patterns this enables.

svzieg:
Hi @DirectXMan12 , We actually fall into this use case. 

We actually have lots of api type about 8 or 9. Each of them inherits from a base type which handles generation of labels and annotations in a way that its satisfies our API-Resource Interface. This interface is then used to generate and modify the cluster resources. 

Since we actually have lots of CRDs, to configure, but one reconcling logic behind. We have to generalise the CRDs.

For the moment we have to copy all methods to each CRD struct, so it fits our interfaces which is very hard to maintain.

PR #427, opened 2020-04-01
✨ Generate schema for types with custom JSON marshaling as Any type

### LINKED ISSUE #391: markers to skip fields that does not have json tags for crd generation
`controller-tools` version: 0.2.2

Currently I am getting this error after upgrading controller-tools from 0.2.2 from 0.1.9 because in our CRD status we included a go struct `URL` (https://github.com/knative/pkg/blob/43ca049cdbdad4892b0c926e7c6536d064efde72/apis/duck/v1beta1/addressable_types.go#L40) which does not have json tags and this blocks generating the crd manifests.
```
/usr/local/go/src/net/url/url.go:357:2: encountered struct field "Scheme" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:358:2: encountered struct field "Opaque" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:359:2: encountered struct field "User" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:360:2: encountered struct field "Host" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:361:2: encountered struct field "Path" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:362:2: encountered struct field "RawPath" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:363:2: encountered struct field "ForceQuery" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:364:2: encountered struct field "RawQuery" without JSON tag in type "URL"
/usr/local/go/src/net/url/url.go:365:2: encountered struct field "Fragment" without JSON tag in type "URL"
```

We want a way to skip fields when generating crd manifests.

skaslev:
Currently controller-gen complains about structs with missing json tags even
when those structs implement custom JSON marshalling.

With this change we check if a type implements custom JSON marshalling and if it
does, we output schema for Any type. This still allows the validation type to
overriden with a marker.

Fixes #391

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -19,6 +19,7 @@ package crd
 import (
 	"fmt"
 	"go/ast"
+	"go/token"
 	"go/types"
 	"strings"
 
@@ -104,6 +105,11 @@ func (c *schemaContext) requestSchema(pkgPath, typeName string) {
 
 // infoToSchema creates a schema for the type in the given set of type information.
 func infoToSchema(ctx *schemaContext) *apiext.JSONSchemaProps {
+	if obj := ctx.pkg.Types.Scope().Lookup(ctx.info.Name); obj != nil && implementsJSONMarshaler(obj.Type()) {
+		schema := &apiext.JSONSchemaProps{Type: "Any"}
+		applyMarkers(ctx, ctx.info.Markers, schema, ctx.info.RawSpec.Type)
+		return schema
+	}
 	return typeToSchema(ctx, ctx.info.RawSpec.Type)
 }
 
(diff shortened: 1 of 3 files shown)


DirectXMan12:
Thanks!  We've wanted something like this for a while.

yuzisun:
@skaslev any chance to update the test and get this in? thank you!

DirectXMan12:
@yuzisun (or anyone else) -- if skaslev doesn't get back to us and you feel like writing the tests, feel free to add an additional commit and submit a new PR (and link to this one).

skaslev:
@DirectXMan12 I added two test cases similar to the problematic types in Knative APIs. The initial version of this PR wasn't entirely correct since it wasn't always applying the markers to the schema for types with custom json serialization. I think the new version works correctly.

Can you take another look?

DirectXMan12:
/lgtm
/approve

apparently I never actually pressed the comment button this -- sorry about that

yuzisun:
Any idea why this is not merged ?

vincepri:
@DirectXMan12 @skaslev Using the tip of the main branch, I'm now getting an invalid CRD:

```
  CustomResourceDefinition.apiextensions.k8s.io "kubeadmcontrolplanes.controlplane.cluster.x-k8s.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[kubeadmConfigSpec].properties[initConfiguration].properties[bootstrapTokens].items.properties[token].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
```

vincepri:
We should improve the error message saying that when using this capability, folks should manually specify the validation type

alvaroaleman:
`Any` is just not valid. If folks have a custom marshaller, they must specify the json tag or use a kubebuilder annotation. This PR doesn't fix the issue it claims to fix while introducing a new issue. Revert in https://github.com/kubernetes-sigs/controller-tools/pull/505

PR #428, opened 2020-04-01
🐛 Fix flattening chains of references to references

### LINKED ISSUE #324: controller-tools creates schema with '$ref' which is not supported in k8s

skaslev:
Without this change running controller-gen on Knative's APIs
results in unexpanded reference for the Conditions[1] type in
the generated OpenAPI schema:

    conditions:
      $ref: '#/definitions/knative.dev~1pkg~1apis~0Conditions'

After this change Conditions gets properly flattened and no references
are present in the resulting schema.

Fixes #324

[1] https://github.com/knative/pkg/blob/ce05a3de31ed80e98862a0ea024977097fd0e4a5/apis/duck/v1/status_types.go#L34

### DIFF
--- a/pkg/crd/schema_visitor.go
+++ b/pkg/crd/schema_visitor.go
@@ -44,16 +44,37 @@ type schemaWalker struct {
 	visitor SchemaVisitor
 }
 
-// walkSchema walks the given schema, saving modifications made by the
-// visitor (this is as simple as passing a pointer in most cases,
-// but special care needs to be taken to persist with maps).
+// walkSchema walks the given schema, saving modifications made by the visitor
+// (this is as simple as passing a pointer in most cases, but special care
+// needs to be taken to persist with maps).  It also visits referenced
+// schemata, dealing with circular references appropriately.  The returned
+// visitor will be used to visit all "children" of the current schema, followed
+// by a nil schema with the returned visitor to mark completion.  If a nil visitor
+// is returned, traversal will no continue into the children of the current schema.
 func (w schemaWalker) walkSchema(schema *apiext.JSONSchemaProps) {
-	subVisitor := w.visitor.Visit(schema)
-	if subVisitor == nil {
-		return
+	// Walk a potential chain of schema references, keeping track of seen
+	// references to avoid circular references
+	subVisitor := w.visitor
+	seenRefs := map[string]bool{}
+	if schema.Ref != nil {
+		seenRefs[*schema.Ref] = true
 	}
-	defer subVisitor.Visit(nil)
+	for {
+		subVisitor = subVisitor.Visit(schema)
+		if subVisitor == nil {
+			return
+		}
+		// mark completion of the visitor
+		defer subVisitor.Visit(nil)
 
+		// Break if schema is not a reference or a cycle is detected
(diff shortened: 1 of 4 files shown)


pkg/crd/schema_visitor.go
```diff
@@ -48,11 +48,18 @@ type schemaWalker struct {
...
 func (w schemaWalker) walkSchema(schema *apiext.JSONSchemaProps) {
-	subVisitor := w.visitor.Visit(schema)
-	if subVisitor == nil {
-		return
+	// Walk a potential chain of schema references
+	subVisitor := w.visitor
+	for {
+		subVisitor = subVisitor.Visit(schema)
+		if subVisitor == nil {
+			return
```
DirectXMan12, line 56: can you add a test for this?

Make sure to have one for circular references too

skaslev: Added a test that on current master produces

```
    typeRefChain:
      $ref: '#/definitions/testdata.kubebuilder.io~1cronjob~1api~0Integer'
```

but works correctly with this PR.

Also added cycle detection logic so that this `for` terminates for circular references.
I couldn't come up with circular test that would compile though.

skaslev:
@DirectXMan12 Thank you for the review. 

Can you take another look?

pkg/crd/schema_visitor.go
skaslev: Found a simpler reproducer from #324 and switched to using it instead.

DirectXMan12:
pushed an extra test, plus a test for circular references

/lgtm
/approve

skaslev:
Thanks for doing this. Seems that it didn't land because of a misspelling in the test:

```
pkg/crd/flatten_type_test.go:112:35: `ciruclar` is a misspelling of `circular` (misspell)
		It("should not infinite-loop on ciruclar references", func() {
```

DirectXMan12:
gah, thanks for fixing the typo

/lgtm

PR #443, opened 2020-05-21
✨  Warn about things that we strongly recommend avoiding

ksaritek:
I created a guestbook kubebuilder init project and added a *ast.StarExpr to guestbook struct

type Env struct {
	A int    `json:"a"`
	B string `json:"b"`
}

// Guestbook is the Schema for the guestbooks API
type Guestbook struct {

	Spec   GuestbookSpec   `json:"spec,omitempty"`
	Status GuestbookStatus `json:"status,omitempty"`
	Envs   map[string]*Env `json:"envs,omitempty"`
}
my patch:

./controller-gen crd object:headerFile="hack/boilerplate.go.txt" paths="./..." output:crd:stdout

---
metadata:
  annotations:
  creationTimestamp: null
  name: guestbooks.webapp.my.domain
spec:
  group: webapp.my.domain
  names:
    kind: Guestbook
    listKind: GuestbookList
    plural: guestbooks
    singular: guestbook
  scope: Namespaced
  validation:
    openAPIV3Schema:
      description: Guestbook is the Schema for the guestbooks API
      properties:
        apiVersion:
          type: string
        envs:
          additionalProperties:
            properties:
              a:
                type: integer
              b:
                type: string
            required:
            - a
            - b
            type: object
          type: object
        kind:
          type: string
        metadata:
          type: object
        spec:
          description: GuestbookSpec defines the desired state of Guestbook
          properties:
            foo:
              description: Foo is an example field of Guestbook. Edit Guestbook_types.go
                to remove/update
              type: string
          type: object
        status:
          description: GuestbookStatus defines the observed state of Guestbook
          type: object
      type: object
  version: v1
  versions:
  - name: v1
    served: true
    storage: true
status:
  acceptedNames:
    kind: ""
    plural: ""
  conditions: null
  storedVersions: null
/go/src/example/api/v1/guestbook_types.go:55:20: map values should be a named type, not *ast.StarExpr
controller-gen prints warning at the end like /go/src/example/api/v1/guestbook_types.go:55:20: map values should be a named type, not *ast.StarExpr.

### DIFF
--- a/pkg/crd/flatten.go
+++ b/pkg/crd/flatten.go
@@ -37,6 +37,15 @@ type ErrorRecorder interface {
 	AddError(error)
 }
 
+// WarningRecorder knows how to record errors. It wraps the part of
+// pkg/loader.Package that we need to record errors in places were it might not
+// make sense to have a loader.Package
+type WarningRecorder interface {
+	// AddWarning records that the given error occurred.
+	// See the documentation on loader.Package.AddWarning for more information.
+	AddWarning(error)
+}
+
 // isOrNil checks if val is nil if val is of a nillable type, otherwise,
 // it compares val to valInt (which should probably be the zero value).
 func isOrNil(val reflect.Value, valInt interface{}, zeroInt interface{}) bool {
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -304,6 +304,7 @@ func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *apiext.JSONSchemaPro
 			return &apiext.JSONSchemaProps{}
 		}
 	case *ast.StarExpr:
+		ctx.pkg.AddWarning(loader.ErrFromNode(fmt.Errorf("map values should be a named type, not %T", mapType.Value), mapType.Value))
 		valSchema = typeToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
 	default:
 		ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map values must be a named type, not %T", mapType.Value), mapType.Value))
--- a/pkg/loader/loader.go
+++ b/pkg/loader/loader.go
@@ -25,6 +25,7 @@ import (
 	"go/types"
 	"io/ioutil"
 	"os"
+	"sort"
 	"sync"
 
 	"golang.org/x/tools/go/packages"
@@ -40,17 +41,14 @@ import (
 // any errors of the kinds specified in filterKinds.  It will
 // return true if any errors were printed.
 func PrintErrors(pkgs []*Package, filterKinds ...packages.ErrorKind) bool {
-	pkgsRaw := make([]*packages.Package, len(pkgs))
-	for i, pkg := range pkgs {
-		pkgsRaw[i] = pkg.Package
-	}
+
 	toSkip := make(map[packages.ErrorKind]struct{})
 	for _, errKind := range filterKinds {
 		toSkip[errKind] = struct{}{}
 	}
 	hadErrors := false
-	packages.Visit(pkgsRaw, nil, func(pkgRaw *packages.Package) {
-		for _, err := range pkgRaw.Errors {
+	visit(pkgs, nil, func(pkg *Package) {
+		for _, err := range pkg.Errors {
 			if _, skip := toSkip[err.Kind]; skip {
 				continue
 			}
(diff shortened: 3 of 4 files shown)


DirectXMan12:
First comment: the beginning of the commit should generally say "what" and "why".  Having examples is fine, but it's generally go to snip to the relevant parts, and to used fenced code blocks.  I'd expect something like:

    We have a set of patterns that we reccomend avoiding (since they're not in
    line with the Kubernetes API conventions/best practices), but for legacy
    reasons can't make a hard error.  This introduces a mechanism to warn about
    such patterns when they're detected.
    
    It functions similarly to the way errors (an `AddWarning` on `Packaged` with
    special handling for `PositionedError`), except that such warnings are not
    considered fatal at the end of processing.
    
    
    This commit only introduces one such warning: warn about map values being
    pointers.
    
    For example:
    
    ```go
    type Env struct {
        A int json:"a"
        B string json:"b"
    }
    
    type Container struct {
        Envs   map[string]*Env `json:"envs,omitempty"`
    }
    ```
    
    yields
    
    ```
    /path/to/types.go:7:20: map values should be a named type, not *ast.StarExpr
    ```

pkg/crd/flatten.go
```diff
@@ -37,6 +37,15 @@ type ErrorRecorder interface {
 	AddError(error)
 }
 
+// WarningRecorder knows how to record errors. It wraps the part of
+// pkg/loader.Package that we need to record errors in places were it might not
+// make sense to have a loader.Package
```
DirectXMan12, line 42: the second sentence here seems like it was copy-pasted verbatim, and doesn't make much sense any more.
I'd go with

```suggestion
// WarningRecorder knows how to record warnings, which are effectively
// non-fatal errors.  It wraps the part of pkg/loader.Package that we need to
// record warnings in places were it might not make sense to have a
// loader.Package.
```

pkg/crd/flatten.go
```diff
@@ -37,6 +37,15 @@ type ErrorRecorder interface {
...
+type WarningRecorder interface {
+	// AddWarning records that the given error occurred.
```
DirectXMan12, line 44:
```suggestion
	// AddWarning records that the given warning (non-fatal error) occurred.
```

pkg/crd/schema.go
```diff
@@ -304,6 +304,7 @@ func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *apiext.JSONSchemaPro
 			return &apiext.JSONSchemaProps{}
 		}
 	case *ast.StarExpr:
+		ctx.pkg.AddWarning(loader.ErrFromNode(fmt.Errorf("map values should be a named type, not %T", mapType.Value), mapType.Value))
```
DirectXMan12, line 307: maybe tweak the message here a bit so it's obvious it's a warning.  I'd do something like

```
as per the Kubernetes API conventions, map values *should* generally be primitives or named types that alias primitives, not %T.  Consider using the "list as map" pattern here, except in cases where map is used for compatiblity/legacy purposes.
```

pkg/loader/loader.go
```diff
@@ -61,6 +59,52 @@ func PrintErrors(pkgs []*Package, filterKinds ...packages.ErrorKind) bool {
 	return hadErrors
 }
 
+// PrintWarnings print errors associated with all packages
```
DirectXMan12, line 62:
```suggestion
// PrintWarnings print warnings associated with all packages
```

pkg/loader/loader.go
```diff
@@ -61,6 +59,52 @@ func PrintErrors(pkgs []*Package, filterKinds ...packages.ErrorKind) bool {
...
+	})
+}
+
+// Visit visits all the packages in the import graph whose roots are
+// pkgs, calling the optional pre function the first time each package
+// is encountered (preorder), and the optional post function after a
+// package's dependencies have been visited (postorder).
+// The boolean result of pre(pkg) determines whether
+// the imports of package pkg are visited.
+func visit(pkgs []*Package, pre func(*Package) bool, post func(*Package)) {
```
DirectXMan12, line 79: do we ever use `pre`?  If not, we can get rid of it.

pkg/loader/loader.go
```diff
@@ -70,6 +114,10 @@ func PrintErrors(pkgs []*Package, filterKinds ...packages.ErrorKind) bool {
 type Package struct {
 	*packages.Package
 
+	// Warnings contains any errors encountered querying the metadata
+	// of the package, or while parsing or type-checking its files.
```
DirectXMan12, line 118:
```suggestion
	// Warnings contains any warnings ("non-fatal" errors) encountered while type-checking or
	// generating output (OpenAPI, etc) from the given package.
```

pkg/loader/loader.go
```diff
@@ -175,6 +223,25 @@ func (p *Package) AddError(err error) {
 	}
 }
 
+// AddWarning adds an error to the warningss associated with the given package.
```
DirectXMan12, line 226:
```suggestion
// AddWarning adds a warning (represented as an error) to the warnings associated with the given package.
// It functions nearly-identically to `AddError`, except that it adds to `Package.Warnings`.
```

Issue #445, opened 2020-05-24
Strip default values from v1beta1 CRD yaml


tamalsaha:
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting says

`To use defaulting, your CustomResourceDefinition must use API version apiextensions.k8s.io/v1.`

So, the default values should be stripped from v1beta1 CRD yaml.

PR #446, opened 2020-05-25
:bug: Remove defaults from crd v1beta1

tamalsaha:
Fixes https://github.com/kubernetes-sigs/controller-tools/issues/445

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -115,6 +115,13 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 			if err != nil {
 				return err
 			}
+			if ver == "v1beta1" {
+				obj := conv.(*apiextlegacy.CustomResourceDefinition)
+				if obj.Spec.Validation != nil &&
+					obj.Spec.Validation.OpenAPIV3Schema != nil {
+					removeDefaults(obj.Spec.Validation.OpenAPIV3Schema)
+				}
+			}
 			versionedCRDs[i] = conv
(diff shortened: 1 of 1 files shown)


vincepri:
It seems a little strange to me that we'd silently remove the defaults markers for an unsupported type, this could lead into confusion, thoughts @alvaroaleman @DirectXMan12 ?

pkg/crd/gen.go
```diff
@@ -182,6 +189,59 @@ func toTrivialVersions(crd *apiextlegacy.CustomResourceDefinition) {
 	crd.Spec.AdditionalPrinterColumns = canonicalColumns
 }
 
+// removeDefaults removes defaults from apiextensions.k8s.io/v1beta1 CRD definition.
+// To use defaulting, CustomResourceDefinitions must use API version apiextensions.k8s.io/v1
+func removeDefaults(schema *apiextlegacy.JSONSchemaProps) {
```
DirectXMan12, line 194: we've got a schema visitor for this purpose, so you shouldn't have to manually write a schema visitor.

DirectXMan12 (changes requested):
left comment, needs a test case, also we should probably warn about this, or maybe just error out.
I'd lean towards just erroring out, but I wouldn't be surprised if people have workarounds (deploying different controller versions or webhooks but same YAML), so a warning is probably better.

DirectXMan12:
Closed by #481, but that'll need some fixups.

Issue #245, opened 2019-07-09
Support floats?


skonto:
In the official docs I still see float support: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/, am I wrong? 
I am currently hitting this error: https://github.com/kubernetes-sigs/controller-tools/blob/abcf7d8d54b0cca0f42c470b705fd3cfe768ede3/pkg/crd/schema.go#L387
Error I get is: `unsupported type "float32"`. I expected my field to be parsed and crd to have a field of type `number` in the yaml format.

DirectXMan12:
so, it's technically supported, but floats are *highly* discouraged (and disallowed in official kubernetes APIs) because they won't round-trip properly, can't actually represent certain numbers exactly, etc (all the normal reasons floats are usually not what you want in a program).  We recommend using `resource.Quantity` or `int32` for your numeric types, in accordance with the Kubernetes API conventions.

Closing this for now, but if you have a compelling argument, I'm happy to reconsider.

tothandras:
I think it would be better to support it and have a warning in the documentation about this. There are scenarios where precision is not important and any workaround might be worse.

DirectXMan12:
it's not just precision -- it's namely that it doesn't round-trip.  I'm can you provide a concrete example here of a scenario where you want to use float?

tothandras:
@DirectXMan12 We have an operator managing [canary releases](https://banzaicloud.com/blog/istio-canary/) (using Istio). The user should be able to provide arbitrary prometheus queries with thresholds (usually a floating number) alongside the built-in metrics (request success rate and request duration).

DirectXMan12:
I think that's probably a better case for quantity than for a normal float, but I see your point.

DirectXMan12:
we could also have an "allowDangerousTypes" option too.

tothandras:
@DirectXMan12 Thank you for considering it! From a UX perspective, by default I would log out a warning with a link to the relevant part of the documentation or a short description, but not fail to build. The `allowDangerousTypes` option could be used to turn off this warning.

gaocegege:
I am glad to share why we need float in CRD spec.

We are implementing an AutoML (a.k.a Hyperparameter tuning) system using kubebuilder. CRD is here https://github.com/kubeflow/katib/tree/master/pkg/apis/controller.

[...]

PR #93, opened 2018-10-10
Remove scaffold pkg and test project

mengqiy:

PR #449, opened 2020-06-02
✨ Add support for AllowDangerousTypes crd flag (enables float support)

matthchr:
This can be used to enable `float32`/`float64` (go type) -> `number` (json schema) for CRDs. It is off by default because floats are not recommended in CRDs.

Discussion about this originated in #245 (which is now closed). @DirectXMan12 had suggested using a flag by this name to allow conditional supporting of floats. There's also at least one other issue asking about this (indirectly) at #93.

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -53,6 +53,16 @@ type Generator struct {
 	// It's required to be false for v1 CRDs.
 	PreserveUnknownFields *bool `marker:",optional"`
 
+	// AllowDangerousTypes allows types which are usually omitted from CRD generation
+	// because they are not recommended.
+	//
+	// Currently the following additional types are allowed when this is true:
+	// float32
+	// float64
+	//
+	// Left unspecified, the default is false
+	AllowDangerousTypes *bool `marker:",optional"`
+
 	// MaxDescLen specifies the maximum description length for fields in CRD's OpenAPI schema.
 	//
 	// 0 indicates drop the description for all fields completely.
@@ -78,6 +88,8 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	parser := &Parser{
 		Collector: ctx.Collector,
 		Checker:   ctx.Checker,
+		// Perform defaulting here to avoid ambiguity later
+		AllowDangerousTypes: g.AllowDangerousTypes != nil && *g.AllowDangerousTypes == true,
 	}
 
 	AddKnownTypes(parser)
--- a/pkg/crd/parser.go
+++ b/pkg/crd/parser.go
@@ -73,6 +73,20 @@ type Parser struct {
 	packages map[*loader.Package]struct{}
 
 	flattener *Flattener
+
+	// AllowDangerousTypes controls the handling of non-recommended types such as float. If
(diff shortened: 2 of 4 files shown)


matthchr:
I signed it

DirectXMan12:
/hold

thanks for the contribution.  This is neat, and I like that it explicitly marks particular types as "this is a known problematic type".  I think it fits nicely with #443 as well, but I think we'll want to think about cases where we use one vs the other -- this feels like "types that we don't use" vs 443, which is "types that we shouldn't use", but I'm not sure how clear the distinction is.  Thoughts?

matthchr:
@DirectXMan12 - it seems like there's a continuum here:
1. Types that are always supported.
2. Types that are allowed by default, but not recommended (warning emitted when they are encountered as per #443). Possibly they are allowed by default for historical reasons and may even be "on their way out" at some point in the future.
3. Types that are not allowed by default, not recommended, but there are some legitimate reasons to need them in certain corner cases. Possibly these types should _also_ emit a warning as per #443 even when they are "switched on" (an integration point between this PR and #443 if desired.

I do think the two ideas actually play nicely together because if you wanted to say demote a type from category 2 to 3, you can do that and then hide it behind the `allowDangerousTypes` flag so that people can turn it back on if they so desire.

*I don't actually know what all types are "not recommended" and presumably the recommended practices are _relatively_ static so I wouldn't imagine much release-to-release change in these once they're solidified.

Thoughts?

matthchr:
@DirectXMan12 - Are there changes that you'd like to see for this PR (and/or its interactions with #443)?

I'm wondering what the process is to get from where it is today to a state where it could be merged.

DirectXMan12:
> I'm wondering what the process is to get from where it is today to a state where it could be merged.

It's largely mergable now, just wanted to make sure we had that conversation and that there was a general plan.  The one thing I'd do is maybe copy what you wrote in that comment (it's well-written :-) ) into a comment in the source code, and maybe add a TODO to make a more formal mechanism for putting "type patterns" in those categories.

DirectXMan12:
then we can get this merged

matthchr:
@DirectXMan12 thanks for the guidance. I've put the text I wrote as a comment on the `AllowDangerousTypes` field in `Parser.go` (I didn't want to put it in `gen.go` because that shows up in customer facing help and this seemed more of an internal thing than something we want in the help).

Let me know if that looks good or if you'd like any modifications to it - thanks for your time!

matthchr:
@DirectXMan12 thanks for the approval. Based on the [PR approval process](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#the-code-review-process) it looks like the `do-not-merge/hold` label is blocking automatic merging. If that's intentional no worries, just wanted to check though as the label may have been left on unintentionally.

DirectXMan12:
ah, yeah, I forgot to

/hold cancel

DirectXMan12:
that should do it.

whalecold:
hi， how can I use the field `allowDangerousTypes` ?

matthchr:
Specify it on the cmdline, something like: `crd:crdVersions=v1,allowDangerousTypes=true`

PR #455, opened 2020-06-29
🐛 Correctly parse and set Scope

alexanderConstantinescu:
Depending on the existance of `genclient:nonNamespaced` set the CRD's scope value to the correct one. 

`genclient:nonNamespaced` is currently a supported field by `informer-gen` and all the likes. However `controller-gen` does not take the annotation into account and consistently overrides the `.Scope` with `Namespaced` whereas the presence of such a field such indicate and set that field to `Cluster`.

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -48,6 +48,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:unservedversion", markers.DescribesType, UnservedVersion{})).
 		WithHelp(UnservedVersion{}.Help()),
+
+	must(markers.MakeDefinition("genclient:nonNamespaced", markers.DescribesType, ClusterScope{})).
+		WithHelp(ClusterScope{}.Help()),
 }
 
 // TODO: categories and singular used to be annotations types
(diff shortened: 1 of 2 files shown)


alexanderConstantinescu:
@droot / @pwittrock : is this repo deprecated or something? Just curious as I don't see a lot of PRs merging here. Let me know, and I'll file it in the right place

pwittrock:
@alexanderConstantinescu The repo is not deprecated, but is neglected more than the controller-runtime and kubebuilder repos.  It could probably use some more ownership if you are looking to get involved in the project.

pwittrock:
@DirectXMan12 I still need to familiarize myself with the new codebase.  Does this look reasonable to you?

DirectXMan12:
We do already have a marker for this -- it's part of the `// +kubebuilder:resource` marker (https://book.kubebuilder.io/reference/markers/crd.html).  We try to otherwise only use generic markers from core.  It seems a bit weird to a) add a duplicate marker (we already have a couple), and b) to add a marker that's specific to a tool we don't support.

Can you clarify the motivation here a bit?

alexanderConstantinescu:
> We do already have a marker for this -- it's part of the `// +kubebuilder:resource` marker (https://book.kubebuilder.io/reference/markers/crd.html). We try to otherwise only use generic markers from core. It seems a bit weird to a) add a duplicate marker (we already have a couple), and b) to add a marker that's specific to a tool we don't support.
> 
> Can you clarify the motivation here a bit?

Thanks for the explanation. There's no motivation at all. I didn't read the documentation well enough and missed the `scope=<string>` part...

Closing this PR

PR #465, opened 2020-07-24
[WIP] ✨ Add openapiv3 generator

vincepri:

⚠️ This only works with CRDv1 definitions.


timothysc:
xref - https://github.com/kubernetes-sigs/kubebuilder/issues/1231

vincepri:
Marked as a WIP to see if this this is what we want/need, I'll update the `testdata` once the final spec of this generator is finalized

vincepri:
For folks that want to try it out:

- Checkout this PR branch, you can use `gh pr checkout 465` if you have `gh` cli installed
- Run `go get ./...` to build the CLI
- Run `controller-gen openapiv3:manifests=./config/crd/bases paths=./api/... output:dir=./manifests` in a kubebuilder project.

jjfeiler:
@vincepri I'm getting a SIGSEGV when I try to run this as you suggested above `controller-gen openapiv3:manifests=./config/crd/bases paths=./api/... output:dir=./manifests` because version.Schema is nil on pkg/openapiv3/gen.go:74. Not sure where that Schema should be coming from in this case.

PR #469, opened 2020-08-11
⚠ Add admissionregisteration/v1 support for webhook

### LINKED ISSUE #419: contoller-gen: webhook generator support for admissionregistration.k8s.io/v1
Hi folks,
Since `admissionregistration.k8s.io/v1beta1` is deprecated in kubernetes v1.16 in favor of `admissionregistration.k8s.io/v1`, do you have a plan for adding support to generate webhook configurations (Validating/MutatingWebhokConfiguration) for `admissionregistration.k8s.io/v1`?

I can work on this if it is needed.

jiachengxu:
This PR implements support to generate {Mutating, Validating}WebhookConfiguration with version `v1`.
Fixes #419

The first intention was to add a package marker `webhookVersions=<[]string>` for the user to specify the versions that want to generate. It will default to `v1` if the `webhookVersions` is not specified. However, after discussion with @estroz (https://github.com/kubernetes-sigs/controller-tools/pull/469#discussion_r469346611), we found out this cannot work nicely because, for different versions of WebhookConfiguraiton, Some fields (SideEffects, AdmissionReviewVersions) can have different available values, and if one value in that marker is invalid for one version then you can't generate the config for either version.
So we decided to use an optional field marker `+kubebuilder:webhookVersions=<[]string>` to specify the versions in the `+kubebuilder:webhookVersions` directly, it works as follows:
- We default to generate `v1` configuration if the marker is not specified.
- Generate `v1` if the marker is `+kubebuilder:webhook:webhookVersions=v1`
- Generate `v1beta1` if the marker is `+kubebuilder:webhook:webhookVersions=v1beta1`
- Generate both `v1` and `v1beta1` if the marker is `+kubebuilder:webhook:webhookVersions=v1;v1beta1` or `+kubebuilder:webhook:webhookVersions=v1beta1;v1`, we throw error out if the `sideEffects` (later we also needed to handle `AdmissionReviewVersions`) cannot meet requirement for either versions.

This is a breaking change since we generate `v1` version configurations by default if the marker is not specified.

### DIFF
--- a/pkg/webhook/conv.go
+++ b/pkg/webhook/conv.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	admissionregv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+var (
+	conversionScheme = runtime.NewScheme()
+)
+
+func init() {
+	utilruntime.Must(admissionregv1.AddToScheme(conversionScheme))
+	utilruntime.Must(admissionregv1beta1.AddToScheme(conversionScheme))
+}
+
+// MutatingWebhookConfigurationAsVersion converts a MutatingWebhookConfiguration from the canonical internal form (currently v1) to some external form.
+func MutatingWebhookConfigurationAsVersion(original *admissionregv1.MutatingWebhookConfiguration, gv schema.GroupVersion) (runtime.Object, error) {
+	return conversionScheme.ConvertToVersion(original, gv)
+}
+
+// ValidatingWebhookConfigurationAsVersion converts a ValidatingWebhookConfiguration from the canonical internal form (currently v1) to some external form.
+func ValidatingWebhookConfigurationAsVersion(original *admissionregv1.ValidatingWebhookConfiguration, gv schema.GroupVersion) (runtime.Object, error) {
+	return conversionScheme.ConvertToVersion(original, gv)
+}
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -19,18 +19,25 @@ limitations under the License.
 //
 // The markers take the form:
 //
-//  +kubebuilder:webhook:failurePolicy=<string>,matchPolicy=<string>,groups=<[]string>,resources=<[]string>,verbs=<[]string>,versions=<[]string>,name=<string>,path=<string>,mutating=<bool>,sideEffects=<string>
+//  +kubebuilder:webhook:webhookVersions=<[]string>,failurePolicy=<string>,matchPolicy=<string>,groups=<[]string>,resources=<[]string>,verbs=<[]string>,versions=<[]string>,name=<string>,path=<string>,mutating=<bool>,sideEffects=<string>
 package webhook
 
 import (
 	"fmt"
 	"strings"
 
-	admissionreg "k8s.io/api/admissionregistration/v1beta1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"sigs.k8s.io/controller-tools/pkg/genall"
 	"sigs.k8s.io/controller-tools/pkg/markers"
+)
+
+// The default {Mutating,Validating}WebhookConfiguration version to generate.
+const (
+	defaultWebhookVersion = "v1"
 )
 
 var (
@@ -38,6 +45,11 @@ var (
 	// Call ToWebhook on the value to get a Kubernetes Webhook.
 	ConfigDefinition = markers.Must(markers.MakeDefinition("kubebuilder:webhook", markers.DescribesPackage, Config{}))
 )
+
+// supportedWebhookVersions returns currently supported API version of {Mutating,Validating}WebhookConfiguration.
+func supportedWebhookVersions() []string {
+	return []string{defaultWebhookVersion, "v1beta1"}
+}
 
 // +controllertools:marker:generateHelp:category=Webhook
 
@@ -91,79 +103,89 @@ type Config struct {
 	// batch.tutorial.kubebuilder.io/v1,Kind=CronJob would be
 	// /validate-batch-tutorial-kubebuilder-io-v1-cronjob
 	Path string
+
+	// WebhookVersions specifies the target API versions of the {Mutating,Validating}WebhookConfiguration objects
+	// itself to generate.  Defaults to v1.
+	WebhookVersions []string `marker:"webhookVersions,optional"`
 }
 
 // verbToAPIVariant converts a marker's verb to the proper value for the API.
 // Unrecognized verbs are passed through.
-func verbToAPIVariant(verbRaw string) admissionreg.OperationType {
+func verbToAPIVariant(verbRaw string) admissionregv1.OperationType {
 	switch strings.ToLower(verbRaw) {
-	case strings.ToLower(string(admissionreg.Create)):
-		return admissionreg.Create
-	case strings.ToLower(string(admissionreg.Update)):
-		return admissionreg.Update
-	case strings.ToLower(string(admissionreg.Delete)):
-		return admissionreg.Delete
-	case strings.ToLower(string(admissionreg.Connect)):
-		return admissionreg.Connect
-	case strings.ToLower(string(admissionreg.OperationAll)):
-		return admissionreg.OperationAll
+	case strings.ToLower(string(admissionregv1.Create)):
+		return admissionregv1.Create
+	case strings.ToLower(string(admissionregv1.Update)):
+		return admissionregv1.Update
+	case strings.ToLower(string(admissionregv1.Delete)):
+		return admissionregv1.Delete
+	case strings.ToLower(string(admissionregv1.Connect)):
+		return admissionregv1.Connect
+	case strings.ToLower(string(admissionregv1.OperationAll)):
+		return admissionregv1.OperationAll
 	default:
-		return admissionreg.OperationType(verbRaw)
+		return admissionregv1.OperationType(verbRaw)
 	}
 }
 
 // ToMutatingWebhook converts this rule to its Kubernetes API form.
-func (c Config) ToMutatingWebhook() (admissionreg.MutatingWebhook, error) {
+func (c Config) ToMutatingWebhook() (admissionregv1.MutatingWebhook, error) {
 	if !c.Mutating {
-		return admissionreg.MutatingWebhook{}, fmt.Errorf("%s is a validating webhook", c.Name)
+		return admissionregv1.MutatingWebhook{}, fmt.Errorf("%s is a validating webhook", c.Name)
 	}
 
 	matchPolicy, err := c.matchPolicy()
 	if err != nil {
-		return admissionreg.MutatingWebhook{}, err
+		return admissionregv1.MutatingWebhook{}, err
 	}
 
-	return admissionreg.MutatingWebhook{
+	return admissionregv1.MutatingWebhook{
 		Name:          c.Name,
 		Rules:         c.rules(),
 		FailurePolicy: c.failurePolicy(),
 		MatchPolicy:   matchPolicy,
 		ClientConfig:  c.clientConfig(),
 		SideEffects:   c.sideEffects(),
+		// TODO(jiachengxu): AdmissionReviewVersions becomes required in admissionregistration/v1, here we default it
+		// to `v1` and `v1beta1`, and we should support to config the `AdmissionReviewVersions` as a marker.
+		AdmissionReviewVersions: []string{defaultWebhookVersion, "v1beta1"},
 	}, nil
 }
 
 // ToValidatingWebhook converts this rule to its Kubernetes API form.
-func (c Config) ToValidatingWebhook() (admissionreg.ValidatingWebhook, error) {
+func (c Config) ToValidatingWebhook() (admissionregv1.ValidatingWebhook, error) {
 	if c.Mutating {
-		return admissionreg.ValidatingWebhook{}, fmt.Errorf("%s is a mutating webhook", c.Name)
+		return admissionregv1.ValidatingWebhook{}, fmt.Errorf("%s is a mutating webhook", c.Name)
 	}
 
 	matchPolicy, err := c.matchPolicy()
 	if err != nil {
-		return admissionreg.ValidatingWebhook{}, err
+		return admissionregv1.ValidatingWebhook{}, err
 	}
 
-	return admissionreg.ValidatingWebhook{
+	return admissionregv1.ValidatingWebhook{
 		Name:          c.Name,
 		Rules:         c.rules(),
 		FailurePolicy: c.failurePolicy(),
 		MatchPolicy:   matchPolicy,
 		ClientConfig:  c.clientConfig(),
 		SideEffects:   c.sideEffects(),
+		// TODO(jiachengxu): AdmissionReviewVersions becomes required in admissionregistration/v1, here we default it
+		// to `v1` and `v1beta1`, and we should support to config the `AdmissionReviewVersions` as a marker.
+		AdmissionReviewVersions: []string{defaultWebhookVersion, "v1beta1"},
 	}, nil
 }
 
 // rules returns the configuration of what operations on what
 // resources/subresources a webhook should care about.
-func (c Config) rules() []admissionreg.RuleWithOperations {
-	whConfig := admissionreg.RuleWithOperations{
-		Rule: admissionreg.Rule{
+func (c Config) rules() []admissionregv1.RuleWithOperations {
+	whConfig := admissionregv1.RuleWithOperations{
+		Rule: admissionregv1.Rule{
 			APIGroups:   c.Groups,
 			APIVersions: c.Versions,
 			Resources:   c.Resources,
 		},
-		Operations: make([]admissionreg.OperationType, len(c.Verbs)),
+		Operations: make([]admissionregv1.OperationType, len(c.Verbs)),
 	}
 
 	for i, verbRaw := range c.Verbs {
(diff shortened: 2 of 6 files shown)


pkg/webhook/parser.go
```diff
@@ -130,6 +134,9 @@ func (c Config) ToMutatingWebhook() (admissionreg.MutatingWebhook, error) {
 		MatchPolicy:   matchPolicy,
 		ClientConfig:  c.clientConfig(),
 		SideEffects:   c.sideEffects(),
+		// TODO(jiachengxu): AdmissionReviewVersions becomes required in admissionregistration/v1, here we default it
+		// to `v1`, and we should support to config the `AdmissionReviewVersions` as a marker.
+		AdmissionReviewVersions: []string{defaultVersion},
```
jiachengxu, line 139: I can implement this `AdmissionReviewVersions` as a marker by following a PR, don't want to do this in this PR since this PR is already large though :)

pkg/webhook/testdata/webhook.go
```diff
@@ -27,7 +27,7 @@ func (c *CronJob) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-// +kubebuilder:webhook:verbs=create;update,path=/validate-testdata-kubebuilder-io-v1-cronjob,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=testdata.kubebuiler.io,resources=cronjobs,versions=v1,name=validation.cronjob.testdata.kubebuilder.io,sideEffects=Some
+// +kubebuilder:webhook:verbs=create;update,path=/validate-testdata-kubebuilder-io-v1-cronjob,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=testdata.kubebuiler.io,resources=cronjobs,versions=v1,name=validation.cronjob.testdata.kubebuilder.io,sideEffects=NoneOnDryRun
```
jiachengxu, line 30: This change is because in `admissionregistration/v1`, the sideEffects can only be `None` and `NoneOnDryRun`

pkg/webhook/parser.go
```diff
@@ -238,7 +248,10 @@ func (c Config) sideEffects() *admissionreg.SideEffectClass {
 	case strings.ToLower(string(admissionreg.SideEffectClassSome)):
```
jiachengxu, line 248: In `admissionregistration/v1`, the sideEffects can only be `None` and `NoneOnDryRun`.
In the current implementation, users can still specify the sideEffects as `Some` or `Unknown` when genreate `v1` webhook configurations, but they will be rejectd by kubernetes api server.
Is that sufficient to handle it like this? I would like not to cut the above switch branches since we still need `Some` and `Unkown` for v1beta1. We can remove that when we drop the support for v1beta1.

jiachengxu: I can also add a check when generate `v1` version configurations, if it is set to `Some` or `Unknown` we default it to `None`
WDYT? @alvaroaleman @vincepri

pkg/webhook/conv.go
```diff
@@ -0,0 +1,47 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	admissionreg "k8s.io/api/admissionregistration/v1"
```
alvaroaleman, line 20: Could you suffix all `admissionreg` imports with their corresponding version to make the code more obvious? (i.E. `admissionregv1` here)

pkg/webhook/conv.go
```diff
@@ -0,0 +1,47 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	conversionScheme = runtime.NewScheme()
+)
+
+func init() {
+	if err := admissionreg.AddToScheme(conversionScheme); err != nil {
```
alvaroaleman, line 31: More for my understanding, why did we not need this before? Also, if you panic its a lot nicer for everyone involved to include the error in the panic :)

pkg/webhook/parser.go
alvaroaleman: Can we just error out if its set to `Some` or `Unknown` rather than silently overwriting it? Users might not be aware that that got changed

pkg/webhook/parser_integration_test.go
```diff
@@ -80,6 +82,23 @@ var _ = Describe("Webhook Generation From Parsing to CustomResourceDefinition",
...
+		By("loading the desired v1 YAML")
+		expectedFile, err = ioutil.ReadFile("manifests.v1.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		expectedMutating, expectedValidating = unmarshalBoth(expectedFile)
+
+		By("comparing the two")
+		assertSame = func(actual, expected interface{}) {
+			ExpectWithOffset(1, actual).To(Equal(expected), "type not as expected, check pkg/webhook/testdata/README.md for more details.\n\nDiff:\n\n%s", cmp.Diff(actual, expected))
+		}
+		assertSame(actualMutating, expectedMutating)
```
alvaroaleman, line 100: Would it maybe make sense to add an `envtest` based test to verify that the kube apiserver actualy accepts these?

pkg/webhook/parser.go
jiachengxu: @alvaroaleman How about we didn't do anything for that? The reasons are as following:
1. `Some` and `Unknown` are still valid values for `SideEffects` for version `v1`, they can not be set directly when you create them. But if the user creates a webhookconfiguration with version `v1beta1` and with `Some` or `Unknown`, and when it is converted to version `v1` by apiserver, the Values are still valid. https://github.com/kubernetes/api/blob/master/admissionregistration/v1/types.go#L277
2. IMHO, we don't need even to overwrite it or do anything about it. If the user tries to apply the configuration to Kubernetes, it will be rejected, and then the user knows it should be `None` or `NoneOnDryRun`. and then they can update the marker and regenerate.  

WDYT?

pkg/webhook/conv.go
jiachengxu: This is not needed before because we only support a single version of WebhookConfiguration, we just created the versioned `v1beta1` objects in the parser.go, and be done with it.
Now we have 2 versions, so we can let the scheme to handle conversion between them.

I will update the panic :) I didn't take look that closely, and copied that from crd package :)

pkg/webhook/parser_integration_test.go
jiachengxu: Yes, sure, do we have any examples for this in our codebase?

pkg/webhook/parser.go
alvaroaleman: sgtm

pkg/webhook/parser_integration_test.go
jiachengxu: Oh, I see. I can use the `Paths` in the `WebhookInstallOptions` https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/webhook.go#L44 to register to apiserver, and then do `env.Start()` to start the apiserver to see if there are any errors
@alvaroaleman Is this what you meant?

pkg/webhook/parser.go
estroz: This should probably include "v1beta1" since controller-runtime uses v1beta1 admission reviews right?

pkg/webhook/testdata/cronjob_types.go
```diff
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-//go:generate ../../../.run-controller-gen.sh webhook paths=. output:dir=.
+//go:generate ../../../.run-controller-gen.sh webhook:webhookVersions={v1beta1,v1} paths=. output:dir=.
```
estroz, line 16: Since v1 is now the default:
```suggestion
//go:generate ../../../.run-controller-gen.sh webhook:webhookVersions={v1,v1beta1} paths=. output:dir=.
```

pkg/webhook/parser_integration_test.go
alvaroaleman: Either that or do a `Create` after its started, that way the error might be more helpful, if any.

pkg/webhook/parser.go
jiachengxu: @estroz Yeah, you are right, we should add `v1beta1` here also.

pkg/webhook/conv.go
jiachengxu: Done

pkg/webhook/conv.go
jiachengxu: Done

pkg/webhook/testdata/cronjob_types.go
jiachengxu: Fixed

pkg/webhook/parser_integration_test.go
jiachengxu: Working on this :)

pkg/webhook/parser.go
```diff
@@ -228,15 +238,15 @@ func (c Config) clientConfig() admissionreg.WebhookClientConfig {
...
-		sideEffects = admissionreg.SideEffectClassSome
+	case strings.ToLower(string(admissionregv1.SideEffectClassNone)):
+		sideEffects = admissionregv1.SideEffectClassNone
+	case strings.ToLower(string(admissionregv1.SideEffectClassNoneOnDryRun)):
+		sideEffects = admissionregv1.SideEffectClassNoneOnDryRun
+	case strings.ToLower(string(admissionregv1.SideEffectClassSome)):
+		sideEffects = admissionregv1.SideEffectClassSome
 	case "":
 		return nil
 	default:
```
estroz, line 252: Nit: combine the `””` and default cases into one default that returns `None`, since this is a required value for v1 configs. Then you can remove this for v1beta1 in the same loop with `AdmissionReviewVersions` below.

pkg/webhook/parser.go
```diff
@@ -279,37 +296,92 @@ func (Generator) Generate(ctx *genall.GenerationContext) error {
...
+		},
+		Webhooks: validatingCfgs,
 	}
 
-	if len(objs) > 0 {
-		return ctx.WriteYAML("manifests.yaml", objs...)
+	versionedWebhooks := make([][]interface{}, len(webhookVersions))
+	for i, ver := range webhookVersions {
+		if len(mutatingCfgs) > 0 {
+			objRaw := mutatingWebhookConfiguration.DeepCopy()
```
estroz, line 328: These don’t need to be copied.

pkg/webhook/parser_integration_test.go
```diff
@@ -80,6 +82,23 @@ var _ = Describe("Webhook Generation From Parsing to CustomResourceDefinition",
...
+		Expect(err).NotTo(HaveOccurred())
+		actualMutating, actualValidating = unmarshalBoth(actualFile)
+
+		By("loading the desired v1beta1 YAML")
+		expectedFile, err = ioutil.ReadFile("manifests.v1beta1.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		expectedMutating, expectedValidating = unmarshalBoth(expectedFile)
+
+		By("comparing the two")
+		assertSame = func(actual, expected interface{}) {
```
estroz, line 97: `assertSame` is already defined above. Perhaps define this function globally.

pkg/webhook/parser_integration_test.go
estroz: I would be ok with adding envtest in a follow-up.

pkg/webhook/testdata/webhook.go
estroz: Are you checking for this case anywhere? There should be an error if anything other than either of those two is set.

pkg/webhook/parser.go
jiachengxu: I don't think this will work, because the default value for `v1beta1` is `Unknown` if we default it to `None` here, in the loop below, we can not distinguish if user set it to `None` or the user didn't set it.

jiachengxu: I added a check in the for loop to default it to `None` for `v1`

pkg/webhook/parser.go
jiachengxu: This is needed I think, because we are modifying the webhookConfiguration in the for loop, for example, we remove the `AdmissionVersions` for `v1beta1`, and if `v1beta1` comes first before `v1`, then `AdmissionVersions` is removed in the first iteration, it will break the generation for `v1`

pkg/webhook/parser_integration_test.go
jiachengxu: @alvaroaleman I have tried to add the `envtest` based test, finally, it works but still some issues around it:
1. We need to add `sigs.k8s.io/controller-runtime` as a dependency. (If we require v0.6.2, then we need to bump k8s version)
2. We need to upgrade the k8s version in test.sh: https://github.com/kubernetes-sigs/controller-tools/blob/master/test.sh#L41 v1.11.0 is too old and `v1` version WebhookConfiguration is not available, and will receive something like this:
```
  Unexpected error:
      <*meta.NoKindMatchError | 0xc0009c5100>: {
          GroupKind: {
              Group: "admissionregistration.k8s.io",
              Kind: "MutatingWebhookConfiguration",
          },
          SearchedVersions: ["v1"],
      }
      no matches for kind "MutatingWebhookConfiguration" in version "admissionregistration.k8s.io/v1"
```
Do you think it is still necessary to do this in this PR? Or we regard this as a follow-up, and I can make small PRs to bump k8s version, upgrade k8s version in the test.sh, and add test?

pkg/webhook/parser_integration_test.go
jiachengxu: Fixed

pkg/webhook/testdata/webhook.go
jiachengxu: No, there is no such check, please check this comment: https://github.com/kubernetes-sigs/controller-tools/pull/469#discussion_r468598242

pkg/webhook/parser_integration_test.go
alvaroaleman: Follow-Up is fine

pkg/webhook/parser.go
estroz: Ah yes you're right.

pkg/webhook/testdata/webhook.go
estroz: In general I would want as much static validation as possible for my operator project so that I don't have to wait for my CI's apiserver to tell me my webhook config is invalid when I upgrade from v1beta1 to v1. Is there a good reason not to statically validate side effects?

estroz: I guess the counter-argument is that having one marker for two versions is nice, and if one value in that marker is invalid for one version then you can't generate the config for either version, which I think is what you're suggesting in your comment.

The right solution IMO is to add a `+kubebuilder:webhook:webhookVersions=<list>` field so you can specify which version to generate for a marker if you want to generate both and configure them differently. Thoughts?

jiachengxu: > I guess the counter-argument is that having one marker for two versions is nice, and if one value in that marker is invalid for one version then you can't generate the config for either version, which I think is what you're suggesting in your comment.

Yeah, exactly. This is my main concern so I decided not to handle the static validation and let kube-apiserver handle it, but it may not be a good solution because users have to modify the configuration every time they re-generate it.

> The right solution IMO is to add an optional +kubebuilder:webhook:webhookVersion=<v1|v1beta1> field so you can specify which version to generate for a marker if you want to generate both and configure them differently. 

I think this will work, but it will double the number of markers if the user wants to generate both `v1` and `v1beta1` webhooks. How about we do the following, with a marker `+kubebuilder:webhook:webhookVersions=<[]string>`:
- We default to generate `v1` configuration if the marker is not specified.
- Generate `v1` if the marker is `+kubebuilder:webhook:webhookVersions=v1`
- Generate `v1beta1` if the marker is `+kubebuilder:webhook:webhookVersions=v1beta1`
- Generate both `v1` and `v1beta1` if the marker is `+kubebuilder:webhook:webhookVersions=v1;v1beta1` or `+kubebuilder:webhook:webhookVersions=v1beta1;v1`, we throw error out if the `sideEffects` (later we also needed to handle `AdmissionReviewVersions`) cannot meet requirement for either versions.
In this way, we still provide a chance for the user to generate both `v1` and `v1beta1` in one marker. 

> The case where this field wasn't added to the marker is the current case, so no version list is needed and this isn't a breaking change. 

I think it will still be a breaking change because we will generate `v1` configurations by default (currently `v1beta1`)?

WDYT @estroz ?

estroz: Works for me.

>this isn't a breaking change.

To the marker definition I mean, since current markers will work as-is.

jiachengxu: @estroz This has been implemented, PTAL :)

pkg/webhook/testdata/manifests.v1beta1.yaml
```diff
@@ -0,0 +1,56 @@
+
```
camilamacedo86, line 1: just a nit: we need to `spec.preserveUnknownFields` defaults to true for v1beta in order to allow it works 1.11+ and 1.16+ as well. See: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#customresourcedefinitionspec-v1beta1-apiextensions-k8s-io

jiachengxu:
kindly ping @vincepri , can you please take a look?

vincepri:
Should we merge this after we release v0.3.1, I guess this would be v0.4.0?

vincepri:
Actually it seems we already merged #468, which is a breaking change, so this can go in and the next release is v0.4.0

jiachengxu:
kindly ping @alvaroaleman , do you have any further comments? If not, I will just unhold the PR, and rock and roll :)

caboteria:
The problem with making `v1` the default is that controller-runtime doesn't yet handle `v1` so now controller-gen by default generates manifests that don't work with controller-runtime.

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/webhook.go#L44

estroz:
@caboteria this was fixed in #474, which forces you to set `admissionReviewVersions={<version>}`, in this case `v1beta1`.

estroz:
@vincepri there should probably be another release for this, but since it breaks the v0.4.0 marker scheme, should it be v0.5.0?

caboteria:
Thanks @estroz for the pointer, that helps! I'm "go get"ing the current master rev of controller-tools so I'm all set, but my concern is that the defaults for controller-tools don't work with the current controller-runtime so it's easy for noobs like me to stub their toe when trying to build an operator. Should v1beta1 be the default until controller-runtime works with v1?

estroz:
>the defaults for controller-tools don't work with the current controller-runtime

There is no `admissionReviewVersion` default anymore (logic for removing the default is [here](https://github.com/kubernetes-sigs/controller-tools/pull/474#discussion_r475050862)), you are forced to set that marker field. If you try running `controller-gen webhook` without that field set, you'll get a fairly indicative error. Does that answer your question?

jiachengxu:
For making controller-runtime work with `admissionReviewVersion` `v1`, we should fix this:https://github.com/kubernetes-sigs/controller-runtime/issues/1123

caboteria:
Thanks @estroz, the current master version is better in that respect than 0.4.0 which happily generates manifests that deploy but cause errors at run-time:

```
2020-10-28T15:18:03.677Z	ERROR	controller-runtime.webhook.webhooks	unable to decode the request	{"webhook": "/validate-egw-acnodal-io-v1-loadbalancer", "error": "no kind \"AdmissionReview\" is registered for version \"admission.k8s.io/v1\" in scheme \"pkg/runtime/scheme.go:101\""}
```
I guess one person's "fairly indicative" is another's "wtf"...

Don't get me wrong, I appreciate the effort that you all are making to move controller-tools forward and I second both the suggestion to release controller-tools 0.5.0 and to make controller-runtime work with v1. At the moment we're in a situation where it's easy for noobs (like me) to waste time having to debug things that should just work.

estroz:
>I guess one person's "fairly indicative" is another's "wtf"...

:rofl: I meant the static errors your see when generating admission webhooks for markers that don't have `admissionReviewVersions={<versions>}`:
```console
$ controller-gen webhook paths=./...
AdmissionReviewVersions is mandatory for v1 {Mutating,Validating}WebhookConfiguration
```

The deploy-time errors you get when deploying a v1 webhook config without `admissionReviewVersions` aren't under the purview of controller-tools. Also you shouldn't be seeing these deploy-time errors when using a webhook config generated using the current master's `controller-gen` binary; if you don't specify `admissionReviewVersions={<versions>}` you'll see the static error from the example above and the `manifests.yaml` won't be generated.

jiachengxu:
@caboteria  A PR is created here: https://github.com/kubernetes-sigs/controller-runtime/pull/1233 to tackle this issue: https://github.com/kubernetes-sigs/controller-runtime/pull/1233

PR #470, opened 2020-08-11
✨ feat: parse negative integers

knight42:
Signed-off-by: knight42 <anonymousknight96@gmail.com>

This adds support for negative integers in the marker syntax parser.  This is useful for markers like `+kubebuilder:validation:Minimum`

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -106,7 +106,7 @@ func init() {
 type Maximum int
 
 // +controllertools:marker:generateHelp:category="CRD validation"
-// Minimum specifies the minimum numeric value that this field can have.
+// Minimum specifies the minimum numeric value that this field can have. Negative integers are supported.
 type Minimum int
 
 // +controllertools:marker:generateHelp:category="CRD validation"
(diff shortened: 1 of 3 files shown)


DirectXMan12:
Code looks good

Two comments:
1. This need a more detailed explanation of what it's doing in the PR/commit message -- maybe something like

```
✨ Support negative integers in marker syntax

This adds support for negative integers in the marker syntax parser.  This is useful for markers like PUT THE MOTIVATING CASE HERE
```

(please dd the motivating marker to the message)

2. Please update the marker syntax explanation Godoc at the top of that file as well as the book reference page (https://book.kubebuilder.io/reference/markers.html) to mention that negative integers are supported.

knight42:
@DirectXMan12 PTAL

alvaroaleman:
Hmm, I am almost inclined to say doc update isn't needed because I intuitively expect an int to support negative values (otherwise its an uint)?
/lgtm
/hold

To give Solly a but of time to respond re docs

DirectXMan12:
@alvaroaleman that's a fair point.  This does just bring things in line with what you might expect.  Consider that objection withdrawn.

PR #433, opened 2020-04-24
:sparkles: Remove status from generated CRDs

mcristina422:
This sets the CRDs status to empty so it is no longer generated in CRD yamls.

This was originally added in https://github.com/kubernetes-sigs/controller-tools/pull/203 which confuses me because v1beta1 has had the optional marker for awhile https://github.com/kubernetes/kubernetes/pull/66102/files#diff-3029a18192feb70af52de676209da9b6R206

This is currently a problem because continuous delivery tools cant converge when applying them against the updated server set properties

Issue #456, opened 2020-06-30
Generated CRDs have status fields in their output set


maxsmythe:
Generated CRDs have blank values in their status field. This appears to interfere with ArgoCD.

Given that manifests should not define status, it may be worth explicitly removing status from the output.

Here is some sample output from running controller-gen in the Gatekeeper repository:

``` bash
$ controller-gen crd:trivialVersions=true paths="./apis/..." paths="./pkg/..." output:stdout

---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.3.0
  creationTimestamp: null
  name: constraintpodstatuses.status.gatekeeper.sh
spec:
  group: status.gatekeeper.sh
  names:
    kind: ConstraintPodStatus
    listKind: ConstraintPodStatusList
    plural: constraintpodstatuses
    singular: constraintpodstatus
  scope: Namespaced
  validation:
    openAPIV3Schema:
      description: ConstraintPodStatus is the Schema for the constraintpodstatuses
        API
      properties:
        apiVersion:
          description: 'APIVersion defines the versioned schema of this representation
            of an object. Servers should convert recognized schemas to the latest
            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
          type: string
        kind:
          description: 'Kind is a string value representing the REST resource this
            object represents. Servers may infer this from the endpoint the client
            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
          type: string
        metadata:
          type: object
        status:
          description: ConstraintPodStatusStatus defines the observed state of ConstraintPodStatus
          properties:
            constraintUID:
              description: Storing the constraint UID allows us to detect drift, such
                as when a constraint has been recreated after its CRD was deleted
                out from under it, interrupting the watch
              type: string
            enforced:
              type: boolean
            errors:
              items:
                description: Error represents a single error caught while adding a
                  constraint to OPA
                properties:
                  code:
                    type: string
                  location:
                    type: string
                  message:
                    type: string
                required:
                - code
                - message
                type: object
              type: array
            id:
              type: string
            observedGeneration:
              format: int64
              type: integer
            operations:
              items:
                type: string
              type: array
          type: object
      type: object
  version: v1beta1
[...]

PR #472, opened 2020-08-17
:sparkles: Revert ":sparkles: Remove status from generated CRDs"

mcristina422:
Reverts kubernetes-sigs/controller-tools#433

In kubernetes-sigs/controller-tools#433 I attempted to drop the entire status block from generated CRDs for tools like ArgoCD. I wrongly assumed that the `omitempty` marker on the CRD status field made it so the status field would not be generated at all. This is not true due to it not being a pointer, causing it always to be generated. We will need to manually set these still to pass validation

### DIFF
--- a/pkg/crd/spec.go
+++ b/pkg/crd/spec.go
@@ -164,7 +164,11 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind, maxDescLen *int) {
 		packages[0].AddError(fmt.Errorf("CRD for %s with version(s) %v does not serve any version", groupKind, crd.Spec.Versions))
 	}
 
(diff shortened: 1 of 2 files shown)


mcristina422:
@vincepri @alvaroaleman @DirectXMan12 @mengqiy I think this should be merged as it's actually causing incorrect CRDs to be generated on master now https://github.com/kubernetes-sigs/controller-tools/issues/456

I made a mistake in my previous commit when updating the testdata additioanlly. I thought that I had to manually update the generated output to the desired output and that's what I used in my [tests](https://github.com/kubernetes-sigs/controller-tools/pull/433#issuecomment-620727321). I understand now that that's not how that is tested or validated in CI

PR #473, opened 2020-08-20
✨ Support timeoutSeconds marker parser for webhook

tcnksm:
Support `timeoutSeconds` marker parser for webhook.

Since `ValidatingWebhookConfiguration` and `AdmissionConfiguration` support [`timeoutSeconds`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts) which decides how long the API server should wait for a webhook to respond before treating the call as a failure, I would like to configure it via the maker.

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -81,6 +81,11 @@ type Config struct {
 	// AdmissionReview sent in the request, and avoiding side effects if that value is "true."
 	SideEffects string `marker:",optional"`
 
+	// TimeoutSeconds specifies the timeout for this webhook. After the timeout passes,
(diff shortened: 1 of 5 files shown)


tcnksm:
> Please also include the new marker which is introduced in this PR to this line: https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/webhook/parser.go#L22

@jiachengxu Thank you for the comment. I added it https://github.com/kubernetes-sigs/controller-tools/pull/473/commits/015af3598cfcaec9d3889858326f3b247a62f1ba

dashanji:
Hi @jiachengxu @tcnksm, does the pr make sense now, and if so could we make this PR merged?

PR #475, opened 2020-08-21
🏃 add a Nuked CRD schema marker

eddycharly:
This PR adds a `Nuked` schema marker.

A nuked object can be used to represent `Any` object.

This is useful when an object does not conform to the Kubernetes schema constraints.
Used in combination with XPreserveUnknownFields it allows using types that would not validate otherwise.

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -53,6 +53,10 @@ var ValidationMarkers = mustMakeAllWithPrefix("kubebuilder:validation", markers.
 	MinItems(0),
 	UniqueItems(false),
 
+	// Object markers
+
+	Nuked{},
+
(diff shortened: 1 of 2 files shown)


eddycharly:
Thinking about it more, I am not a big fan of this solution.

It should be possible to detect references to objects with the type Any when flattening. When those references are detected, the property could be skipped while adding the preserve unknown field flag.

What are your thoughts on this ?

eddycharly:
Did someone thought about it ? Any feedback welcome. I'd be happy to change the implementation if we can agree on a design.

Maybe we could simply add a `skip` annotation on fields ? Combining skip and preserve unknown fields annotations would do the trick.

/cc @DirectXMan12

Issue #476, opened 2020-08-22
preserveUnknownFields=false doesn't work with crdVersions=v1


jonnylangefeld:
`crdVersions=v1` which is the new default, doesn't work in conjunction with `preserveUnknownFields=false`. The `spec.preserveUnknownFields` simply doesn't appear in the crd. There is an [integration test](https://github.com/kubernetes-sigs/controller-tools/pull/365/files#diff-0ab3d8fb1161ec6dd0c0bfde3d82db95R153-R163) for `v1beta1`, but not for `v1`.

I'm running the following command in a kubebuilder repo:
```
controller-gen "crd:preserveUnknownFields=false,crdVersions=v1" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```

I'm using defaults in my schema and if applying to a `1.16.13-gke.1` cluster, I get:
```
The CustomResourceDefinition "foo.example.com" is invalid: spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema
```
If I add spec.preserveUnknownFields manually to the crd it works

My workaround in the makefile right now is
```
yq w -i config/crd/bases/cluster.paas.getcruise.com_clusters.yaml spec.preserveUnknownFields false
```

pugangxa:
I'm hitting the same issue, any ideas?

npiganeau:
I had the same issue here. It was an issue about upgrading from v1beta1 to v1 CRD.
The `spec.preserveUnknownFields` was true in the cluster (probably because that was the default in v1beta1).
When I upgraded to v1, `spec.preserveUnknownFields` was not set in the manifest, but because it was in the cluster, I had the `spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema` error.

The solution was simply to manually add `spec.preserveUnknownFields: false` in the generated CRD manifest and apply it once. 
From then, it worked again when applying `controller-gen` generated manifests (without  `spec.preserveUnknownFields` set).

bhiravabhatla:
Any updates on this?

elanv:
Any updates?

kevindelgado:
I took a stab at this, but it's a bit hairier than I expected. 

Basically, the root cause is that `CRDSpec.PreserveUnknownFields` went from a type *bool in v1beta1 to a type bool in v1. The default value for a bool is false and because the field has an `omitempty` tag on it, a false value will never be written out to the yaml. 

[...]

PR #440, opened 2020-05-06
🐛 Fix validation errors happening in v1.18

dgrisonnet:
This PR intends to fix validation errors happening when creating generated controllers in Kubernetes v1.18.

```
The CustomResourceDefinition "cronjobs.testdata.kubebuilder.io" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[jobTemplate].properties[spec].properties[template].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[jobTemplate].properties[spec].properties[template].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[embeddedResource].properties: Required value: must not be empty if x-kubernetes-embedded-resource is true without x-kubernetes-preserve-unknown-fields
```

```
The CustomResourceDefinition "cronjobs.testdata.kubebuilder.io" is invalid:
* spec.validation.openAPIV3Schema.properties[spec].properties[defaultedObject].default.nested.foo: Required value
```

In order to fix embeddedResource, and defaultedObject required value errors, this PR adds the missing values inside of `cronjob_types.go`.

For the protocol validation errors, this PR injects the [default Protocol value](https://godoc.org/k8s.io/api/core/v1#ContainerPort) inside of the scheme.

Fixes #438

PR #480, opened 2020-08-26
✨ Set default value for corev1.Protocol type if not already specified

### LINKED ISSUE #444: Default value not allowed for in x-kubernetes-list-map-keys
I filed the issue in k/k: https://github.com/kubernetes/kubernetes/issues/91395

I am cross posting here because I don't know if this is an issue in the upstream CRD validation or the YAML generated by controller-tools.

munnerz:
Mitigation to help resolve #444 whilst waiting for https://github.com/kubernetes/enhancements/pull/1928 to merge.

This patch will cause resources that embed e.g. `ServiceSpec` or `PodSpec` to have the `default` field set for the `protocol` field:

```yaml
                          ports:
                            items:
                              properties:
                                appProtocol:
                                  type: string
                                name:
                                  type: string
                                nodePort:
                                  format: int32
                                  type: integer
                                port:
                                  format: int32
                                  type: integer
                                protocol:
                                  default: TCP
                                  type: string
                                targetPort:
                                  anyOf:
                                  - type: integer
                                  - type: string
                                  x-kubernetes-int-or-string: true
                              required:
                              - port
                              type: object
                            type: array
                            x-kubernetes-list-map-keys:
                            - port
                            - protocol
                            x-kubernetes-list-type: map
```

This will workaround the issue described in #444, and can be easily extended to new fields/types as they are discovered.

We should probably also consider fixing #445 as well, as this may cause some users that are not specifying `// +kubebuilder:default` annotations in their resources to begin specifying a `default` field (if they utilise this `Protocol` type), which will cause `kubectl` client side validation errors (which will require users to add `--validate=false` to their `kubectl create` usages)

/cc @pwittrock @tamalsaha

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -24,6 +24,14 @@ import (
 // KnownPackages overrides types in some comment packages that have custom validation
 // but don't have validation markers on them (since they're from core Kubernetes).
 var KnownPackages = map[string]PackageOverride{
+	"k8s.io/api/core/v1": func(p *Parser, pkg *loader.Package) {
+		// Explicit defaulting for the corev1.Protocol type in lieu of https://github.com/kubernetes/enhancements/pull/1928
+		p.Schemata[TypeIdent{Name: "Protocol", Package: pkg}] = apiext.JSONSchemaProps{
+			Type:    "string",
+			Default: &apiext.JSON{Raw: []byte(`"TCP"`)},
+		}
+		p.AddPackage(pkg)
+	},
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1": func(p *Parser, pkg *loader.Package) {
 		// ObjectMeta is managed by the Kubernetes API server, so no need to
--- a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
+++ b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
@@ -1619,6 +1619,7 @@ spec:
                                               can be referred to by services.
                                             type: string
                                           protocol:
+                                            default: TCP
                                             description: Protocol for port. Must be
                                               UDP, TCP, or SCTP. Defaults to "TCP".
                                             type: string
@@ -2828,6 +2829,7 @@ spec:
                                               can be referred to by services.
                                             type: string
                                           protocol:
+                                            default: TCP
                                             description: Protocol for port. Must be
                                               UDP, TCP, or SCTP. Defaults to "TCP".
                                             type: string


pwittrock:
This is great.  Is this possible to unit test for?

pwittrock:
Any idea why the tests are failing?

munnerz:
I've updated the testdata (the CronJob resource used embeds a PodSpec, so this change causes the `default` field to be correctly generated).

In terms of testing specifically this change, we don't currently have any unit test for `gen.go` or `known_types.go`. All testing functionality is encompassed through the existing `testdata` directory with the single CronJob CRD (and is being exercised due to the embedded PodSpec)

The only behaviour that is _not_ tested is ensuring that if `// +kubebuilder:default` is explicitly specified on this field (or any other marker that gets added once the KEP merges), that it takes precedence over this built-in. The only way we can test _that_ is if the annotation/marker actually does exist on the upstream type, as `k8s.io/api/core/v1` is hardcoded into the `known_types` file.

The only way I can think of to work around this is to have the integration test framework inject a 'copy' of this package/entry into KnownTypes during *tests only*, which could contain a type named `Protocol` and thus make the existing logic in that function usable against our fake test package. If you don't think that sounds too brittle/awful, I can put that together 😄

munnerz:
Okay, I did some more messing around with this and concluded that adding a test to ensure that the default value, if explicit specified with a marker, overrides this built-in default is actually not particularly useful (and not trivial, as during the evaluation of the functions defined in known_types.go, the markers have not yet been processed so we are unable to determine whether an explicit default has been provided here).

Given that this default value is not going to change within `k8s.io/api/core/v1`, I think it is safe for us to always set this default value to TCP. #481 will help users running older Kubernetes versions (pre 1.18) that do not support `default` values either.

@pwittrock this is ready for review now and passing 😄

pwittrock:
@munnerz Can you look into the failed tests?

munnerz:
This PR seems to do the same thing: https://github.com/kubernetes-sigs/controller-tools/pull/440

pwittrock:
> This PR seems to do the same thing: #440

The comments on that PR don't seem like blockers.  Do you have a preference for which PR we take?

pugangxa:
It still does not work for me. if "crdVersions=v1" is specified default value is not generated, see https://github.com/kubernetes-sigs/controller-tools/issues/476

If I use v1beta1, then although default value is generated, but during apply it report error like: 
```
* spec.validation.openAPIV3Schema.properties[spec].properties[predictor].properties[triton].properties[ports].items.properties[protocol].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[predictor].properties[tensorflow].properties[ports].items.properties[protocol].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[predictor].properties[onnx].properties[ports].items.properties[protocol].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[predictor].properties[pytorch].properties[ports].items.properties[protocol].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[predictor].properties[sklearn].properties[ports].items.properties[protocol].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[predictor].properties[xgboost].properties[ports].items.properties[protocol].default: Forbidden: must not be set
```
Any ideas? Thanks~

estroz:
I'm hoping this can be cherry-picked to fix controller-gen in v0.3 for users still wanting v1beta1 CRDs and webhooks by default (ex. in kubebuilder's go/v2 plugin).

/cherry-pick release-0.3

PR #481, opened 2020-08-26
✨ Remove 'default' fields from v1beta1 CRDs

### LINKED ISSUE #445: Strip default values from v1beta1 CRD yaml

munnerz:
It is not valid to specify `default` field values in v1beta1 CRDs.

This pull request strips the generated openapi schema of all instances of `default` being specified.

My only issue with this implementation is that it is non-obvious to users that this has happened, which could cause unexpected behaviour.

Fixes #445 

related #480

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -156,6 +156,11 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		}
 
 		for i, crd := range versionedCRDs {
+			// defaults are not allowed to be specified in v1beta1 CRDs, so strip them
+			// before writing to a file
+			if crdVersions[i] == "v1beta1" {
+				removeDefaultsFromSchemas(crd.(*apiextlegacy.CustomResourceDefinition))
+			}
 			var fileName string
 			if i == 0 {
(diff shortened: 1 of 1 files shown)


munnerz:
I'm looking into ways we can write a unit/integration test for this, similar to `parser_integration_test.go`. The issue is that the parser tests do not call `Generator.Generate` (defined in gen.go), so I'll need to write a new `gen_test.go` file which brings up a Generator and compares the output of the generated content to ensure that the `defaults` field is stripped when writing v1beta1.

DirectXMan12:
In the future, let's not merge things like this w/o tests :-/.
See also: https://github.com/kubernetes-sigs/controller-tools/pull/446#pullrequestreview-485484406

DirectXMan12:
@munnerz can you do a follow up with a) a warning, and b) tests?
They can be separate PR if the warning is quicker.

kevindelgado:
Hey @munnerz I'm gonna try adding the warning/test mentioned by Solly above if you're not already working on it. Just wanted to check and make sure I'm not stepping on your toes.

PR #485, opened 2020-09-01
🐛 Don't allow different generators to interfere with eachother

LukeShu:
I noticed something really weird: I got different `zz_generated.deepcopy.go` files depending on whether I ran `controller-gen crd && controller-gen object` separately, or `controller-gen crd object` together.

The first 3 commits are improving the error checking and reporting so that there's a hint that something's going wrong, instead of just silently doing the wrong thing.

The next 6 commits enhance the deepcopy integration test to catch this.

The penultimate commit fixes the bug.

The final commit fixes a bug waiting to happen that I initially thought was the cause of this bug, before I figured out that the context can't be shared at all.


LukeShu:
/assign @droot

(I think that's what I'm supposed to do now?)

DirectXMan12:
I think we need to prune some maintainers.  A couple of the maintainers switched off the project recently, and still need to be removed from the review rotation.

/unassign @droot 
/assign @DirectXMan12

DirectXMan12:
/hold
Good work on this investigation!

I plan on reading through this tomorrow in more detail.  Sorry for the delay

DirectXMan12:
hmm... as anticipated, this almost doubles the generation time when running two generators, a bit less than triples when running 3, etc, since the vast majority of the time is spent type-checking in controller-tools.

Gonna see if I can take your initial work on making things clearer, etc, and figure out a way to keep the shared context.  Any ideas?

DirectXMan12:
Closing in favor of #493, which swaps out the last two commits but keeps the rest of your work.

PR #488, opened 2020-09-05
✨ add min/max properties validation

### LINKED ISSUE #487: support min/max properties
The apiserver supports validation of `minProperties`/`maxProperties`, having support for that here would be great. `oneOf` is not supported (yet #461), this is a neat workaround to enable mutually exclusive parameters ([see post on SO](https://stackoverflow.com/questions/21134029/how-to-define-mutually-exclusive-query-parameters-in-swagger-openapi/49199240#49199240)).

moolen:
This PR adds `minProperties`/`maxProperties` validation markers.

fixes #487


moolen:
Hey @vincepri sure, that totally makes sense! I missed that on the first glance (i was looking for `*_test.go` files, but found none).

PR #493, opened 2020-09-21
:bug: Don't allow generators to interfere with eachother's type checking

DirectXMan12:
Continuation of #485 with some changes to the actual bugfix.

From the original author, LukeShu:

> I noticed something really weird: I got different zz_generated.deepcopy.go files depending on whether I ran controller-gen crd && controller-gen object separately, or controller-gen crd object together.

> The first 3 commits are improving the error checking and reporting so that there's a hint that something's going wrong, instead of just silently doing the wrong thing.

> The next 6 commits enhance the deepcopy integration test to catch this.

The final commit is the fix: we gather all filters ahead of time and consider their union: if any filter returns true for a given node, that node is considered interesting.


DirectXMan12:
cc @LukeShu

Issue #448, opened 2020-05-29
Objectmeta of embedded object goes missing during CRD conversion


champak:
Hi folks,
I ran into an odd issue with CRD conversion using controller-tools v0.2.4/kubebuilder 2.2.0/apiserver 17.1 This CRD has embedded native specs like the PodTemplateSpec in its spec. I found that in the conversion functions the ObjectMeta of the embedded PodTemplateSpec is nulled out. Other attributes look OK. There is nothing in the validation callback. I had needed to add preserveUnknownFields: false to the CRD to be able to deploy the yaml; not sure if that has some sideffect

The ObjectMeta for the CRD itself shows up fine. It is the embedded spec's ObjectMeta that seemed to be affected. Does this sound like any known issue ? One that looked somewhat related from a search was kubernetes-sigs/controller-tools#279 with comments from @DirectXMan12 but was not quite the same problem.
Thanks for any insight !

DirectXMan12:
/kind bug
/priority critical-urgent
/needs-more-info

any idea if this is happening during the conversion webhook?  Can you post a reproducer?

champak:
I had added logs to the ConvertTo(dstRaw conversion.Hub) and ConvertFrom(srcRaw conversion.Hub) functions in the file api/v<>/<>_conversion.go to display the contents of the ObjectMeta from the CRD and the embedded PodTemplateSpec. The ObjectMeta.Name etc from the CRD level had the expected contents but the contents from the embedded podtemplatespec were nulled.   Are there other places that may be useful to log ? I was wondering if the api-server itself could have chosen to not send the metadata of the embedded spec to the webhook. Have not looked yet at what all layers are in play between the api-server and the ConvertTo/From functions. I am wondering if I am missing some knob. I was following the kubebuilder book on the conversion.

The CRD structure is on the lines of ;
CRDfooSpec {
<name> barspec
}
barspec {
Pod corev1.PodTemplateSpec  ...
...
}


 The embedded podtemplate spec contents are visible in the last-applied-configuration annotation of the CRD so the info did make it to the api-server. I will try to post a repro. Thanks !

champak:
I created a simple repro for this issue.  @DirectXMan12 please see https://github.com/champak/repro-kbuilder-version-issue . 

If the sample yaml at https://github.com/champak/repro-kbuilder-version-issue/blob/master/config/samples/left_v1alpha1_repro.yaml

is applied, the logs at the webhook converter look like

[...]

PR #494, opened 2020-09-26
(:warning:, major), Adding an option flag for fixing up object meta-data

champak:
Added an option for preserving embedded object meta-data ( discussed at https://github.com/kubernetes-sigs/controller-tools/issues/448).

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -131,7 +131,7 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		crdRaw := parser.CustomResourceDefinitions[groupKind]
 		addAttribution(&crdRaw)
 
-		if *g.PreserveObjectMeta {
+		if g.PreserveObjectMeta != nil && *g.PreserveObjectMeta {
 			FixupMetadata(crdRaw)
 		}
 
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -83,6 +83,9 @@ type Generator struct {
 	// You'll need to use "v1" to get support for features like defaulting,
 	// along with an API server that supports it (Kubernetes 1.16+).
 	CRDVersions []string `marker:"crdVersions,optional"`
+
+	// preserve object meta data
+	PreserveObjectMeta *bool `marker:",optional"`
 }
 
 func (Generator) CheckFilter() loader.NodeFilter {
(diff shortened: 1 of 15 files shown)


champak:
@DirectXMan12  just doing a poll; is this what you were describing as needed on #395 ? Thanks

pkg/crd/gen.go
```diff
@@ -83,6 +83,9 @@ type Generator struct {
 	// You'll need to use "v1" to get support for features like defaulting,
 	// along with an API server that supports it (Kubernetes 1.16+).
 	CRDVersions []string `marker:"crdVersions,optional"`
+
+	// preserve object meta data
+	PreserveObjectMeta *bool `marker:",optional"`
```
DirectXMan12, line 88: this flag seems to do the inverse of what i'd expect, if I'm reading this right

Here's what I'd expect:

if you pass PreserveObjectMeta, you get an object with no top-level metadata (it's enforced by the apiserver), but all the embedded object metadata fields are filled out.

If you don't pass PreserveObjectMeta, you get the behavior today, which is that *all* objectmeta fields are wiped.

DirectXMan12 (changes requested):
close to it -- comments inline

pkg/crd/gen.go
champak: @DirectXMan12 thanks for taking a look ! I probably misunderstood but it had seemed when PreserveObjectMeta is either unspecified or set to false, with this update we would not invoke the new fnc FixupMetaData() and hence the behavior would be what we have today. I shall re-read and flip the check if that is what you meant.

dvaldivia:
I'm testing this PR, and I see `metadata` being generated for the top CRD schema, which causes an error 
```
* spec.versions[1].schema.openAPIV3Schema.properties[metadata]: Forbidden: must not specify anything other than name and generateName, but metadata is implicitly specified
```

I think this one @DirectXMan12 explicitly mentions not to generate

Also, it seems like `metadata. managedFields.fieldsV1` causes some problem, as the kubernetes API is rejecting the CRD with the following problem
```
* spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[pools].items.properties[volumeClaimTemplate].properties[metadata].properties[managedFields].items.properties[fieldsV1].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
```

this error seems valid, I cannot find any reference in the openAPI documentation for this `Any` type of value

pkg/crd/gen.go
dvaldivia: @DirectXMan12 I think the top leve metadata will be rejected by the API server regardless, I thought this PR was about keeping the metadata for embedded `ObjectMeta` in the CRDs

dvaldivia: cc @champak

PR #496, opened 2020-10-01
✨  Install golangci-lint if not present in test.sh

ChunyiLyu:
Currently, If you do not have `golangci-lint`, test.sh just outputs a line "using tools" and nothing else, which is a bit confusing.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -99,7 +99,11 @@ function setup_envs {
 
 header_text "using tools"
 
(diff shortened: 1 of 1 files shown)


ChunyiLyu:
I signed it

ChunyiLyu:
@droot @pwittrock Bitesize change and could be reviewed quickly :)

PR #501, opened 2020-10-07
:seedling: Update dependencies to v1.19.2

vincepri:


.run-controller-gen.sh
```diff
@@ -4,11 +4,25 @@
...
+            fi
+        fi
+        echo "you're probably on OSX.  Please install gnu readlink -- otherwise you're missing the most useful readlink flag."
+        exit 1
+    fi
+}
+current_dir=$(pwd)
+check_readlink
+cd $(dirname $(${readlink} -f ${BASH_SOURCE[0]}))
 go run -v -exec "./.run-in.sh ${current_dir} " ./cmd/controller-gen $@
```
vincepri, line 28: @DirectXMan12 Made some changes to support gnu readlink on OSX when it's not linked (usually installed with brew install coreutils)

go.mod
```diff
@@ -1,20 +1,20 @@
...
-	golang.org/x/tools v0.0.0-20200616195046-dc31b401abb5
-	gopkg.in/yaml.v3 v3.0.0-20190905181640-827449938966
-	k8s.io/api v0.18.2
-	k8s.io/apiextensions-apiserver v0.18.2
-	k8s.io/apimachinery v0.18.2
+	golang.org/x/tools v0.0.0-20201007032633-0806396f153e
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
+	k8s.io/api v0.19.2
+	k8s.io/apiextensions-apiserver v0.19.2
+	k8s.io/apimachinery v0.19.2
```
vincepri, line 18: @DirectXMan12 It seems that a test is now failing after updating to v0.19.2 

```
  Expected success, but got an error:
      <*errors.errorString | 0xc0000b8ed0>: {
          s: "converting (v1.MutatingWebhookConfiguration) to (v1beta1.MutatingWebhookConfiguration): unknown conversion",
      }
      converting (v1.MutatingWebhookConfiguration) to (v1beta1.MutatingWebhookConfiguration): unknown conversion
```

From 1.19 changelog:
```
K8s.io/apimachinery - scheme.Convert() now uses only explicitly registered conversions - default reflection based conversion is no longer available. +k8s:conversion-gen tags can be used with the k8s.io/code-generator component to generate conversions.
```
Change came in: https://github.com/kubernetes/kubernetes/pull/90018


It seems we're relying on scheme.Convert to convert webhooks, and I can't seem to find where/how to register the conversion functions in the admission APIs 🤔, I might be missing something here, although just registering the types with `AddToScheme` in https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/webhook/conv.go#L27-L33 doesn't seem enough for the conversions to work.

cc @wojtek-t @liggitt

liggitt: Conversions from v1 directly to v1beta1 AdmissionReview objects are not defined, even internally in kube-apiserver.

The API server would convert from one to the other by way of the internal (unversioned) AdmissionReview type:

* [v1beta1 -> internal](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/admission/v1beta1/zz_generated.conversion.go#L185-L188)
* [internal -> v1](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/admission/v1/zz_generated.conversion.go#L204-L207)

This conversions and the internal AdmissionReview type are registered in https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/admission/install/install.go but are not exported for use outside the kube-apiserver

The reflect-based conversion controller-tools was depending on previously was closer to decoding v1beta1 JSON into the v1 struct and hoping all the field names stayed the same.

liggitt: > The reflect-based conversion controller-tools was depending on previously was closer to decoding v1beta1 JSON into the v1 struct and hoping all the field names stayed the same.

It looks like this happens to be the case:

```sh
diff staging/src/k8s.io/api/admission/{v1,v1beta1}/types.go
```

```diff
2c2
< Copyright 2019 The Kubernetes Authors.
---
> Copyright 2017 The Kubernetes Authors.
17c17
< package v1
---
> package v1beta1
26a27,31
> // +k8s:prerelease-lifecycle-gen:introduced=1.9
> // +k8s:prerelease-lifecycle-gen:deprecated=1.19
> // This API is never server served.  It is used for outbound requests from apiservers.  This will ensure it never gets served accidentally
> // and having the generator against this group will protect future APIs which may be served.
> // +k8s:prerelease-lifecycle-gen:replacement=admission.k8s.io,v1,AdmissionReview
118c123
< 	// This must be copied over from the corresponding AdmissionRequest.
---
> 	// This should be copied over from the corresponding AdmissionRequest.
```

but the fact that the structs happened to be identical between versions was luck, and doesn't mean this is a correct way to convert between them

vincepri: Thanks Jordan, it seems the removal of the default reflection-based conversion was definitely a good step, things might have failed silently. For this codebase, I think the best way is to add support for both types, or deprecate one, or write our own conversion functions. @DirectXMan12 @alvaroaleman what do you all think?

alvaroaleman: Why do we have a need to convert? APIServers should still serve both or not?

liggitt: AdmissionReview is not a served API, it is the API sent by the API server to admission webhooks and expected in response.

I'm not sure why this component is converting... is it trying to unify v1 and v1beta1 AdmissionReview handling to a single type?

DirectXMan12: IIRC, we have a single interface that represents "webhook" that can interact with whatever the hook sends, and we autoconvert from whatever the server sends to the version the interface uses.

Part of it was simplifying the writing of the webhook while still giving folks access to the underlying information.  We *should* be able to paper over it, I think, but I haven't looked at this bit in a while.

estroz:
@vincepri this can be updated now.

vincepri:
@estroz Rebased on top of the latest changes, although we still need to apply the same patch we did in Controller Runtime here

```
Webhook Generation From Parsing to CustomResourceDefinition
/Users/vince/go/src/sigs.k8s.io/controller-tools/pkg/webhook/parser_integration_test.go:38
  should properly generate the webhook definition [It]
  /Users/vince/go/src/sigs.k8s.io/controller-tools/pkg/webhook/parser_integration_test.go:43

  Expected success, but got an error:
      <*errors.errorString | 0xc00031adb0>: {
          s: "converting (v1.MutatingWebhookConfiguration) to (v1beta1.MutatingWebhookConfiguration): unknown conversion",
      }
      converting (v1.MutatingWebhookConfiguration) to (v1beta1.MutatingWebhookConfiguration): unknown conversion

  /Users/vince/go/src/sigs.k8s.io/controller-tools/pkg/webhook/parser_integration_test.go:70
```

.run-controller-gen.sh
```diff
@@ -4,11 +4,25 @@
...
-fi
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readlink=$(command -v readlink)
 
-cd $(dirname $(readlink -f ${BASH_SOURCE[0]}))
+check_readlink() {
+    if ! ${readlink} -f &>/dev/null; then
```
estroz, line 14: This will always fail with no file argument.
```console
$ uname -o
GNU/Linux
$ readlink -f
readlink: missing operand
Try 'readlink --help' for more information.
```

tamalsaha:
Could this be upgraded to 1.120 instead since controller-runtime is already using 1.20 https://github.com/kubernetes-sigs/controller-runtime/pull/1268

vincepri:
We should probably release two versions, one for 1.19 and one for 1.20

vincepri:
@estroz Do you have some time to finish up these changes?

Issue #502, opened 2020-10-08
Issue generating CRDs with controller-gen 0.4.0


chrispatmore:
Hey, I have been using controller-gen 0.3.0 to generate my CRD. And with no changes the CRD fails to generate when using 0.4.0

The problem is that the validation for stats.conditions in my CRD is being completely removed and replaced with type: Any

I have been using this package https://github.com/operator-framework/operator-lib/blob/v0.1.0/status/conditions.go
for the status conditions field in my api 

```
import (
	"github.com/operator-framework/operator-lib/status"
)
Conditions status.Conditions `json:"conditions"`
```

error
`CustomResourceDefinition.apiextensions.k8s.io "<group>" is invalid: spec.validation.openAPIV3Schema.properties[status].properties[conditions].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"`

Is there a way to fix this error and get things working again?

alvaroaleman:
Same problem, works fine with 0.3.0: https://github.com/kubernetes/test-infra/pull/19522#issuecomment-707240265

PR #505, opened 2020-10-12
:bug: Don't generate schema with `Any` in it, its not supported

alvaroaleman:
This reverts commit e0d7c9d9723baea95e57a12751864e2e8f7d7c0f.

`Any` is not a valid (tested with both v1.18 and v1.19):
```
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[params].items.properties[default].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
```
See also https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#type although that list doesn't match whats in the error

cc @skaslev 

Fixes https://github.com/kubernetes-sigs/controller-tools/issues/502

/assign @vincepri

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -19,7 +19,6 @@ package crd
 import (
 	"fmt"
 	"go/ast"
-	"go/token"
 	"go/types"
 	"strings"
 
@@ -109,11 +108,6 @@ func (c *schemaContext) requestSchema(pkgPath, typeName string) {
 
 // infoToSchema creates a schema for the type in the given set of type information.
 func infoToSchema(ctx *schemaContext) *apiext.JSONSchemaProps {
-	if obj := ctx.pkg.Types.Scope().Lookup(ctx.info.Name); obj != nil && implementsJSONMarshaler(obj.Type()) {
-		schema := &apiext.JSONSchemaProps{Type: "Any"}
-		applyMarkers(ctx, ctx.info.Markers, schema, ctx.info.RawSpec.Type)
-		return schema
-	}
 	return typeToSchema(ctx, ctx.info.RawSpec.Type)
 }
 
@@ -430,17 +424,4 @@ func builtinToType(basic *types.Basic, allowDangerousTypes bool) (typ string, fo
 	}
 
 	return typ, format, nil
-}
-
-// Open coded go/types representation of encoding/json.Marshaller
-var jsonMarshaler = types.NewInterfaceType([]*types.Func{
-	types.NewFunc(token.NoPos, nil, "MarshalJSON",
-		types.NewSignature(nil, nil,
-			types.NewTuple(
-				types.NewVar(token.NoPos, nil, "", types.NewSlice(types.Universe.Lookup("byte").Type())),
-				types.NewVar(token.NoPos, nil, "", types.Universe.Lookup("error").Type())), false)),
-}, nil).Complete()
-
-func implementsJSONMarshaler(typ types.Type) bool {
-	return types.Implements(typ, jsonMarshaler) || types.Implements(types.NewPointer(typ), jsonMarshaler)
 }
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -23,9 +23,7 @@ limitations under the License.
 package cronjob
 
 import (
-	"encoding/json"
 	"fmt"
-	"net/url"
 
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


vincepri (approved):
/approve
/lgtm

maxsmythe:
This PR breaks the ability to generate CRDs that embed JSONSchemas. Is there some way to retain that support?

At a minimum, maybe we can use x-kubernetes-preserve-unknown-fields?

alvaroaleman:
> This PR breaks the ability to generate CRDs that embed JSONSchemas. Is there some way to retain that support?

No it does not. `Any` is not allowed as vs value, see the PRs title and description.

maxsmythe:
The fact remains that previously it was possible to generate objects that had JSONSchema embedded in them via controller-gen  and now it is not.

From my eyes as a user this is a regression. Before it would be possible for me to generate the CRD and clean it up later with kustomize or similar. Now that option is no longer available.

Per my suggestion, finding some way to use `x-kubernetes-preserve-unknown-fields` would be a way to generate actual valid CRD output.

alvaroaleman:
> From my eyes as a user this is a regression. Before it would be possible for me to generate the CRD and clean it up later with kustomize or similar. Now that option is no longer available.

Generating an invalid object that you can then manually fix is not a feature, its a bug. Please describe the problem you are having in a new issue rather than resurrecting old issues and PRs.

maxsmythe:
From a pragmatic standpoint, I disagree. This eliminates the possibility of users who are aware that this is an edge case of including controller-gen in their generation pipeline, the purpose for which controller-gen was intended.

If we want to assume that we only generate valid output (not necessarily wise, as it couples controller-gen to specific K8s versions), we could at least flag-gate the feature as in `crd:allowDangerousTypes=true`

Further, there is a pre-existing issue that this PR re-breaks:

#291

I am happy to re-open

alvaroaleman:
> From a pragmatic standpoint, I disagree

This change broke anyones crds generation who had a type in them that had a JSONUnmarshaler implementation and no kubebuilder tags.

> If we want to assume that we only generate output that passes K8s validation (not necessarily wise, as it couples controller-gen to specific K8s versions)

`Any` is invalid on all kube versions.

> I am happy to re-open

Please do that or create a new issue and please also feel free to suggest a fix that doesn't break ppl the way this one did. Conversations on old issues and PRs will go unnoticed for many.

maxsmythe:
> > From a pragmatic standpoint, I disagree
> 
> This change broke anyones crds generation who had a type in them that had a JSONUnmarshaler implementation and no kubebuilder tags.

Interesting. Can you be more specific? This would happen even if the struct was not tagged with kubebuilder tags that happened to be a package touched by controller-gen?

> 
> > If we want to assume that we only generate output that passes K8s validation (not necessarily wise, as it couples controller-gen to specific K8s versions)
> 
> `Any` is invalid on all kube versions.

That is true for this specific instance. As a general design principal, I'm not sure it passes muster. Imagine `Any` becomes valid in some future kube version? What about other validation mechanisms that change over time, such as the introduction of structural schemas?

> 
> > I am happy to re-open
> 
> Please do that or create a new issue and please also feel free to suggest a fix that doesn't break ppl the way this one did. Conversations on old issues and PRs will go unnoticed for many.

For sure, happy to continue the discussion. Note that the reason this thread was started is because this PR came up on a closed bug for which people were still experiencing issues.

So far I have two (hand-wavy) proposals I've mentioned that satisfy your criteria:

   1. If JSONSchema is embedded, just include it and set `x-kubernetes-preserve-unknown-fields` instead of using the "Any" type.
   2. Un-revert the PR but add a logic gate similar to `crd:allowDangerousTypes=true` that lets people opt into generating more edge-case output that may require further cleanup

Depending on if the bug gets re-opened, I'll add these suggestions there, though this PR is already linked to the old bug and will be linked to any new bugs.

alvaroaleman:
> Interesting. Can you be more specific? This would happen even if the struct was not tagged with kubebuilder tags and happened to be a package touched by controller-gen?

https://github.com/kubernetes-sigs/controller-tools/issues/502, as mentioned in the PR body. This would only happen if the struct was not tagged with kubebuilder tags.

> That is true for this specific instance. As a general design principal, I'm not sure it passes muster. Imagine Any becomes valid in some future kube version? What about other validation mechanisms that change over time, such as the introduction of structural schemas?

Optimizing for something that one thinks could happen in the future without any evidence for that to actually happening is premature optimization. Once we would have this case, we would most likely allow it and tell ppl that it might fail on kube versions older min supported.

> For sure, happy to continue the discussion. Note that the reason this thread was started is because this PR came up on a closed bug for which people were still experiencing issues.


https://github.com/kubernetes-sigs/controller-tools/issues/291 is a kitchen sink bug for all kind of issues ppl had over time, mostly with types that are part of the k8s.io api. Please open a new bug, paste the relevant part of your type and mention what you want to happen. Right now this is a thin air discussion. The revert happened because
So re-open that bug. As you can see in the PRs description, the change as is breaks ppl. Losing the ability to create an invalid schema that can then manually be fixed up in order to unbreak ppl is definitely a good tradeoff.

> If JSONSchema is embedded, just include it and set x-kubernetes-preserve-unknown-fields instead of using the "Any" type.

Its unclear to me what you mean by "if jsonschema is embedded'. The reverted change made the assumption that ppl want an invalid CRD if they have types that implement a JSONUnmarshaler and do not have kubebuilder tags. That assumption does definitely not make sense. The same would go for adding `x-kubernetes-preserve-unknown-fields`, such behavior must only be triggered explicitly and never implicitly. Ppl might simply have types with a JSONUnmarshaler and no kubebuilder tags because they import the type from a different project.

maxsmythe:
> #502, as mentioned in the PR body

Ack, the linked bug clarifies what you were fixing.

> Optimizing for something that one thinks could happen in the future

In this case I am not suggesting optimizing for a case that we think could happen in the future.

1) I am suggesting __not__ doing something
2) There is clearly some unmet need here as you and I are having this conversation

Unless one of us is steering the actual direction behind this project, it's probably not worth continuing this discussion here, as it would be a lot of debate with zero impact.

> #291 is a kitchen sink bug for all kind of issues

#291 is most definitely not a kitchen sink bug. From its title, it is for users trying to embed [JSONSchemaProps](https://github.com/kubernetes/apiextensions-apiserver/blob/a7ee1efe41fc19623ce57f047c7595d624df4c17/pkg/apis/apiextensions/v1beta1/types_jsonschema.go#L20) in their CRDs.

I agree that this reverted fix cast too wide a net. I'm looking at what it would take to add a new tag so people can opt-in to the schema-less behavior wrapping CRDs inside of CRDs appears to require.

PR #508, opened 2020-10-13
:bug: Set string as type if field of struct has a string json tag

### LINKED ISSUE #507: Controller-gen doesn't intepreate "string" json tag
Given this struct
```
type Something struct {
	SomeAtrribute int               `json:"someAtrribute,string,omitempty"`
}
```
the validation schema will be as 
```
someAtrribute:
   type: integer
```
and all requests with this field set will fail.

xpivarc:
Fixes #507

Signed-off-by: L. Pivarc <lpivarc@redhat.com>


xpivarc:
I signed it

PR #510, opened 2020-10-19
:bug: keep name in ObjectMeta for volumeClaimTemplates

bonifaido:
I have a CRD which will become a StatefuleSet (amongst many other things) at the end with the help of an operator, and it is possible to define `volumeClaimTemplates` as well for that StatefuleSet. The issue is when I'm using the CRD generator for this type, the `PersistentVolumeClaim.metadata.name` field gets removed from the generated CRD YAML, which makes the parsed type miss this name values as well, which has a defined meaning in this case and should be present, otherwise, the statefulset won't produce any pods. With this PR I'm trying to heal this situation (and hopefully won't break other things), I know it adds the metadata.name for other types as well, but that seems to do no harm.

Tested locally and the vault-operator seems to be working this way.

https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/apis/vault/v1alpha1/vault_types.go#L314

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -38,6 +38,13 @@ var KnownPackages = map[string]PackageOverride{
 		// generate validation for it.
 		p.Schemata[TypeIdent{Name: "ObjectMeta", Package: pkg}] = apiext.JSONSchemaProps{
 			Type: "object",
+			// The name property is an exception, it has a special role volumeClaimTemplates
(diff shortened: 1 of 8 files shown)


bonifaido:
I just checked and there is https://github.com/kubernetes-sigs/controller-tools/pull/254 which tried to solve this and got reverted later on. What would be the proper solution here?

PR #499, opened 2020-10-02
✨ Add test and warning for v1beta1 default removal

kevindelgado:
This is a small follow up to #481 to include test cases and a warning for when v1beta1 CRDs are stripped of defaults.

Also fixes a gosec lint issue caused by "Implicit memory aliasing in for loop"

PR #511, opened 2020-10-23
🐛 Don't generate description on metadata of CRD structural schema

### LINKED ISSUE #386: Avoid generating description in metadata of CRD structural schema
CRD structural schema requires that `if metadata is specified, then only restrictions on metadata.name and metadata.generateName are allowed.`

If I add comments to `metadata` field in Go struct, kubebuilder sets that as `metadata.description` in the generated CRd yaml. This generated yaml is considered invalid as structural schema.

Example:
- https://github.com/appscode/voyager/commit/5f8a169dae4eb73738ec965575328177026bf4b9#diff-621dd318cc94d163f872e0712dc053a8L47-L48
- https://github.com/appscode/voyager/commit/5f8a169dae4eb73738ec965575328177026bf4b9#diff-c793f5d116450276c7d0a4cb7e90b538L49

I wonder if controller-tools can avoid generating metadata.description at all times?

kevindelgado:
### What

Ensures that generated CRDs do not have a description generated from godoc comments on the ObjectMeta of the corresponding go type.

### Why

Fixes #386 

CRD structural schema requires that if metadata is specified, then only restrictions on metadata.name and metadata.generateName are allowed.

metadata.description is not allowed

xref: kubernetes/kubernetes#86800

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -161,10 +161,14 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		}
 
 		for i, crd := range versionedCRDs {
-			// defaults are not allowed to be specified in v1beta1 CRDs, so strip them
-			// before writing to a file
+			// defaults are not allowed to be specified in v1beta1 CRDs and
+			// decriptions are not allowed on the metadata regardless of version
+			// strip them before writing to a file
 			if crdVersions[i] == "v1beta1" {
 				removeDefaultsFromSchemas(crd.(*apiextlegacy.CustomResourceDefinition))
+				removeDescriptionFromMetadataLegacy(crd.(*apiextlegacy.CustomResourceDefinition))
+			} else {
+				removeDescriptionFromMetadata(crd.(*apiext.CustomResourceDefinition))
 			}
 			var fileName string
 			if i == 0 {
@@ -179,6 +183,49 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	}
 
 	return nil
+}
+
+func removeDescriptionFromMetadata(crd *apiext.CustomResourceDefinition) {
+	for _, versionSpec := range crd.Spec.Versions {
+		if versionSpec.Schema != nil {
+			removeDescriptionFromMetadataProps(versionSpec.Schema.OpenAPIV3Schema)
+		}
+	}
+}
+
+func removeDescriptionFromMetadataProps(v *apiext.JSONSchemaProps) {
+	if m, ok := v.Properties["metadata"]; ok {
+		meta := &m
+		if meta.Description != "" {
+			fmt.Fprintf(os.Stderr, "Warning: metadata description unsupported. Removing description: %s\n", meta.Description)
+			meta.Description = ""
+			v.Properties["metadata"] = m
+
+		}
+	}
+}
+
+func removeDescriptionFromMetadataLegacy(crd *apiextlegacy.CustomResourceDefinition) {
+	if crd.Spec.Validation != nil {
+		removeDescriptionFromMetadataPropsLegacy(crd.Spec.Validation.OpenAPIV3Schema)
+	}
+	for _, versionSpec := range crd.Spec.Versions {
+		if versionSpec.Schema != nil {
+			removeDescriptionFromMetadataPropsLegacy(versionSpec.Schema.OpenAPIV3Schema)
+		}
+	}
+}
+
+func removeDescriptionFromMetadataPropsLegacy(v *apiextlegacy.JSONSchemaProps) {
+	if m, ok := v.Properties["metadata"]; ok {
+		meta := &m
+		if meta.Description != "" {
+			fmt.Fprintf(os.Stderr, "Warning: metadata description unsupported. Removing description: %s\n", meta.Description)
+			meta.Description = ""
+			v.Properties["metadata"] = m
+
+		}
+	}
 }
 
 // removeDefaultsFromSchemas will remove all instances of default values being
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -197,7 +197,6 @@ func removeDescriptionFromMetadataProps(v *apiext.JSONSchemaProps) {
 	if m, ok := v.Properties["metadata"]; ok {
 		meta := &m
 		if meta.Description != "" {
-			fmt.Fprintf(os.Stderr, "Warning: metadata description unsupported. Removing description: %s\n", meta.Description)
 			meta.Description = ""
 			v.Properties["metadata"] = m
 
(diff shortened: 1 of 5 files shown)


kevindelgado:
Was waiting for #499 to land with the basic gen_integration_test before nudging this, but that's merged now, can you take a look @DirectXMan12

pkg/crd/gen.go
```diff
@@ -312,6 +312,11 @@ func FindKubeKinds(parser *Parser, metav1Pkg *loader.Package) map[schema.GroupKi
 			switch namedField.Obj().Name() {
 			case "ObjectMeta":
 				hasObjectMeta = true
+				// Clear the godoc comment from the ObjectMeta field
+				// so it does not show up as the description, because
+				// description is not allowed in metadata of CRD structural schema.
+				field.Doc = ""
+				info.Fields[i] = field
```
whalecold, line 319: Hi, `info.Fields[i].Doc = ""` may be better.

kevindelgado: Thanks!

kevindelgado:
friendly ping @DirectXMan12

DirectXMan12:
/lgtm cancel
/approve cancel

we should be doing this in the output openapi, not in the input godoc

pkg/crd/gen.go
```diff
@@ -161,10 +161,14 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
 
 		for i, crd := range versionedCRDs {
-			// defaults are not allowed to be specified in v1beta1 CRDs, so strip them
-			// before writing to a file
+			// defaults are not allowed to be specified in v1beta1 CRDs and
+			// decriptions are not allowed on the metadata regardless of version
+			// strip them before writing to a file
 			if crdVersions[i] == "v1beta1" {
 				removeDefaultsFromSchemas(crd.(*apiextlegacy.CustomResourceDefinition))
+				removeDescriptionFromMetadataLegacy(crd.(*apiextlegacy.CustomResourceDefinition))
```
kevindelgado, line 169: Not loving the duplicated functions for v1 and v1beta1, but I couldn't find a cleaner way to make it work. I thought of passing the generic `runtime.Object` into a single `removeDescriptionFromMetadata()` function, but it didn't seem any better since we already have to do the crdVersions check for removing defaults. Let me know if there's a better way.

kevindelgado (commented):
metadata description is now stripped on the output open API schema instead of the input godoc.

PTAL @DirectXMan12

pkg/crd/gen.go
```diff
@@ -181,6 +185,49 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
+			removeDescriptionFromMetadataProps(versionSpec.Schema.OpenAPIV3Schema)
+		}
+	}
+}
+
+func removeDescriptionFromMetadataProps(v *apiext.JSONSchemaProps) {
+	if m, ok := v.Properties["metadata"]; ok {
+		meta := &m
+		if meta.Description != "" {
+			fmt.Fprintf(os.Stderr, "Warning: metadata description unsupported. Removing description: %s\n", meta.Description)
```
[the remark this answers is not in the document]
whalecold: Hi, I test your code in myself environment but it don't work well as lack of consider the other fields which may container the struct `JSONSchemaProps`also. I try to changed the code as follow and it works. That's may be a suggestion.
```
	if v == nil {
		return
	}
	for str, property := range v.Properties {
		if str == "metadata" {
			if property.Description != "" {
				fmt.Fprintf(os.Stderr, "Warning: version metadata description unsupported. Removing description: %s\n", property.Description)
				property.Description = ""
			}
		}
		if property.Items != nil {
			removeDescriptionFromMetadataProps(property.Items.Schema)
			for i := range property.Items.JSONSchemas {
				removeDescriptionFromMetadataProps(&property.Items.JSONSchemas[i])
			}
		}
		if property.AdditionalProperties != nil {
			removeDescriptionFromMetadataProps(property.AdditionalProperties.Schema)
		}
		if property.AdditionalItems != nil {
			removeDescriptionFromMetadataProps(property.AdditionalItems.Schema)
		}
		removeDescriptionFromMetadataProps(&property)
		v.Properties[str] = property
	}
```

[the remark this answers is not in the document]
whalecold: friendly ping @kevindelgado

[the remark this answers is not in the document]
kevindelgado: Thanks!

kevindelgado:
PTAL @DirectXMan12

pkg/crd/gen.go
```diff
@@ -181,6 +185,87 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
+			}
+			if property.AdditionalProperties != nil {
+				removeDescriptionFromMetadataProps(property.AdditionalProperties.Schema)
+			}
+			if property.AdditionalItems != nil {
+				removeDescriptionFromMetadataProps(property.AdditionalItems.Schema)
+			}
+			// property var reference is fine -- we handle the persistence of the modfications on the line below
+			//nolint:gosec
+			removeDescriptionFromMetadataProps(&property)
```
DirectXMan12, line 220: why are we recursing here?

kevindelgado: See feedback from @whalecold [here](https://github.com/kubernetes-sigs/controller-tools/pull/511/commits/cea95b3c509d4c30168e91d59f0a1e4eada49390#r530237711)

I took it to mean that there could be objects embedded within the object (e.g. within the spec field), is that wrong? See the first commit for the non-recursive approach.

@whalecold can you provide an example of where this breaks if you don't recurse the fields that contain addition `JSONSchemaProps`?

pkg/crd/gen.go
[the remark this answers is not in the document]
DirectXMan12: top-level metadata shouldn't have any fields -- it's explicitly stripped because the kubernetes API server replaces/ignores most of it.  

AFAIK, the API server only enforces this constraint on the top-level metadata field, not any embedded metadata, so it should be sufficient to simply clear description from the the root metadata field, and leave it at that.

@whalecold if this doesn't work, can you please provide an example that's giving you issues

pkg/crd/gen.go
kevindelgado: Reverted back to modifying only the top level metadata description, as @DirectXMan12 and I tested that kube apiserver accepts embedded objects with metadata descriptions.

pkg/crd/gen.go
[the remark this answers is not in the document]
DirectXMan12: I think this warning might actually just confuse people too, so prob best to remove it -- it's gonna hit a lot of codebases and doesn't really have a material impact on them.

Issue #287, opened 2019-07-26
map values must be a named type


tamalsaha:
My apis package has a Go struct that is not a CRD. But controller-gen keeps throwing the following error:

```console
Generating CRD manifests
/go/src/kubedb.dev/apimachinery/apis/kubedb/v1alpha1/report.go:21:27: map values must be a named type, not *ast.StarExpr
/go/src/kubedb.dev/apimachinery/apis/kubedb/v1alpha1/report.go:22:27: map values must be a named type, not *ast.StarExpr
Error: not all generators ran successfully
```

```go
// +kubebuilder:object:root=false
// +k8s:deepcopy-gen=false
// +k8s:gen-deepcopy=false
type ReportSummary struct {
	Postgres      map[string]*PostgresSummary      `json:"postgres,omitempty"`
	Elasticsearch map[string]*ElasticsearchSummary `json:"elastic,omitempty"`
}
```

Full Example: 
https://github.com/kubedb/apimachinery/blob/report-kubebuilder-bug/apis/kubedb/v1alpha1/report.go

Clone and checkout this branch and then run `make gen` to reproduce.

How do I tell controller-gen to ignore this struct?

ifilonenko:
any progress on this?

max8899:
@tamalsaha try to use the latest controlller-gen in master branch

```
GO111MODULE=on go get -v sigs.k8s.io/controller-tools/cmd/controller-gen@master
```

and re-generate will be work

mishindm:
> @tamalsaha try to use the latest controlller-gen in master branch
> 
> ```
> GO111MODULE=on go get -v sigs.k8s.io/controller-tools/cmd/controller-gen@master
> ```
> 
> and re-generate will be work

Issue is still there:  map values must be a named type, not *ast.InterfaceType

jaypipes:
Yes, this is still an issue. Something where this pops up for me is any API type that contains the [`aws.JSONValue`](https://github.com/aws/aws-sdk-go/blob/776f18cd8fff2e6abd14a0f64a1edc88669ca357/aws/jsonvalue.go#L12) type, which is simply an alias for `map[string]interface{}`.

Until this issue is addressed, I think my only solution is to manually change the CRD field's type from `aws.JSONValue` to `string` and handle the marshal/unmarshal manually within my controller.

Related: https://github.com/aws/aws-controllers-k8s/issues/119

PatrickOsborne:
[...]

PR #518, opened 2020-11-11
🐛 Allow nested maps in Go types

Porges:
These map to nested maps in CRDs. Also improve an error message.

I have marked this as a bug, as the omission of support for these nested maps seems like an oversight, since there is already support for and tests for nested maps in the deep-copy code (in `legacy_deepcopy_cases.go`).

Note that in addition a type like `map[string]Foo` where `Foo` has a field `map[string]string` is already permitted, so this only allows an “unwrapped” version of the above. (An example of this is also added into the test as part of this PR.)

This will help address some cases in #287.


Porges:
I noticed that previously an error was being generated, but it does not get surfaced anywhere and `controller-gen` does not fail. Is there something else that needs fixing here?

munnerz:
> On the other hand maybe this doesn't work with Kubernetes itself? We shall soon find out!

Using maps in Kubernetes API resources at all is not recommended (as per the sig-architecture api-conventions doc):
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

Not too sure if it's something the project still wants to accept anyway - but it'd be good for us to encourage following the api conventions doc through the feature-set exposed

Porges:
> > On the other hand maybe this doesn't work with Kubernetes itself? We shall soon find out!
> 
> Using maps in Kubernetes API resources at all is not recommended (as per the sig-architecture api-conventions doc):
> https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps
> 
> Not too sure if it's something the project still wants to accept anyway - but it'd be good for us to encourage following the api conventions doc through the feature-set exposed

In our case we are setting up configurations for external resources which require maps and nested maps in their configuration. Without support in `controller-gen` and friends we have to accept arbitrary JSON in the spec (and maybe write a custom acceptor/validator). Perhaps controller-gen could produce a warning along the lines of "maps are not recommended for use in Kubernetes APIs", or require a flag to enable support?

vijtrip2:
This will also unblock [ACK (AWS Controllers for Kubernetes)](https://github.com/aws-controllers-k8s/community) project. ACK uses controller-gen to create Kubernetes custom resources for AWS resources. There are AWS resources which contain [nested map](https://github.com/vijtrip2/aws-sdk-go/blob/6a71e1594856a4350bed5c29ba63724b66372591/models/apis/apigatewayv2/2018-11-29/api-2.json#L5019) and fails with controller-gen.

----

I also second @Porges 's thought here.
`
Perhaps controller-gen could produce a warning along the lines of "maps are not recommended for use in Kubernetes APIs", or require a flag to enable support? There is precedence for similar things where float values are “not recommended” but are permitted with a flag to enable them`

vijtrip2 (approved):
LGTM!

munnerz:
Sorry for the delay on this - for context, I do not have approval rights here :)

> In our case we are accepting specs for external resources which require maps and nested maps in their configuration.

Embedding _external_ types (i.e. ones that were not originally designed to be used as part of a Kubernetes API) is also something that has been typically best avoided. I would always strongly recommend either building your own abstraction over these types (and writing functions that can be used to convert from your API type to the external project's types), or otherwise using something like `apiextensions.JSON` as you noted.

When you begin to generate e.g. openapi schemas, conversion functions, and even DeepCopy functions, you can quickly run into all sorts of issues if you have embedded an external type. I'm conscious that by supporting it, we'd encourage this kind of thing in other projects/APIs.

I can appreciate that there are some projects _wanting_ to use maps, but I think as a project we should help users to _do the right thing_. Interested to hear viewpoints from others (cc @DirectXMan12 @pwittrock)

Porges:
The kinds of things we are embedding are configurations, so it's nice for the end user if the configuration format is the same inside Kubernetes or outside it.

----

With regards to API complexity: I'd note that you can already write `map[string]Foo` where `Foo` contains a `map[string]…` field, so allowing `map[string]map[string]...` is essentially the same, but dropping one layer of indirection. This to me seems the strongest argument for permitting it, since something “almost isomorphic” is already permitted.

DirectXMan12:
> With regards to API complexity: I'd note that you can already write map[string]Foo

This is not supposed to be possible, and I'd almost consider it a bug that we allow it.  At the very least, it's a non-reccomended type -- maps in Kubernetes APIs are generally restricted to scalar values, and the occasional slice in very specific circumstances.  Otherwise, Kubernetes-style associated lists are the reccomended pattern.

DirectXMan12:
If we're gonna support this (which, I'm not super in favor of it because it's in violation of Kubernetes API conventions), it should at least be behind a flag similar to the existing dangeroustypes flag, to turn off warnings on non-conformant APIs or something

vijtrip2:
Created a PR with unit tests for dangerousTypes and adding nested maps to DangerousTypes. Please take a look.
https://github.com/kubernetes-sigs/controller-tools/pull/569 

@Porges @DirectXMan12 @munnerz

smarterclayton:
To comment directly here, kube discourages but does not disallow certain constructs in code apis.  I would not consider any construct described here as dangerous, nor would I say our api conventions are a counter to the stated goal for CRDs to allow the kube server to accommodate new concepts to extend Kube.  As an API reviewer I might guide a core API author away from map of map, but that should not constrain how users can express valid CRD APIs in wide use.

alvaroaleman (approved):
/hold cancel
/assign @vincepri

alvaroaleman:
@Porges can you resolve the merge conflicts, please?

Porges:
> @alvaroaleman can you resolve the merge conflicts, please?

voilà!

PR #522, opened 2020-11-16
🐛 Generate plural words for Kind correctly

iawia002:
Changing `Kind` to lowercase first will cause the `Pluralize` function to misjudge word formation, and will generate wrong plurals for words with special endings, we take `TestQuota` as an example (ending with `ta`):

before:

```
TestQuota --ToLower--> testquota --will be Pluralized as testquota--> testquota
```

after:

```
TestQuota --will be Pluralized as Quota--> Testquotas --ToLower--> testquotas
```

/assign @DirectXMan12

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/fatih/color v1.7.0
-	github.com/gobuffalo/flect v0.2.0
+	github.com/gobuffalo/flect v0.2.2
 	github.com/google/go-cmp v0.3.0
 	github.com/mattn/go-colorable v0.1.2 // indirect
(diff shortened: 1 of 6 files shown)


alvaroaleman:
@iawia002 could you please provide a testcase so we don't regress here?

iawia002:
> @iawia002 could you please provide a testcase so we don't regress here?

Hi @alvaroaleman a test case has been added, PTAL

alvaroaleman (approved):
Thanks!

/lgtm

DirectXMan12:
> label tide/merge-method-squash

What sorcery is this? :mage_woman:

PR #524, opened 2020-11-19
:bug: Skipping unexpected field for CRD generation

shawn-hurley:
Unexported fields can not have a JSON tag due to go vet struct tags but currently, this requires a JSON tag for unexported fields. This skips unexported fields during CRD generation.


williammck:
FWIW, we're encountering the same issue when trying to use a struct generated from a protobuf as the Spec on a CRD. (since `state`, `sizeCache`, and `unknownFields` all exist on the generated specs)
Is there a way we can help get traction on this?

shawn-hurley:
@DirectXMan12 @joelanford  bump on reviews if at all possible?

pkg/crd/testdata/cronjob_types.go
```diff
@@ -154,13 +154,18 @@ type CronJobSpec struct {
 
 	// This tests that min/max properties work
 	MinMaxProperties MinMaxObject `json:"minMaxProperties,omitempty"`
+
+	// This is a new field that is unexported and should be ignored
+	testing string
```
estroz, line 159: Can you also add an inlined field?

joelanford: :+1: Might be good to test inlined fields from both exported and unexported struct types.

```go
type unexportedStruct struct {
	Foo string `json:"foo"`
	bar string
}

type ExportedStruct struct {
	Baz string `json:"baz"`
	qux string
}
```

And then in this struct:
```go
...
	unexportedStruct `json:",inline"`
	ExportedStruct `json:",inline"`
...
```

With this, we would expect to see `foo` and `baz`?

110y:
@shawn-hurley @estroz @joelanford 

Any updates? Since I'm facing the same problem as @williammck mentioned in this PR, I'm interested in taking over this PR.

williammck:
/remove-lifecycle stale

We're still having to use the code in this PR in order to have our CRD YAMLs generate properly from Protobuf structs.
What are next steps for getting this merged? Happy to work on the mergeflict and/or anything else needed here.

estroz:
@williammck you can open another PR if you'd like. Tests needed.

PR #528, opened 2020-11-26
:sparkles: Add crd:validation:Schemaless marker

### LINKED ISSUE #291: controller-gen doesn't work with k8s.io/apiextensions-apiserver

maxsmythe:
This marker will avoid trying to do any type detection on any struct field on which it is set. This gives users a safety valve when they hit an edge case where type inference does the wrong thing for them.

This fixes #291, which was recently re-broken by fixing #502

Signed-off-by: Max Smythe <smythe@google.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,5 @@
 run:
+  modules-download-mode: readonly
   # Increase the default deadline from 1m as some module operations can take a
   # while if uncached!
   deadline: 5m
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,4 @@
 run:
-  modules-download-mode: readonly
-
   # Increase the default deadline from 1m as some module operations can take a
   # while if uncached!
   deadline: 5m
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -26,6 +26,10 @@ import (
 	"sigs.k8s.io/controller-tools/pkg/markers"
 )
 
+const (
+	SchemalessName = "kubebuilder:validation:Schemaless"
+)
+
 // ValidationMarkers lists all available markers that affect CRD schema generation,
 // except for the few that don't make sense as type-level markers (see FieldOnlyMarkers).
 // All markers start with `+kubebuilder:validation:`, and continue with their type name.
@@ -82,6 +86,9 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:validation:EmbeddedResource", markers.DescribesField, XEmbeddedResource{})).
 		WithHelp(XEmbeddedResource{}.Help()),
+
+	must(markers.MakeDefinition(SchemalessName, markers.DescribesField, Schemaless{})).
+		WithHelp(Schemaless{}.Help()),
 }
 
 // ValidationIshMarkers are field-and-type markers that don't fall under the
@@ -224,6 +231,16 @@ type XPreserveUnknownFields struct{}
 // running apiserver. It is not necessary to add any additional schema for these
 // field, yet it is possible. This can be combined with PreserveUnknownFields.
 type XEmbeddedResource struct{}
+
+// +controllertools:marker:generateHelp:category="CRD validation"
+// Schemaless marks a field as being a schemaless object.
+//
+// Schemaless objects are not introspected, so you must provide
+// any type and validation information yourself. One use for this
+// tag is for embedding fields that hold JSONSchema typed objects.
+// Because this field disables all type checking, it is recommended
+// to be used only as a last resort.
+type Schemaless struct{}
 
 func (m Maximum) ApplyToSchema(schema *apiext.JSONSchemaProps) error {
 	if schema.Type != "integer" {
--- a/pkg/crd/markers/zz_generated.markerhelp.go
+++ b/pkg/crd/markers/zz_generated.markerhelp.go
@@ -306,6 +306,17 @@ func (Resource) Help() *markers.DefinitionHelp {
 	}
 }
 
+func (Schemaless) Help() *markers.DefinitionHelp {
+	return &markers.DefinitionHelp{
+		Category: "CRD validation",
+		DetailedHelp: markers.DetailedHelp{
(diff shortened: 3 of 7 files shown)


lkysow:
We have a situation where we want to model `map[string]interface{}`, e.g.
```yaml
spec:
  config: # we want this to be map[string]interface
    foo: '{some-fun-json}'
    members: 3
```

Currently we're using:
```go
	Config json.RawMessage `json:"config,omitempty"`
```

And then converting to `map[string]interface{}`.
```go
func (in *ProxyDefaults) convertConfig() map[string]interface{} {
	if in.Spec.Config == nil {
		return nil
	}
	var outConfig map[string]interface{}
	json.Unmarshal(in.Spec.Config, &outConfig)
	return outConfig
}
```

The only CRD type we've found that works is `object`:
```yaml
          properties:
            config:
              description: yada
              type: object
```

However the CRD generation gives us:
```yaml
            config:
              description: yada
              format: byte
              type: string
```

If we use the annotation
```go
	// +kubebuilder:validation:Type=object
	Config json.RawMessage `json:"config,omitempty"`
```

We get an error:
```
github.com/hashicorp/consul-k8s/api/v1alpha1:-: conflicting types in allOf branches in schema: string vs object
```

Would this change allow us to do:
```go
	// +kubebuilder:validation:Type=object
        // +kubebuilder:validation:Schemaless
	Config json.RawMessage `json:"config,omitempty"`
```
?

maxsmythe:
It should be able to.

Some alternatives for you:

- Use [`runtime.RawExtension`](https://github.com/open-policy-agent/gatekeeper/blob/57dabe230d214a8f07ef9a1dfb837dcf786a7976/apis/mutations/v1alpha1/assign_types.go#L52) to wrap an object one level higher, and user XPreserveUnknownFields to allow any field to be set on that object. You will need to do the serialization/deserialization yourself if you do this.
- Consider a different API design, such as maps of objects as accepting arbitrary fields is against the [K8s API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps), though this is certainly not universally applied.

If neither of these alternatives work for you, this PR would be helpful.

.golangci.yml
```diff
@@ -1,6 +1,4 @@
 run:
-  modules-download-mode: readonly
```
DirectXMan12, line 2: ??

DirectXMan12 (changes requested):
minor comment on a go modules file change, otherwise lgtm

.golangci.yml
maxsmythe: That had fixed an error I was having running the linter. It's been a while, so I can't remember what the error was.

Testing again with this line re-added doesn't cause a problem anymore. I wonder if it only breaks if your go modules aren't already up-to-date?

In any case, not really in scope for this PR, so reverting.

maxsmythe:
Thanks!

PR #498, opened 2020-10-01
:book: Add versioning guidelines

DirectXMan12:
This adds versioning/releasing guidelines that reference the common
guidelines in kubebuilder-release-tools, with a few
controller-tools-specific reminders.

PR #539, opened 2021-01-27
⚠ Generate Embedded ObjectMeta in the CRDs

dvaldivia:
This PR makes it so embedded `ObjectMeta` in the CRD get's properly generated if the generator option `generateEmbeddedObjectMeta=true` is passed, this is needed because if a CRD has embedded `ObjectMeta` in any field and `preserveUnknowFields` is set to false, all the metadata will be lost when doing conversion between versions.

An example on how to have the embedded ObjectMeta generated in the resulting CRD

```
controller-gen crd:maxDescLen=0,generateEmbeddedObjectMeta=true paths=. output:dir=.
```

This PR makes it so by default any embedded `ObjectMeta` is not generated in the resulting CRD, however the top level `ObjectMeta` belonging to the CRD itself is never generated as the kubernetes API disallows changes to the CRD metadata between conversions.

The generated `ObjectMeta` is also only a subset of the original set of fields inside `ObjectMeta` this is due to the fact that other runtime fields are problematic if they are being traded with the kubernetes API, such as `creationTimeStamp` (https://github.com/rancher/rancher/issues/23857) so this only generates `name, namespace, labels, annotations and finalizers` which from a design perspective should be enough. This follows the recommendation by @sttts who recommended a `EmbeddedObjectMeta` but instead of using a different type, we have a reduced set of fields that are "just enough" for a CRD design.

An example of why we need this is for example if a CRD had a `volumeClaimTemplate` (for an underlying statefulset) which include `ObjectMeta` such as Labels, Annotations and/org name which are meant to be passed to the PVC where getting lost between conversions.

This PR is based on the work by @arjunrn and @champak following the discussion on #448 and the PR #498 and it's also similar to #395 as it addressed the same problem, but adds the type casting for `FieldsV1` and the requested argument to control this feature (per @DirectXMan12 on #395)

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -34,8 +34,6 @@ var KnownPackages = map[string]PackageOverride{
 	},
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1": func(p *Parser, pkg *loader.Package) {
-		// ObjectMeta is managed by the Kubernetes API server, so no need to
-		// generate validation for it.
 		p.Schemata[TypeIdent{Name: "ObjectMeta", Package: pkg}] = apiext.JSONSchemaProps{
 			Type: "object",
 		}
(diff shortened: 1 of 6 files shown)


dvaldivia:
I updated the PR description

pkg/crd/known_types.go
```diff
@@ -113,6 +111,52 @@ var KnownPackages = map[string]PackageOverride{
...
 }
 
+// ObjectMetaPackages overrides the ObjectMeta in all types
+var ObjectMetaPackages = map[string]PackageOverride{
+	"k8s.io/apimachinery/pkg/apis/meta/v1": func(p *Parser, pkg *loader.Package) {
+		// execute the KnowPackages for `k8s.io/apimachinery/pkg/apis/meta/v1` if any
+		if f, ok := KnownPackages["k8s.io/apimachinery/pkg/apis/meta/v1"]; ok {
+			f(p, pkg)
+		}
+		// This is a allow-listed set of properties of ObjectMeta, other runtime properties are not part of this list
```
DirectXMan12, line 121: as per the meeting today, please link to sttts's comment here

DirectXMan12 (changes requested):
really minor comment as per the meeting today, otherwise looks good

pkg/crd/known_types.go
coderanger: To clarify, I think @DirectXMan12 is referring to this comment https://github.com/kubernetes-sigs/controller-tools/pull/395#issuecomment-691919433. Can you confirm?

DirectXMan12: yeah, that's what I was talking about.

dvaldivia: @DirectXMan12 @coderanger I'll update the PR

dvaldivia: done, rebased, squashed and comment added.

DirectXMan12:
subsumed by #557 (with attribution :-) )

PR #541, opened 2021-01-28
✨ Add deprecated marker for versions

dvaldivia:
This PR introduces the marker `deprecated` which would mark a CRD version as deprecated in the generated output, it supports a deprecation warning message

Example  usage: 
```
// +kubebuilder:deprecated:warning="This version is deprecated in favor of other.crd.com/v2"
```

the result is that the version that has this marker would include the following fields
```
deprecated: true
deprecatedWarning: "This version is deprecated in favor of other.crd.com/v2"
```

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -48,6 +48,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:unservedversion", markers.DescribesType, UnservedVersion{})).
 		WithHelp(UnservedVersion{}.Help()),
+
+	must(markers.MakeDefinition("kubebuilder:deprecatedversion", markers.DescribesType, DeprecatedVersion{})).
+		WithHelp(DeprecatedVersion{}.Help()),
 }
 
 // TODO: categories and singular used to be annotations types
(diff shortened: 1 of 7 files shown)


pkg/crd/markers/crd.go
```diff
@@ -316,3 +319,29 @@ func (s UnservedVersion) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, ve
 }
 
 // NB(directxman12): singular was historically distinct, so we keep it here for backwards compat
+
+// +controllertools:marker:generateHelp:category=CRD
+
+// DeprecatedVersion marks this version as deprecated.
+type DeprecatedVersion struct {
+	// Warning message to be shown on the deprecated version
+	Warning string
```
DirectXMan12, line 328: is this optional, or is it required in the actual CRD.  If optional, you can make it a pointer or do something like

```go
type DeprecatedVersion struct {
   Warning string `marker:",optional"`
}
```

So that the equivalent field is also optional, meaning either of the following would work:

```go
// +kubebuilder:deprecated
type SomeKind struct {}

// {  deprecated: true }
```

```go
// +kubebuilder:deprecated:warning="blah"
type SomeKind struct{}

// { deprecated: true, deprecationWarning: "blah" }
```

pkg/crd/markers/crd.go
```diff
@@ -48,6 +48,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:unservedversion", markers.DescribesType, UnservedVersion{})).
 		WithHelp(UnservedVersion{}.Help()),
+
+	must(markers.MakeDefinition("kubebuilder:deprecated", markers.DescribesType, DeprecatedVersion{})).
```
DirectXMan12, line 52: should this be `deprecatedversion` to match with `unservedversion` and in case we ever get field-level deprecation, or do we think we can just use the same marker for field-level deprecation if we ever get it?

DirectXMan12 (commented):
generally looks good, minor feedback

pkg/crd/markers/crd.go
dvaldivia: it's optional, but the CRD for CRD also takes a pointer for the deprecation message, so I added the marker optional and also made it a pointer

pkg/crd/markers/crd.go
dvaldivia: it makes sense to keep them consistent, I changed it to `deprecatedversion`

DirectXMan12:
pushed a test & squashed

/lgtm
/approve

DirectXMan12:
thanks, and sorry for the lag!

dvaldivia:
@DirectXMan12 no problem, any time

PR #549, opened 2021-03-19
🐛 markers: Allow floats inside markers

christarazi:
This fixes the following error when controller-gen is invoked as

```
$ controller-gen crd paths=./crds output:crd:artifacts:config=/tmp/0x123
```

```
Error: unable to parse option
"output:crd:artifacts:config=/tmp/tmp.0x123": [hexadecimal literal has no digits (at <input>:1:17)]
```

Signed-off-by: Chris Tarazi <tarazichris@gmail.com>

### DIFF
--- a/pkg/markers/parse.go
+++ b/pkg/markers/parse.go
@@ -755,7 +755,7 @@ func (d *Definition) loadFields() error {
 func parserScanner(raw string, err func(*sc.Scanner, string)) *sc.Scanner {
 	scanner := &sc.Scanner{}
 	scanner.Init(bytes.NewBufferString(raw))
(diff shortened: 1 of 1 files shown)


DirectXMan12:
a) I'm not entirely certain how this is how we should approach that issue
b) can you post reproduction steps and environment please?  I'm unable to reproduce the issue with the sample you gave (tried on both zsh & bash, versions 5.8 & 5.1.4(1), respectively).  Is your shell doing something weird, perhaps?

/hold

christarazi:
@DirectXMan12 Hmm, I'm unable to reproduce for some reason. I'll close this for now and if it reoccurs, then we can revisit this.

PR #555, opened 2021-04-07
✨feature(master): support  mutating name and validating name for webhook

cuisongliu:
Hello Kubernetes Community!

This PR support customize ValidatingWebhookConfiguration and MutatingWebhookConfiguration name.

ex:

```controller-gen  webhook:mutatingName=mutating-config,validatingName=validating-config```

Have a nice day

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -282,7 +282,12 @@ func (c Config) webhookVersions() ([]string, error) {
 // +controllertools:marker:generateHelp
 
 // Generator generates (partial) {Mutating,Validating}WebhookConfiguration objects.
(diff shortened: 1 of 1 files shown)


munnerz:
This code seems correct, but it'd be good to understand the actual motivation behind this change. Would a kustomize overlay be more appropriate for renaming these resources, rather than markers passed to `controller-gen`? Is there a reason this would _not_ suit your use case?

munnerz:
(also: welcome to the project! 😄)

cuisongliu:
I think the kustomize is override the  additional attributes, is not override like name  the main feild.

PR #557, opened 2021-04-15
⚠ Generate Embedded ObjectMeta in the CRDs, plus addressed comments

BlaineEXE:
This PR addresses issue comments from #539, notably linking to a comment by @sttts. The original description of #539 is copied below.

This is desired for an issue we are seeing in Rook with CRD gen: https://github.com/rook/rook/issues/7629
We wish to keep annotations in embedded object meta.

--------

This PR makes it so embedded `ObjectMeta` in the CRD get's properly generated if the generator option `generateEmbeddedObjectMeta=true` is passed, this is needed because if a CRD has embedded `ObjectMeta` in any field and `preserveUnknowFields` is set to false, all the metadata will be lost when doing conversion between versions.
 
An example on how to have the embedded ObjectMeta generated in the resulting CRD

```
controller-gen crd:maxDescLen=0,generateEmbeddedObjectMeta=true paths=. output:dir=.
```
 
This PR makes it so by default any embedded `ObjectMeta` is not generated in the resulting CRD, however the top level `ObjectMeta` belonging to the CRD itself is never generated as the kubernetes API disallows changes to the CRD metadata between conversions.

The generated `ObjectMeta` is also only a subset of the original set of fields inside `ObjectMeta` this is due to the fact that other runtime fields are problematic if they are being traded with the kubernetes API, such as `creationTimeStamp` ([rancher/rancher#23857](https://github.com/rancher/rancher/issues/23857)) so this only generates `name, namespace, labels, annotations and finalizers` which from a design perspective should be enough. This follows the recommendation by @sttts who recommended a `EmbeddedObjectMeta` but instead of using a different type, we have a reduced set of fields that are "just enough" for a CRD design.

An example of why we need this is for example if a CRD had a `volumeClaimTemplate` (for an underlying statefulset) which include `ObjectMeta` such as Labels, Annotations and/org name which are meant to be passed to the PVC where getting lost between conversions.

This PR is based on the work by @arjunrn and @champak following the discussion on #448 and the PR #498 and it's also similar to #395 as it addressed the same problem, but adds the type casting for `FieldsV1` and the requested argument to control this feature (per @DirectXMan12 on #395)

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -286,9 +286,7 @@ func FixTopLevelMetadata(crd apiext.CustomResourceDefinition) {
 				schemaProperties["metadata"] = apiext.JSONSchemaProps{Type: "object"}
 			}
 		}
-
 	}
-
 }
 
 // toTrivialVersions strips out all schemata except for the storage schema,
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -84,6 +84,9 @@ type Generator struct {
 	// You'll need to use "v1" to get support for features like defaulting,
 	// along with an API server that supports it (Kubernetes 1.16+).
 	CRDVersions []string `marker:"crdVersions,optional"`
+
+	// GenerateEmbeddedObjectMeta specifies if any embedded ObjectMeta in the CRD should be generated
+	GenerateEmbeddedObjectMeta *bool `marker:",optional"`
 }
 
 func (Generator) CheckFilter() loader.NodeFilter {
@@ -98,6 +101,8 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		Checker:   ctx.Checker,
 		// Perform defaulting here to avoid ambiguity later
 		AllowDangerousTypes: g.AllowDangerousTypes != nil && *g.AllowDangerousTypes == true,
+		// Indicates the parser on whether to register the ObjectMeta type or not
+		GenerateEmbeddedObjectMeta: g.GenerateEmbeddedObjectMeta != nil && *g.GenerateEmbeddedObjectMeta == true,
 	}
 
 	AddKnownTypes(parser)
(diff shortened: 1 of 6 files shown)


pkg/crd/gen.go
```diff
@@ -224,6 +232,20 @@ func removeDefaultsFromSchemaProps(v *apiextlegacy.JSONSchemaProps) {
...
+// FixTopLevelMetadata resets the schema for the top-level metadata field which is needed for CRD validation
+func FixTopLevelMetadata(crd apiext.CustomResourceDefinition) {
+	for _, v := range crd.Spec.Versions {
+		if v.Schema != nil && v.Schema.OpenAPIV3Schema != nil && v.Schema.OpenAPIV3Schema.Properties != nil {
+			schemaProperties := v.Schema.OpenAPIV3Schema.Properties
+			if _, ok := schemaProperties["metadata"]; ok {
+				schemaProperties["metadata"] = apiext.JSONSchemaProps{Type: "object"}
+			}
+		}
+
```
leseb, line 244: nit blank line

pkg/crd/gen.go
```diff
@@ -224,6 +232,20 @@ func removeDefaultsFromSchemaProps(v *apiextlegacy.JSONSchemaProps) {
...
+	}
+
```
leseb, line 246: ditto

DirectXMan12 (approved):
/lgtm
/approve

jMarkP:
Just came across the same issue and it's so nice to find the fix here all ready to go! Thanks for this @BlaineEXE :)

pkg/crd/known_types.go
```diff
@@ -113,6 +111,53 @@ var KnownPackages = map[string]PackageOverride{
...
+				},
+				"labels": apiext.JSONSchemaProps{
+					Type: "object",
+					AdditionalProperties: &apiext.JSONSchemaPropsOrBool{
+						Schema: &apiext.JSONSchemaProps{
+							Type: "string",
+						},
+					},
+				},
+				"finalizers": apiext.JSONSchemaProps{
```
sttts, line 148: this is odd. Why finalizers?

jessefeinman: I had the same thought, I'd have expected either all the ObjectMeta fields or the non-system fields here https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/

sbueringer: @jessefeinman Can you open an issue to discuss this?

jessefeinman: https://github.com/kubernetes-sigs/controller-tools/issues/805

PR #563, opened 2021-04-26
✨ Generate embedded ObjectMeta with schemapatch as well

markusthoemmes:
Quite similar to https://github.com/kubernetes-sigs/controller-tools/pull/557, this applies the very same options now available on the CRD generator to be available on the `schemapatch` command. Since this option only really affects the generation of the schema anyway, it seems like it should be available on both equally.

### DIFF
--- a/pkg/schemapatcher/gen.go
+++ b/pkg/schemapatcher/gen.go
@@ -84,6 +84,9 @@ type Generator struct {
 	// n indicates limit the description to at most n characters and truncate the description to
 	// closest sentence boundary if it exceeds n characters.
 	MaxDescLen *int `marker:",optional"`
+
+	// GenerateEmbeddedObjectMeta specifies if any embedded ObjectMeta in the CRD should be generated
+	GenerateEmbeddedObjectMeta *bool `marker:",optional"`
 }
 
 var _ genall.Generator = &Generator{}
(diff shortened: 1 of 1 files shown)


pkg/schemapatcher/gen.go
```diff
@@ -142,6 +147,12 @@ func (g Generator) Generate(ctx *genall.GenerationContext) (result error) {
 				fullSchema = *fullSchema.DeepCopy()
 				crdgen.TruncateDescription(&fullSchema, *g.MaxDescLen)
 			}
+
+			// Fix top level ObjectMeta regardless of the settings.
+			if _, ok := fullSchema.Properties["metadata"]; ok {
```
DirectXMan12, line 152: this doesn't seem quite right, because top-level metadata is actually allowed to contain a couple validations, and this is one of the things schemapatch enables.

DirectXMan12 (changes requested):

pkg/schemapatcher/gen.go
markusthoemmes: Interesting, I basically copied the approach over from the `crd` generator, where this exact change is done unconditionally as well :thinking:. 

See

https://github.com/kubernetes-sigs/controller-tools/blob/97ee66cc83b03bd632cf8da35496812ca091ae4e/pkg/crd/gen.go#L137-L138

and

https://github.com/kubernetes-sigs/controller-tools/blob/97ee66cc83b03bd632cf8da35496812ca091ae4e/pkg/crd/gen.go#L281-L290

Could you elaborate on why this should be different in `schemapatch`?

DirectXMan12: oh hmm this might've been me reviewing too fast.  My thought was that maybe people had put in place custom validation, but schemapatch is gonna overwrite that anyway.  nevermind :-)

PR #566, opened 2021-05-03
🐛 Add format for float types

tamalsaha:

https://swagger.io/docs/specification/data-models/data-types/#numbers

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -427,6 +427,10 @@ func builtinToType(basic *types.Basic, allowDangerousTypes bool) (typ string, fo
 		format = "int32"
 	case types.Int64, types.Uint64:
 		format = "int64"
+	case types.Float32:
+		format = "float"
+	case types.Float64, types.UntypedFloat:
+		format = "double"
 	}
 
 	return typ, format, nil


alvaroaleman:
Relevant issue is https://github.com/kubernetes-sigs/controller-tools/issues/245 and it doesn't seem clear if we should allow this or not. The kube api conventions excplicitly mention not to allow floats in spec: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

/hold

matthchr:
@alvaroaleman possibly worth seeing discussion on #449 for history.

pkg/crd/schema.go
```diff
@@ -427,6 +427,10 @@ func builtinToType(basic *types.Basic, allowDangerousTypes bool) (typ string, fo
 		format = "int32"
 	case types.Int64, types.Uint64:
 		format = "int64"
+	case types.Float32:
+		format = "float"
+	case types.Float64, types.UntypedFloat:
+		format = "double"
```
camilamacedo86, line 433: Hi @tamalsaha, 

That is nice. But did you check the comment on top of this method? See:

```
// builtinToType converts builtin basic types to their equivalent JSON schema form.
// It *only* handles types allowed by the kubernetes API standards. Floats are not
// allowed unless allowDangerousTypes is true
```
 
In this way, I understand that we cannot add the floats here.

PR #584, opened 2021-07-13
:bug: Skip unexported fields when generating CRDs

williammck:
Unexported fields currently cause CRD YAML generation to fail. Since these fields can't and don't have JSON tags, `structToSchema` adds errors due to the missing tags.

Example output when using structs generated from a protobuf:
(error text [originates here](https://github.com/kubernetes-sigs/controller-tools/blob/ffa54949a74756bbb4d5c41ae4fe8d6fea759ae7/pkg/crd/schema.go#L338))
```
encountered struct field "state" without JSON tag in type "Storage"
encountered struct field "sizeCache" without JSON tag in type "Storage"
encountered struct field "unknownFields" without JSON tag in type "Storage"
```

This adds tests for @shawn-hurley's original work in https://github.com/kubernetes-sigs/controller-tools/pull/524.

I've also added a generator flag so this isn't enabled by default, as this is more of an advanced feature.
e.g. `controller-gen crd:trivialVersions=true,ignoreUnexportedFields=true`

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -41,6 +41,11 @@ const defaultVersion = v1
 
 // Generator generates CustomResourceDefinition objects.
 type Generator struct {
+	// IgnoreUnexportedFields indicates that we should skip unexported fields.
+	//
+	// Left unspecified, the default is false.
+	IgnoreUnexportedFields *bool `marker:",optional"`
+
 	// AllowDangerousTypes allows types which are usually omitted from CRD generation
 	// because they are not recommended.
 	//
@@ -85,7 +90,8 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		Collector: ctx.Collector,
 		Checker:   ctx.Checker,
 		// Perform defaulting here to avoid ambiguity later
-		AllowDangerousTypes: g.AllowDangerousTypes != nil && *g.AllowDangerousTypes == true,
+		IgnoreUnexportedFields: g.IgnoreUnexportedFields != nil && *g.IgnoreUnexportedFields == true,
+		AllowDangerousTypes:    g.AllowDangerousTypes != nil && *g.AllowDangerousTypes == true,
 		// Indicates the parser on whether to register the ObjectMeta type or not
 		GenerateEmbeddedObjectMeta: g.GenerateEmbeddedObjectMeta != nil && *g.GenerateEmbeddedObjectMeta == true,
 	}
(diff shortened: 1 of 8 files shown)


williammck:
@DirectXMan12 @droot bump on review? this is a followup of https://github.com/kubernetes-sigs/controller-tools/pull/524, @estroz and @joelanford called out that Shawn's PR needed some tests.

joelanford:
Thanks for following up on this @williammck. I have less working knowledge in controller-tools as I do in kubebuilder/controller-runtime, so I'll defer to others with more experience (btw @ DirectXMan12 has moved on to other projects, so we'll have to find some other reviewers).

One concern that comes to mind for me is that this change (if I understand it correctly) will now start silently ignoring unexported fields. It's very possible that beginners in this space have seen the error message you're seeing and realized the fix was to export the field and add a json tag. 

I just wonder if we should put up some guardrails for this. For example, we could require users to opt in to ignoring unexported fields, or maybe we could log a warning when skipping unexported fields?

williammck:
> Thanks for following up on this @williammck. I have less working knowledge in controller-tools as I do in kubebuilder/controller-runtime, so I'll defer to others with more experience (btw @ DirectXMan12 has moved on to other projects, so we'll have to find some other reviewers).

Thanks for the detail there, I appreciate it. If there's any way I can assist, let me know.

> One concern that comes to mind for me is that this change (if I understand it correctly) will now start silently ignoring unexported fields. It's very possible that beginners in this space have seen the error message you're seeing and realized the fix was to export the field and add a json tag.
> 
> I just wonder if we should put up some guardrails for this. For example, we could require users to opt in to ignoring unexported fields, or maybe we could log a warning when skipping unexported fields?

Ah, that's a good point... Since this is probably a less-common use case, I'd lean towards adding a flag. (e.g. `controller-gen crd:trivialVersions=true,ignoreUnexportedFields=true`)
I'm about to be away for the next week or so but can tinker with adding that once I'm back.

williammck:
@joelanford Sorry for the delay on that, just got around to implementing the flag.
Have you found any new reviewers? Let me know if there's any way I can help push that along.

110y:
As @williammck mentioned, I'm also longing this kind of changes regarding unexported fields.
My case is same as @williammck describes in the description, I'd like to use `struct` generated by `protoc` with `controller-gen`.

As [Istio](https://github.com/istio/istio) team do, I think writing CRDs as Protobuf and using structs generated by protoc is useful in terms of clearly declaring schemas.

williammck:
Is there still any chance this could be addressed? I understand it's a small use case, but there are people out there who want this - see the mentioned PRs and issues in this discussion.
@alvaroaleman @joelanford @vincepri thoughts?

alvaroaleman (approved):
thanks!

PR #590, opened 2021-07-31
🐛 WIP: Force preserveUnknownFields=false for v1 CRDs

### LINKED ISSUE #476: preserveUnknownFields=false doesn't work with crdVersions=v1

kevindelgado:
This is a proof-of-concept of how to force controller-gen to explicitly set "preserveUnknownFields=false" for v1 CRDs. We should always be setting this field, because the default value has flipped from true to false between CRD api version v1beta1 and v1.

Fixes #476 
Related: https://kubernetes.slack.com/archives/CAR30FCJZ/p1611919182225100

The way this works is by creating a new type that is identical to `apiext.CustomResourceDefinition` with the one exception being that the json tag for `Spec.PreserveUnknownFields` does not contain omitempty.

It feels pretty hacky to have to create a separate type that duplicates most of the CRD type, so I am definitely open to alternative ways of doing this.

Some other approaches could be:
* trying to dynamically edit the json tag of `apiext.CustomResourceDefinition`
* post-processing the yaml that gets written out to add the field on afterwards.

Both of these options have their own drawbacks and I went with this one because it seemed like the quickest way to get a working POC so that I could get a discussion started at least.

### DIFF
--- a/pkg/crd/crd_alias_types.go
+++ b/pkg/crd/crd_alias_types.go
@@ -1,0 +1,69 @@
+package crd
+
+import (
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// v1AliasCRD is an alias of v1.CustomResourceDefinition.
+// It has identical fields, the only difference is the json tag
+// on spec.PreserveUnknownFields does not contain omitempty.
+//
+// The source of truth of v1.CustomResourceDefinition lives at
+// https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go
+//
+// This is used in order to force controller-gen to generate
+// v1 CRD yamls with preserveUnknownFields=false explicitly, because
+// otherwise controller will omit it which becomes problematic for users
+// upgrading existing CRDs from v1beta to v1
+//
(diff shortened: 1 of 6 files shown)


kevindelgado:
cc @munnerz based on your convo at https://kubernetes.slack.com/archives/CAR30FCJZ/p1611919182225100

Issue #514, opened 2020-11-04
newline discrepancy in CRDs generated by controller-gen


marcoderama:
Using controller-gen v0.4.0 (and older versions) for CRD generation on ubuntu, long descriptions are all on one line.  On a Mac, they're broken across multiple lines.  While both are equivalent from a YAML perspective, they're very different as far as git is concerned.  This means if dev A on ubuntu commits a CRD, and then dev B on MacOS makes a change that is effectively a small change to the yaml, the git diff is huge and completely obscures the actual change.  (And obviously the same happens if the order is dev B then dev A making changes.)

This makes it very difficult to review/understand CRD diffs and confuses devs.

Is there a way to make the outputs consistent?  If not, then I'd consider this a bug.

As an example, here's what happens for one of the descriptions in our CRD when going from generated-on-Mac to generated-on-ubuntu.  There is no diff yaml-wise but `git diff` includes:
```
           kind:
-            description: 'Kind is a string value representing the REST resource this
-              object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
```

marcoderama:
Fwiw, I tried

derekperkins:
We have the same issues in v0.3.0

marcoderama:
I guess we can just run the controller-gen output through a YAML pretty-printer before committing it.  It would seem reasonable to assume that's not necessary though.

marcoderama:
This happens after running `make manifests`, as setup by kubebuilder.  I assumed it was due to `controller-gen` but it turns out the culprit is `kustomize`.
c.f. https://github.com/kubernetes-sigs/kustomize/issues/947

Sorry for casting aspersions your way.

(And working around it isn't a simple matter of piping things through `yq` (yet).  See https://github.com/mikefarah/yq/issues/573)

zhangtbj:
Hi all,

I also found this problem on my Mac (without line-breaking), but my collea all generate WITH line-breaking.

I also tried on my Ubuntu VM, it also generates with line-breaking.

[...]

PR #593, opened 2021-08-15
:seedling: Regenerate CRD Integration Test testdata

erikgb:
I am considering submitting a functional PR, and had a look in the codebase - to get a bit familiar. Since I am probably going to suggest a new CRD marker, I took special note on the [process for generating testdata](https://github.com/kubernetes-sigs/controller-tools/tree/master/pkg/crd/testdata#crd-integration-test-testdata).

And it seems like the "golden output file" hasn't been re-generated lately. At least I get a big diff related to newlines in the desciption fields. Possibly related issue: #514.

### DIFF
--- a/pkg/crd/testdata/plural/plural.example.com_testquotas.yaml
+++ b/pkg/crd/testdata/plural/plural.example.com_testquotas.yaml
@@ -21,10 +21,14 @@ spec:
       openAPIV3Schema:
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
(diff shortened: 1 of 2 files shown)


erikgb:
@mengqiy @pwittrock Would it make sense to split this PR? One for generating the missing help message (🐛) and one for re-generating the test resources (🌱)? And do you require an issue for each PR?

erikgb:
> @mengqiy @pwittrock Would it make sense to split this PR? One for generating the missing help message (🐛) and one for re-generating the test resources (🌱)? And do you require an issue for each PR?

I think so, and created https://github.com/kubernetes-sigs/controller-tools/pull/597 for the bugfix.

erikgb:
/assign @mengqiy

Any chance to have a review on this? It is "just" a re-generation of what appears to be outdated files - and no "real" changes - just YAML reformatting. This is kinda blocking any progress on my functional PR....

camilamacedo86:
Hi @erikgb, 

So, the testdata has not been checked if is updated for each PR? 
If that is the case, IHMO the solution should be we start to check that in the CI as we do in Kubebuilder. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/testdata.yml

erikgb:
@camilamacedo86 Thx for taking a look at this old PR!

> So, the testdata has not been checked if is updated for each PR?

No, it seems like some PRs have been merged without update to the testdata.

> If that is the case, IHMO the solution should be we start to check that in the CI as we do in Kubebuilder. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/testdata.yml

I totally agree! I will try to update this PR with a GHA-job that verifies that the generated code is up to date. Will ping you for a new review when that is done.

PR #598, opened 2021-08-23
:running: feat(dockerfile): add binary image

georgettica:
# What
this PR adds dockerfile support to allow for execution of these tools in a container and not directly on the host

# Why
We have `controller-gen` installed on our CI server and migrating to the newer version there might break stuff for other repos running on that server

thus adding an image allows us to run it and not the binary on the host

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY cmd/ ./cmd
 
 # build all of the binaries and show they were built
 RUN for binary_folder in $(find cmd -maxdepth 1 -mindepth 1); do \
-		go build -o $(basename $binary_folder) $binary_folder/*; \
+		CGO_ENABLED=0 go build -o $(basename $binary_folder) $binary_folder/*; \
 	done ; \
 	find . -type f -executable
 ##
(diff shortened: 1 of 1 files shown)


georgettica:
/hold
I need help here because for some reason the dockerfile is not allowing me to run the binaries using this second builder image

if you could help me with that I will remove the hold

georgettica:
seems the issue is I need a `musl-libc` package.. working on adding it to a dev container

georgettica:
the binary spews out for now:
```
Error: err: go command required, not found: exec: "go": executable file not found in $PATH: stderr:
```

georgettica:
seems that even if I have golang, the build becomes:
```
controller-gen --version
Version: (devel)
```
as explained in https://github.com/kubernetes-sigs/controller-tools/blob/ea8d4ead3e34dac7bb792cedc6d3cd7decf63137/pkg/version/version.go#L35-L46

camilamacedo86:
Hi @georgettica, 

I am not sure if I understand the motivation here.
Why is required a container for controller-gen?

In the ci, you can go get the wished version for the controller-gen and use it.

georgettica:
It is required for testing in a server where I cannot control the go version or the controller-gen version I want to install.
It's a multi-tenant server and I would like to pull the required version for my code to run

georgettica:
@camilamacedo86 could you elaborate on the approach of getting the version and not breaking other users of co troller-gen?

georgettica:
Closing as this has not got any traction for a while :(

PR #602, opened 2021-08-27
:running: Add makefile so we can keep testcommands in the repo

alvaroaleman:
Requirement for https://github.com/kubernetes/test-infra/pull/23393

This adds a very simplistic Makefile as preparation of running tests
with Prow (right now we don't run any tests at all). The makefile has
the advantage that we can change/extend what we test and get feedback
for that, whereas changing Prow config doesn't give any feedback.

/assign @vincepri


vincepri (approved):
/approve
/lgtm

Issue #609, opened 2021-09-01
Allow skipping unexported fields in CRD generation


bcho:
I am using controller-gen in a project that, needs to generate CRD from existing struct with unexported fields. The simplest 
 real world setup is like:

```go
// In our application, we want to reuse the protobuf generated struct as part of the CRD spec.
// In this case, we might have a spec struct like this:
type ResourceSpec struct {
  // internal fields used by protobuf
  state protoimpl.MessageState
  sizeCache protoimpl.SizeCache
  unknownFields protoimpl.UnknownFields

  ExportedField string `protobuf:"bytes,1,opt,name=exported_field,proto3" json:"exported_field,omitempty"`
}

//+kubebuilder:object:root=true

// the custom resource struct
type Resource struct {
  metav1.TypeMeta `json:",inline"`
  metav1.ObjectMeta `json:"metadata,omitempty"

 Spec ResourceSpec `json:"spec,omitempty"
}
```

If we use controller-gen to generate CRD for above settings, it would fail with error like: 

> encountered struct field "state" without JSON tag in type "ResourceSpec"

But I think since those fields are unexported and without JSON tag, we could safely ignore them from the CRD generation.

bcho:
I had created a draft PR for implement this, please let me know your thoughts of this feature 😸

PR #610, opened 2021-09-01
✨ feat: add support for skipping unexported fields without JSON tag

bcho:
For #609

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -342,8 +342,13 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiext.JSON
 	}
 
 	for _, field := range ctx.info.Fields {
+		isExported := ast.IsExported(field.Name)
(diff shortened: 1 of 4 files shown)


pkg/crd/schema.go
```diff
@@ -342,8 +342,13 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiext.JSON
...
 
 	for _, field := range ctx.info.Fields {
+		isExported := ast.IsExported(field.Name)
 		jsonTag, hasTag := field.Tag.Lookup("json")
-		if !hasTag {
+		if !hasTag && !isExported {
+			// if the field doesn't have a JSON tag and is unexported, skip from output
+			continue
+		}
+		if !hasTag && isExported {
```
Porges, line 351: test is duplicated directly after; second if block should be removed?

bcho:
close in favor of #584

PR #617, opened 2021-09-16
✨ Allow maps of arrays of non-string types

Porges:
Lifts the restriction currently imposed on map value types (that if they are arrays they must be arrays of `string`s), per https://github.com/kubernetes-sigs/controller-tools/pull/518#issuecomment-920387673. We have encountered this while trying to use `controller-gen` (https://github.com/Azure/azure-service-operator/issues/1489).

Also add a way to add simple schema-transform tests, rather than by adding items to `cronjob_types.go` (where the output is validated by inspection).

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -39,12 +39,10 @@ const (
 	defPrefix = "#/definitions/"
 )
 
-var (
-	// byteType is the types.Type for byte (see the types documention
-	// for why we need to look this up in the Universe), saved
-	// for quick comparison.
-	byteType = types.Universe.Lookup("byte").Type()
-)
+// byteType is the types.Type for byte (see the types documention
+// for why we need to look this up in the Universe), saved
+// for quick comparison.
+var byteType = types.Universe.Lookup("byte").Type()
 
 // SchemaMarker is any marker that needs to modify the schema of the underlying type or field.
 type SchemaMarker interface {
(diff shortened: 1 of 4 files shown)


pkg/schemapatcher/zz_generated.markerhelp.go
```diff
@@ -40,6 +40,10 @@ func (Generator) Help() *markers.DefinitionHelp {
 				Summary: "specifies the maximum description length for fields in CRD's OpenAPI schema. ",
 				Details: "0 indicates drop the description for all fields completely. n indicates limit the description to at most n characters and truncate the description to closest sentence boundary if it exceeds n characters.",
 			},
+			"GenerateEmbeddedObjectMeta": {
```
Porges, line 43: This change appeared: it should probably already have been checked in.

Porges:
@alvaroaleman please take a look, this is a follow-on to the [last PR](https://github.com/kubernetes-sigs/controller-tools/pull/518)

alvaroaleman (approved):

alvaroaleman:
/lgtm cancel
floats are generally not allowed, ref https://github.com/kubernetes-sigs/controller-tools/issues/245

Porges:
> @alvaroaleman  floats are generally not allowed, ref #245

Corrected golden file test.

alvaroaleman:
@porges Can you fix the compile error please?

Porges:
@alvaroaleman looks like a bad linter, have updated code to fix.

PR #618, opened 2021-09-18
:bug: Be more helpful with floats when allowDangerousTypes=false

alvaroaleman:
Currently, we fall through and just emit an error telling that floats
are not supported which lead to ppl creating prs to add support. We
actually do support them although that is discouraged. This change makes
us emit a more helpful error.

Closes https://github.com/kubernetes-sigs/controller-tools/pull/566
Closes https://github.com/kubernetes-sigs/controller-tools/issues/245

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -17,6 +17,7 @@ limitations under the License.
 package crd
 
 import (
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


camilamacedo86 (commented):
/lgtm

That makes sense for me. 🥇

PR #619, opened 2021-09-19
🐛 Allow floating-point values in validations

### LINKED ISSUE #543: Allow floating point values for validation:{Minimum|Maximum|MultipleOf}
The Kubebuilder CRD validation markers for Minimum, Maximum and MultipleOf currently only takes integral values. If one tries to assign a floating point value, e.g. `// +kubebuilder:validation:Minimum=0.5` generation fails with a message like this `extra arguments provided: ".5"`. If the decorated property is a float then it makes sense to have floating point bounds and this is also supported by k8s, see [here](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/types_jsonschema.go#L34-L40)

Porges:
Closes #543. We have also run into this while working on https://github.com/Azure/azure-service-operator.

- Extend marker parser to allow parsing float/JSON `number` values.
- Allow validations for `Minimum`, `Maximum`, and `MultipleOf` to have float values when applied to float fields.

Actual use of float fields is still restricted by the `allowDangerousTypes` flag, so no further restrictions are needed as part of this change.


Porges:
/remove-lifecycle rotten 

bad bot

pkg/markers/parse.go
```diff
@@ -84,6 +84,8 @@ const (
 	InvalidType ArgumentType = iota
 	// IntType is an int
 	IntType
+	// NumberType is a float64
+	NumberType
```
erikgb, line 88: Wouldn't the same `FloatType` make more sense here? Or is there are reason for this?

erikgb (commented):
Did a quick scan. Looks reasonable to me, but I am not too familiar with this codebase.

/ok-to-test

pkg/markers/parse.go
Porges: These names reflect the schema-level type names and not the go-level names; in the CRD output this will be "type: number".

AlmogBaku: That's what the json schema calls it

pkg/crd/parser_integration_test.go
```diff
@@ -75,6 +75,7 @@ var _ = Describe("CRD Generation From Parsing to CustomResourceDefinition", func
 			Collector:              &markers.Collector{Registry: reg},
 			Checker:                &loader.TypeChecker{},
 			IgnoreUnexportedFields: true,
+			AllowDangerousTypes:    true, // need to allow “dangerous types” in this file for testing
```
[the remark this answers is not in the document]
erikgb: https://book.kubebuilder.io/reference/controller-gen.html#generators

Using float is not recommended in the K8s API, in particular not in `spec`. Ref. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

pkg/markers/parse.go
erikgb: Thanks for the explanation.

pkg/crd/parser_integration_test.go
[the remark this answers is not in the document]
AlmogBaku: fair enough

AlmogBaku (approved):

pkg/markers/parse.go
```diff
@@ -346,8 +354,11 @@ func guessType(scanner *sc.Scanner, raw string, allowSlice bool) *Argument {
 		if nextTok == '-' {
 			nextTok = subScanner.Scan()
 		}
+
 		if nextTok == sc.Int {
 			return &Argument{Type: IntType}
+		} else if nextTok == sc.Float {
+			return &Argument{Type: NumberType}
 		}
```
joelanford, line 362: Nit:
```suggestion
		}
		if nextTok == sc.Float {
			return &Argument{Type: NumberType}
		}
```

pkg/markers/parse.go
```diff
@@ -755,7 +794,7 @@ func (d *Definition) loadFields() error {
 func parserScanner(raw string, err func(*sc.Scanner, string)) *sc.Scanner {
 	scanner := &sc.Scanner{}
 	scanner.Init(bytes.NewBufferString(raw))
-	scanner.Mode = sc.ScanIdents | sc.ScanInts | sc.ScanStrings | sc.ScanRawStrings | sc.SkipComments
+	scanner.Mode = sc.ScanIdents | sc.ScanInts | sc.ScanStrings | sc.ScanRawStrings | sc.SkipComments | sc.ScanFloats
```
joelanford, line 797: Nit:
```suggestion
	scanner.Mode = sc.ScanIdents | sc.ScanInts | sc.ScanFloats | sc.ScanStrings | sc.ScanRawStrings | sc.SkipComments
```

joelanford:
For some nits:
/hold

But its nothing to block merging over, so:
/approve

If you want to accept the nits, ping me again and I can re-add lgtm and approve labels.

Porges:
> Hi, Thanks for your contribution!
> 
> Overall it looks good to me, I'm just wondering about this "AllowDangerousTypes" that you also added in the marker

It’s required in order to use float values at all in the test file.

Porges:
Thanks for reviewing @joelanford

PR #620, opened 2021-09-20
🐛 Further deduplication of RBAC rules

### LINKED ISSUE #612: Deduplicate RBAC rules
The deduplication of RBAC rules currently only merges verbs for tags with the same `groups` / `resources` / `resourceNames` / `urls`. It would be nice if we could generate rules without any duplicate permissions.

For example, the following tags:

```golang
// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinedeployments;machinedeployments/status;machinedeployments/finalizers,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinedeployments;machinedeployments/finalizers,verbs=get;list;watch;update;patch
// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinedeployments,verbs=get;list;watch
// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinedeployments,verbs=get;list;watch;create;update;patch;delete
```

Result in:

```yaml
- apiGroups:
  - cluster.x-k8s.io
  resources:
  - machinedeployments
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - cluster.x-k8s.io
  resources:
  - machinedeployments
  - machinedeployments/finalizers
  verbs:
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - cluster.x-k8s.io
  resources:
  - machinedeployments
  - machinedeployments/finalizers
  - machinedeployments/status
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
```

The last 2 have been merged because the resource is equal, the others are not.

Note: This can also be solved by splitting the tags to have a separate tag for each resource.

Porges:
Closes #612.

Check if rules are rendered irrelevant by other rules and remove them if so.


pkg/rbac/parser_test.go
```diff
@@ -0,0 +1,151 @@
...
+	g.Expect(nrules[1].Subsumes(nrules[2])).To(gomega.BeTrue())
+	g.Expect(nrules[1].Subsumes(nrules[3])).To(gomega.BeTrue())
+
+	// distinct
+	g.Expect(nrules[3].Subsumes(nrules[2])).To(gomega.BeFalse())
+	g.Expect(nrules[2].Subsumes(nrules[3])).To(gomega.BeFalse())
+}
+
+func Test_SubsumesIsReflexive(t *testing.T) {
+	g := gomega.NewWithT(t)
```
Porges, line 78: use some kind of property testing for this and following test?

pkg/rbac/testdata/controller.go
```diff
@@ -10,4 +12,9 @@ package controller
 // +kubebuilder:rbac:groups=cron;batch,resources=jobs/status,verbs=get;create
 // +kubebuilder:rbac:groups=batch,resources=jobs/status,verbs=watch;watch
 // +kubebuilder:rbac:groups=art,resources=jobs,verbs=get,namespace=park
-// +kubebuilder:rbac:groups=batch.io,resources=cronjobs,resourceNames=foo;bar;baz,verbs=get;watch
+// +kubebuilder:rbac:groups=batch.io,resources=cronjobs2,resourceNames=foo;bar;baz,verbs=get;watch
```
Porges, line 15: renamed to `cronjobs2` or else it is subsumed by the `kubebuilder:rbac:groups=batch.io,resources=cronjobs,verbs=get;watch;create` line.

pkg/rbac/parser.go
```diff
@@ -63,83 +60,70 @@ type Rule struct {
...
-// new slice of strings.
-func removeDupAndSort(strs []string) []string {
-	set := make(map[string]bool)
-	for _, str := range strs {
-		if _, ok := set[str]; !ok {
-			set[str] = true
-		}
-	}
+func (nr *NormalizedRule) GenerateSortKey() {
+	nr.SortKey = strings.Join(
```
Porges, line 96: This should probably have a better name as it is correctness-critical?

Porges:
/remove-lifecycle rotten

it's not rotten, it's floundering

pkg/rbac/parser_test.go
sbueringer: What do you mean with property testing?

sbueringer (commented):
Not a maintainer, but I try to help out :)

A few suggestions

pkg/rbac/stringset.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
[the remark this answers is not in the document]
Porges: Done.

pkg/rbac/parser.go
```diff
@@ -253,6 +192,90 @@ func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{
...
+			// not needed; another rule handles this case
+			return dest
+		}
+
+		if it.Subsumes(other) {
+			// rebuild whole list
+			result := []*NormalizedRule{it}
+			for _, d := range dest {
+				result = insertRule(result, d)
+			}
```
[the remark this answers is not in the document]
Porges: It could, but since `it.Subsumes(other)` and we start the new list with `it` being contained, then `other` will be omitted by the previous check above. Explicitly skipping it makes the code a bit messier.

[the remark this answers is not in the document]
alvaroaleman: What you are doing here is both harder to read and less eficient than:
```
func remove(s []int, i int) []int {
    s[i] = s[len(s)-1]
    return s[:len(s)-1]
}
return append(remove(dest, idx), it)
```

pkg/rbac/parser.go
```diff
@@ -63,83 +61,93 @@ type Rule struct {
...
 
-// addVerbs adds new verbs into a Rule.
-// The duplicates in `r.Verbs` will be removed, and then `r.Verbs` will be sorted.
-func (r *Rule) addVerbs(verbs []string) {
-	r.Verbs = removeDupAndSort(append(r.Verbs, verbs...))
+type NormalizedRule struct {
+	// if two different rules have the same comparison key then they can have their verbs merged
+	// this key should comprise all the fields below except Verbs and Namespace (since rules
+	// are partitioned by namespace before merging)
+	ComparisonKey string
```
alvaroaleman, line 97: Sorry, I don't understand why this is needed. If one rule is a superset of another rule, the subet rule can be removed. Why is any additional merging needed?

alvaroaleman (commented):
Hi @Porges 

Thanks for submitting this change. There are a couple of issues with it IMHO, most notably that is too invasive and not properly tested, with the result that it is incorrect.


From what I can tell by looking at https://github.com/kubernetes-sigs/controller-tools/issues/612, all of it can be solved by by removing rules that have a supset. As a result, I would expect the change to the existing code to look something like this:

```diff
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -194,6 +194,8 @@ func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{
                        ruleMap[key].addVerbs(rule.Verbs)
                }
 
+               ruleMap = removeRulesThatHaveSuperset(ruleMap)
+
                // sort the Rules in rules according to their ruleKeys
                keys := make([]ruleKey, 0, len(ruleMap))
                for key := range ruleMap {
```

As a general rule of thumb for making any changes:
* Keep the change as small as possible
* Do not make any unrelated stylistic changes
* Keep the change scoped, do not mix refactorings (change of existing code that does not intend to change behavior) with behavioural changes

pkg/rbac/parser.go
```diff
@@ -63,83 +61,93 @@ type Rule struct {
...
-	var result []string
-	for str := range set {
-		result = append(result, str)
-	}
+	result := set.UnsortedList()
 	sort.Strings(result)
 	return result
 }
 
+// Subsumes indicates if one rule entirely determines another,
```
alvaroaleman, line 130: This might be due to the fact that I am not a native English speaker, but I find a name like `IsSuperset` way more obvious.

pkg/rbac/parser.go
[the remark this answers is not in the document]
Porges: @alvaroaleman that doesn’t result in the quite same outcome; if Rule X is superset of Rule Y in the list, it could also be a superset of Rule Z (and others), so we still need to go through and check the remainder of the list.

pkg/rbac/parser.go
```diff
@@ -63,83 +61,93 @@ type Rule struct {
...
+// meaning that the other is unnecessary.
+// Remember that Kubernetes RBAC rules are purely additive, there
+// are no deny rules.
+func (nr *NormalizedRule) Subsumes(other *NormalizedRule) bool {
```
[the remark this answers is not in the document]
Porges: Have updated this versus the documentation on the `PolicyRule` struct. Thanks, this is the kind of feedback I wanted on the correctness. :+1:

pkg/rbac/parser.go
Porges: This is to handle verb list merging (a part of the existing codebase as well); if two rules match exactly the same things then they can be unified and their verb lists merged.

pkg/rbac/parser.go
```diff
@@ -23,22 +23,20 @@ limitations under the License.
...
 )
 
-var (
-	// RuleDefinition is a marker for defining RBAC rules.
-	// Call ToRule on the value to get a Kubernetes RBAC policy rule.
-	RuleDefinition = markers.Must(markers.MakeDefinition("kubebuilder:rbac", markers.DescribesPackage, Rule{}))
-)
+// RuleDefinition is a marker for defining RBAC rules.
+// Call ToRule on the value to get a Kubernetes RBAC policy rule.
+var RuleDefinition = markers.Must(markers.MakeDefinition("kubebuilder:rbac", markers.DescribesPackage, Rule{}))
```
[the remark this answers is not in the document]
Porges: Unfortunately this was auto-formatter collateral damage.

pkg/rbac/parser_test.go
Porges: See updated tests; in this case using the `testing/quick` package. We generate arbitrary `Rule`s and then make assertions about them.

pkg/rbac/parser.go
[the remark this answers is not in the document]
Porges: I have been building out the unit tests (given that there were none to start with) but definitely could do with more.

Porges:
Ideally we'd also rely on [the k8s code](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/apis/rbac/v1/helpers.go) to help enforce requirements on the `PolicyRule` but importing this requires [a bunch of `replace`s added into `go.mod`](https://github.com/kubernetes-sigs/cloud-provider-huaweicloud/blob/master/go.mod). Is this viable/okay to proceed with?

alvaroaleman:
> he k8s code to help enforce requirements on the PolicyRule (see PolicyRuleBuilder) but importing this requires a bunch of replaces added into go.mod. Is this viable/okay to proceed with?

Unfortunately not, because that will break `go install`.

In general, k8s itself states you must never import k8s/kubernetes (I can't find the reference for that off-hand, but it exists). The only way to get us to use upstream code would be to talk to upstream sig-apimachinery if they are open to move this to staging (i.E. below the `staging` directory in the repo). Then it gets published to repos tthat can be imported without `replacfe`. On a quick look, it seems to be that this might be possible for this package, as it only seems to use external api. But I might be missing context.

PR #626, opened 2021-09-29
🐛 Updating WriteYAML to not add empty line to start files

robscott:
Currently generated YAML files all have  `\n---\n` prepended to them. Although that's not an issue when using `kubectl apply` with those files, it requires extra work when trying to do something simple like parse CRD yaml from a controller. Although not a significant issue, it's caused us ([Gateway API](https://gateway-api.sigs.k8s.io/) + associated controller authors) to have to workaround it in several places. When parsing YAML files like this, the leading new line is parsed as an empty yaml struct, so we need to write code that can handle this empty/invalid object that starts all our generated YAML CRDs.

It looks like this code originated from https://github.com/kubernetes-sigs/controller-tools/pull/208, and this particular part of the change does not seem like it was intentional.

Initially I wrote this to simply not add the delimiter to the start of the file, but as @howardjohn pointed out, that would make the concatenated files invalid YAML. This change does mean that concatenating YAML files that don't have a trailing new line will not work. Of course all files generated by this tool do have a trailing new line, and that seems like a reasonable expectation of other YAML files.

### DIFF
--- a/pkg/crd/testdata/gen/bar.example.com_foos.v1beta1.yaml
+++ b/pkg/crd/testdata/gen/bar.example.com_foos.v1beta1.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
--- a/pkg/crd/testdata/gen/bar.example.com_foos.yaml
+++ b/pkg/crd/testdata/gen/bar.example.com_foos.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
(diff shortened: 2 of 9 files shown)


howardjohn (commented):
hypothetically it means `cat *.yaml` is no longer valid yaml doc. I think its a good change, just one hypothetical reason it *could* be useful

robscott:
@howardjohn that's a good point, I was on the fence with the approach here but ended up deciding to go back to a delimiter to start the file, just without the extra prepended new line, hopefully that's a reasonable balance here.

alvaroaleman (approved):
/lgtm
/approve

thanks!

PR #627, opened 2021-09-29
:bug: fix: Don't flatten schemas for type idents we don't know about

### LINKED ISSUE #624: panic: runtime error: invalid memory address or nil pointer dereference
Using controller-ben v0.6.2 I get a panic error when trying to generate the crd for https://github.com/Orange-OpenSource/casskop/tree/fix-crds/multi-casskop

```
controller-gen --version
Version: v0.6.2
make controller-gen
make[1]: Entering directory '/home/circleci/*******/multi-*******'
Makefile:102: warning: overriding recipe for target 'build'
../shared.mk:127: warning: ignoring old recipe for target 'build'
make[1]: Nothing to be done for 'controller-gen'.
make[1]: Leaving directory '/home/circleci/*******/multi-*******'
/go/bin/controller-gen crd paths=./pkg/apis/... output:dir=./deploy/crds schemapatch:manifests=./deploy/crds
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xa15efb]

goroutine 1 [running]:
sigs.k8s.io/controller-tools/pkg/crd.schemaWalker.walkSchema({{0xcdc240, 0xc004cac1e0}}, 0x0)
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/pkg/crd/schema_visitor.go:63 +0xbb
sigs.k8s.io/controller-tools/pkg/crd.EditSchema(...)
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/pkg/crd/schema_visitor.go:42
sigs.k8s.io/controller-tools/pkg/crd.FlattenEmbedded(0xc0021048d0, {0xcdc280, 0xc0000ede00})
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/pkg/crd/flatten.go:210 +0x7b
sigs.k8s.io/controller-tools/pkg/crd.(*Parser).NeedFlattenedSchemaFor(0xc003649ce0, {0xc0000ede00, {0xc0004fa070, 0x2}})
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/pkg/crd/parser.go:204 +0xe8
sigs.k8s.io/controller-tools/pkg/schemapatcher.Generator.Generate({{0xc0001c293a, 0xb43d80}, 0x0, 0x0}, 0xc001c2f7c0)
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/pkg/schemapatcher/gen.go:143 +0x9e9
sigs.k8s.io/controller-tools/pkg/genall.(*Runtime).Run(0xc00024ca00)
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/pkg/genall/genall.go:207 +0x206
main.main.func1(0xc0002b8500, {0xc00028dc00, 0x4, 0x4})
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/cmd/controller-gen/main.go:176 +0x6a
github.com/spf13/cobra.(*Command).execute(0xc0002b8500, {0xc00019e010, 0x4, 0x4})
	/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:856 +0x60e
github.com/spf13/cobra.(*Command).ExecuteC(0xc0002b8500)
	/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:974 +0x3bc
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:902
main.main()
	/go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/cmd/controller-gen/main.go:200 +0x2c6
make: *** [../shared.mk:122: generate] Error 2
```

I'm trying to replace `operator-sdk generate k8s` by using `controller-gen` but hit this error.

abayer:
I hit this problem with cases where there are multiple packages with the same group and version. The schema patcher would hit a nil pointer trying to flatten a schema for a kind that exists in one package, but not another, because it seems to assume that everything's in one package. By skipping cases where the parser hasn't already found the given kind in the given package, we got past that panic and everything works correctly.

This adds an option for `schemapatch`, `allowMultiPackageGroup` - when that is true, we won't error out in scenarios where a group is used in multiple packages, while in the default case, we'll error out with the package path and kind that couldn't be found.

I believe this fixes #624, though I can't be sure if this is the only path that can lead to that panic.

Signed-off-by: Andrew Bayer <andrew.bayer@gmail.com>


alvaroaleman:
So you have multiple go packages that all have types for the same api group and version? Why?

alvaroaleman:
> @alvaroaleman Honestly? The best explanation is probably that we screwed up. =)

So what keeps you from reorganizing them?

I am fine better handling this case and emit a proper error rather than panicking. But I am not sure we want to actually support it.

abayer:
That's a good question - Tekton's pretty established at this point, which obviously would make any significant reorganization more difficult. I'll raise the topic, but given that this is the first time I'm aware of us hitting problems due to our quirky structure, and we're only hitting the problem now that we're trying to add auto-generated OpenAPI schemas to our previously-handcrafted CRDs, I'm not sure if there'll be much, if any, willingness to shift things around.

abayer:
Ok, so https://github.com/tektoncd/pipeline/tree/main/pkg/apis has three relevant subdirectories - `pipeline`, `run`, and `resource`. `run` and `resource` just have `v1alpha1` subdirs/versions, while `pipeline` has both `v1alpha1` and `v1beta1`. I'm trying to see if `pkg/apis/resource/v1alpha1`'s contents can be moved to `pkg/apis/pipeline/v1alpha1`, but due to how we have `pipeline`'s `v1alpha1` and `v1beta1` set up, with many types in `v1alpha1` just being aliases for what's in `v1beta1`, and both `pipeline/v1alpha1` and `pipeline/v1beta1` having dependencies on `resource/v1alpha1`, everything gets really messy trying to move it around...

So yeah, we definitely screwed up - since Tekton came out of Knative, I assume our original `pkg/apis/...` layout was based on how it was done in Knative projects - i.e., https://github.com/knative/serving/tree/main/pkg/apis has `autoscaling` and `serving` (and around the time Tekton started as Knative Build, [it also had `networking`](https://github.com/knative/serving/tree/v0.2.0/pkg/apis). And then when we added `.../pipeline/v1beta1`, we probably didn't do _that_ quite right either (the type aliasing to `v1beta1` from `v1alpha1` like [this](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1alpha1/pipeline_types.go#L217) doesn't feel right to me, but I can't claim to be anything like an expert!), just making things more confusing, while we left `run` and `resource` at `v1alpha1` with another bunch of type aliasing into `pipeline/v1beta1` (and `pipeline/v1alpha1`, for that matter). It's definitely a mess.

I'll continue to see if we can reorganize, and definitely understand not wanting to support a weird edge case of a layout like Tekton's, but it'd be really handy for us (and possibly for others) if we could at least have an option to enable the behavior I add in this PR (as well as default behavior that errors out nicely rather than panicking).

abayer:
I've updated the PR to add an `allowMultiPackageGroup` option, defaulting to erroring out.

vdemeester:
We did not screwed up per se, it was done on purpose with only go type (and aliasing stuff in mind), to ease the migration of the whole code base from v1alpha1 to v1beta1. The assumption we made (or at least I made) was that nothing would/should prevent us from having multiple packages presenting one api version. I would still make that assumption but I also understand that it can be trickier to support.

The real trick in our case is the deprecated, in alpha, resources that we didn’t want to move to the v1beta1 package (and api version) - but still allowing our user to refer to these resources from the v1beta1 for the time being.

The plan is to get rid of v1alpha1 at some point. I also see no real problems to remove type aliasing, etc.. and having some duplication to allow us to fix this (sometime duplication is better than the wrong abstraction).

pkg/schemapatcher/gen.go
```diff
@@ -75,6 +75,10 @@ type Generator struct {
 
 	// GenerateEmbeddedObjectMeta specifies if any embedded ObjectMeta in the CRD should be generated
 	GenerateEmbeddedObjectMeta *bool `marker:",optional"`
+
+	// AllowMultiPackageGroups specifies whether cases where a group is used for multiple packages should be allowed,
+	// rather than the default behavior of failing.
```
camilamacedo86, line 80: After adding this PR thedefault behaviour would not be a failing right? 
Could you please fix the comment to allow it describes what the variable represents and when it will be used such as we have for the other ones?

pkg/schemapatcher/gen_integration_test.go
```diff
@@ -41,7 +41,7 @@ var _ = Describe("CRD Patching From Parsing to Editing", func() {
 		var crdSchemaGen genall.Generator = &Generator{
 			ManifestsPath: "./invalid",
 		}
-		rt, err := genall.Generators{&crdSchemaGen}.ForRoots("./...")
+		rt, err := genall.Generators{&crdSchemaGen}.ForRoots("./apis/kubebuilder/...", "./apis/legacy/...")
```
camilamacedo86, line 44: Why did you need to change it for the specific pkgs?

pkg/schemapatcher/gen_integration_test.go
```diff
@@ -67,7 +67,7 @@ var _ = Describe("CRD Patching From Parsing to Editing", func() {
 		var crdSchemaGen genall.Generator = &Generator{
 			ManifestsPath: "./valid",
 		}
-		rt, err := genall.Generators{&crdSchemaGen}.ForRoots("./...")
+		rt, err := genall.Generators{&crdSchemaGen}.ForRoots("./apis/kubebuilder/...", "./apis/legacy/...")
```
camilamacedo86, line 70: same. I think it needs to be reverted to ensure that the changes made here do not break any current test. Could you please clarify why this change was required?

camilamacedo86 (commented):
https://github.com/kubernetes-sigs/controller-tools/pull/627#issuecomment-931355857
+1 
Hi @abayer, @vdemeester,

IHMO it is missing ( probably in the issue )
- An example of the scenario
- A description over when it is required 

Also, we would need to check if it is a valid scenario or if this PR is trying to add support to a scenario that ought to not be done. 

Are you trying here to allow in the same project you specific the same GVK? 
If you deprecated the version then,  ought you not have a different GVK since the version would not be the same? 

if I understood properly your scenario then I think the controller-gen out to raise an error saying that the user is trying to specify more than one time the same GVK instead to add support to allow it.

abayer:
@camilamacedo86 Sorry for the delay! I missed the notifications originally and just remembered to check this PR. 🤦 Thanks for your comments, and I'll address them today.

pkg/schemapatcher/gen.go
abayer: The default behavior would still be failing in the multiple-packages-for-a-single-group scenario. The new option allows you to change that behavior, but the default stays the same.

pkg/schemapatcher/gen_integration_test.go
abayer: This is because I made the (probably wrong!) decision to reuse the contents of `pkg/schemapatcher/testdata` for both the existing tests and the new testing of multiple-packages-per-group. I'll separate the test data for those situations.

abayer:
@camilamacedo86 So the situation in Tekton Pipeline is that we have `pkg/apis/pipeline/v1alpha1`, `pkg/apis/pipeline/v1beta1`, `pkg/apis/resource/v1alpha1`, and `pkg/apis/run/v1alpha1`. They all have `tekton.dev` as their group. See above for some details on why we have things that way. `schema-gen` fails because it doesn't expect to see the same group and version for different packages. It fails with the same panic that's in #624, but I don't know if that's actually caused by the same thing or not. I should make a new issue, and will do that.

PR #628, opened 2021-09-30
✨ Adds a tool for linting OpenAPI schemas on CRD resources

munnerz:
This PR adds a new CLI tool named `crd-linter` which will lint/run checks against CRD resources and their corresponding schema.

Initially this includes just 4 linters to check for:

* SchemaProvided: ensures a schema is provided for all API versions
* PreserveUnknownFields: ensures that `provideUnknownFields` and `x-kubernetes-preserve-unknown-fields` is not used
* MaxLengthStrings: ensures `string` typed fields have a `maxLength` specified
* MaxItemsArrays: ensures `array` typed fields have a `maxItems` specified

This is all in the effort to ensure a higher quality of CRDs, which helps protect apiservers against e.g. too much data being stored in a CRD, or ever-growing number of items in arrays that can cause scalability issues.

These linters aren't aimed at all being "best practice", but providing this tool allows teams to check for potential issues/better understand APIs they are installing and where issues may be found.

I'm not sure where this is best to live, nor am I wedded to any parts of the design of it, so welcome any and all feedback 😄

ref https://kubernetes.slack.com/archives/C0EG7JC6T/p1628592082067800

### DIFF
--- a/cmd/crd-linter/README.md
+++ b/cmd/crd-linter/README.md
@@ -1,0 +1,51 @@
+# Linter for Kubernetes CustomResourceDefinition resources
+
+The crd-linter encompasses a number of linters that can be run against
+Kubernetes CustomResourceDefinition (CRD) objects to help ensure best-practices
+are followed when CRDs are authored.
+
+## Usage
+
+```
+$ crd-linter --help
+Usage of crd-linter:
+      --exceptions-file string   Path to a list of exceptions for linter failures
+      --linters strings          Full list of linters to run against discovered CustomResourceDefinitions (default [SchemaProvided,PreserveUnknownFields,MaxLengthStrings,MaxItemsArrays])
+      --output-exceptions        If true, an exception list file will be written to the file denoted by '--exceptions-file'
+      --path string              Path to recursively search for CustomResourceDefinition objects
+```
+
+The crd-linter must be pointed at a directory containing YAML files with CRD
+object(s) defined in them using the `--path` flag.
+
+For example:
+
+```
+crd-linter --path ./path/to/crds/
+```
+
+### Linters
+
+The following linters are supported:
+
+| Name                  | Description                                                         |
+|-----------------------|---------------------------------------------------------------------|
+| MaxLengthStrings      | Ensures that all 'string' type fields have a `maxLength` option set |
+| MaxItemsArrays        | Ensures that all 'array' type fields have a `maxItems` option set   |
+| SchemaProvided        | Ensures that all versions of all CRDs provide an OpenAPI schema     |
+| PreserveUnknownFields | Ensures that no fields within a CRD schema permit unknown fields    |
+
+### Exceptions
+
+An exceptions file may be provided which will permit certain rule violations to
+be skipped. This is useful when gradually improving/fixing issues within a
+repository of CRDs, whilst not permitting new violations.
+
+You can generate an exceptions file by setting `--output-exceptions=true`
+whilst also providing a `--exceptions-file`.
+This will output all linter failures to the named exceptions file, so that
+future runs of the linter will skip these failures.
+
+Use this in combination with the `--linters` flag to generate an exceptions
+file that only contains a subset of linter failures (e.g. only the nosiest or
+least problematic failures).
--- a/cmd/crd-linter/exceptions/exceptions.go
+++ b/cmd/crd-linter/exceptions/exceptions.go
@@ -74,10 +74,11 @@ func (l *ExceptionList) Size() int {
 }
 
 func (l *ExceptionList) WriteToFile(path string) error {
+	// #nosec G306
 	return ioutil.WriteFile(path, []byte(l.String()), 0644)
 }
 
-func New() *ExceptionList {
+func NewExceptionList() *ExceptionList {
 	return &ExceptionList{Exceptions: make([]Exception, 0)}
 }
 
@@ -89,7 +90,7 @@ func LoadFromFile(path string) (*ExceptionList, error) {
 		return nil, err
 	}
 
-	list := New()
+	list := NewExceptionList()
 	entries := strings.Split(string(data), "\n")
 	for lineNo, e := range entries {
 		// skip empty lines
--- a/cmd/crd-linter/exceptions/exceptions.go
+++ b/cmd/crd-linter/exceptions/exceptions.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package exceptions
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+)
+
+type ExceptionList struct {
+	Exceptions []Exception
+}
+
+type Exception struct {
+	filename, crdName, linter, fieldRef string
+}
+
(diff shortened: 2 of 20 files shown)


munnerz:
Just one failure left, and I'm not sure if it makes sense to change the code here:

```
cmd/crd-linter/exceptions/exceptions.go:77:9: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	return ioutil.WriteFile(path, []byte(l.String()), 0644)
```

erikgb:
> Just one failure left, and I'm not sure if it makes sense to change the code here:
> 
> 
> 
> ```
> 
> cmd/crd-linter/exceptions/exceptions.go:77:9: G306: Expect WriteFile permissions to be 0600 or less (gosec)
> 
> 	return ioutil.WriteFile(path, []byte(l.String()), 0644)
> 
> ```

IMO this should be considered a false positive and suppressed with  an annotation.

cmd/crd-linter/exceptions/exceptions.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (l *ExceptionList) WriteToFile(path string) error {
+	return ioutil.WriteFile(path, []byte(l.String()), 0644)
+}
+
+func New() *ExceptionList {
+	return &ExceptionList{Exceptions: make([]Exception, 0)}
+}
+
+// LoadFromFile will load a pipe (|) separated list of Exceptions from the
```
erikgb, line 84: This doc appears a bit confusing to me. The exception list seems to be separated by newline; one exception per line. But one exception seems to be serialized in a pipe-separated format.

Have you considered other file formats? The suggested format is very compact, but the semantics are not apparent when looking at a file.

cmd/crd-linter/linters/interface.go
```diff
@@ -0,0 +1,32 @@
...
+
+import v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+
+// A Linter evaluates a CRD resource and provides information about policy violations.
+type Linter interface {
+	// Name returns the name of this linter, used during report output
+	Name() string
+
+	// Execute runs the linter against the given custom resource definition
+	Execute(*v1.CustomResourceDefinition) []string
```
erikgb, line 27: Have you considered a more "structured" return type for this function? When implementing new future linters, I would prefer a "tighter" API than just `[]string`. To understand a lint violation, I think the user at least need some reference to the violation inside the CRD-file. Path and/or line number?

cmd/crd-linter/exceptions/exceptions.go
munnerz: 👍 the format here wasn't so much chosen to be compact, but rather just to have each entry on a single line to make it clear to see how many violations there are. An example line:

```
path/to/the/file/jaegertracing.io_jaegers_crd.yaml|jaegers.jaegertracing.io|MaxLengthStrings|spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.collector.properties.volumes.items.properties.flexVolume.properties.fsType.maxLength is not specified on a field of type 'string'
```

Happy to adjust this format however you feel makes it clearer - what sort of format do you think would be better/what do you have in mind?

cmd/crd-linter/linters/interface.go
munnerz: Done ✅

cmd/crd-linter/exceptions/exceptions.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (l *ExceptionList) Size() int {
+	return len(l.Exceptions)
+}
+
+func (l *ExceptionList) WriteToFile(path string) error {
+	return ioutil.WriteFile(path, []byte(l.String()), 0644)
+}
+
+func New() *ExceptionList {
```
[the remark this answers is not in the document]
munnerz: Done ✅

cmd/crd-linter/exceptions/exceptions.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return buf.String()
+}
+
+func (l *ExceptionList) Size() int {
+	return len(l.Exceptions)
+}
+
+func (l *ExceptionList) WriteToFile(path string) error {
+	return ioutil.WriteFile(path, []byte(l.String()), 0644)
```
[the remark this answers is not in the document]
munnerz: Done ✅

cmd/crd-linter/linters/preserve_unknown_fields.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// PreserveUnknownFields checks if a CRD contains a 'preserveUnknownFields' marker
+// either at the top level, or on any specific sub-field of the resource.
+type PreserveUnknownFields struct{}
```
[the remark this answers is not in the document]
munnerz: Done ✅

munnerz:
I've addressed all the feedback - there's just the one open question on the serialisation format for the exceptions file left :)

cmd/crd-linter/exceptions/exceptions.go
erikgb: @munnerz Sorry for the delay. You have reasonable arguments for selecting the chosen file format, and I think it should be ok. Only downside I see, is that it is not very human readable. The format must remain stable over time, so we should decide if human readability is important now. I'll guess an IDE could be a consumer of these files. Do we have any reference or common practice to refer to?

cmd/crd-linter/README.md
```diff
@@ -0,0 +1,51 @@
...
+### Linters
+
+The following linters are supported:
+
+| Name                  | Description                                                         |
+|-----------------------|---------------------------------------------------------------------|
+| MaxLengthStrings      | Ensures that all 'string' type fields have a `maxLength` option set |
+| MaxItemsArrays        | Ensures that all 'array' type fields have a `maxItems` option set   |
+| SchemaProvided        | Ensures that all versions of all CRDs provide an OpenAPI schema     |
+| PreserveUnknownFields | Ensures that no fields within a CRD schema permit unknown fields    |
```
camilamacedo86, line 36: Is not this attr only valid for v1beta1 CRDs? This project no longer supports or work with v1beta1

cmd/crd-linter/README.md
```diff
@@ -0,0 +1,51 @@
...
+      --path string              Path to recursively search for CustomResourceDefinition objects
+```
+
+The crd-linter must be pointed at a directory containing YAML files with CRD
+object(s) defined in them using the `--path` flag.
+
+For example:
+
+```
+crd-linter --path ./path/to/crds/
```
camilamacedo86, line 24: This project is responsible for gen resources? 
could you add in the first comment why do you think that this project ought to also linter the resources? I think we need here to know when and why it would be required and used? 

Also, would that be a new bin or it would also be hipped with controller-gen? Have we an enhancement proposal for it? and then, why add support to lint only the CRD and not all manifests generated by the project?

camilamacedo86 (commented):
Have we an enhancement proposal for this one?

cmd/crd-linter/linters/preserve_unknown_fields.go
```diff
@@ -0,0 +1,52 @@
...
+	if crd.Spec.PreserveUnknownFields {
+		errs = append(errs, Warning{Message: "spec.preserveUnknownFields is set to 'true'"})
+	}
+	return append(errs, recurseAllSchemas(crd.Spec.Versions, func(props v1.JSONSchemaProps, path string) []Warning {
+		if props.XPreserveUnknownFields != nil && *props.XPreserveUnknownFields == true {
+			return newWarningList(fmt.Sprintf("%s.x-kubernetes-preserve-unknown-fields is set to 'true'", path))
+		}
+		return nil
+	})...)
+}
```
roee88, line 52: The description mentions that v1 forbids it, but that seems to be true only to spec.preserveUnknownFields and not to x-kubernetes-preserve-unknown-fields on specific properties. Am I missing something?

erikgb:
/remove-lifecycle stale

I would like to use this tool! 😊

PR #630, opened 2021-10-17
:bug: CRD generation: remove status before writing

justinsb:
It's tricky to get the existing CRD types to not include the status,
so we instead allow for filtering of the map just before writing.

Issue #456

### DIFF
--- a/pkg/genall/genall.go
+++ b/pkg/genall/genall.go
@@ -17,13 +17,14 @@ limitations under the License.
 package genall
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 
 	"golang.org/x/tools/go/packages"
-	"sigs.k8s.io/yaml"
+	rawyaml "gopkg.in/yaml.v2"
 
 	"sigs.k8s.io/controller-tools/pkg/loader"
 	"sigs.k8s.io/controller-tools/pkg/markers"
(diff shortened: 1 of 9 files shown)


justinsb:
It's odd to include status for two reasons:

1) We don't control the status; we control the spec.
2) This seems to cause server-side apply to repeatedly reapply (on some k8s versions?)

pkg/genall/genall.go
```diff
@@ -147,6 +160,41 @@ func (g GenerationContext) WriteYAML(itemPath string, objs ...interface{}) error
 	return nil
 }
 
+// MarshalToYAML is based on sigs.k8s.io/yaml.Marshal, but allows for transforming the final data before writing.
```
yuwenma, line 163: typo?

yuwenma:
This PR can greatly highlight and reemphasize the definition of [object spec and status](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#object-spec-and-status)
and avoids the potential misuse of status which should be a [control plane](https://kubernetes.io/docs/reference/glossary/?all=true#term-control-plane) task to "continually and actively manages every object's actual state".

mengqiy (approved):
/lgtm
/approve

pkg/genall/genall.go
justinsb: Well spotted - thanks!

justinsb:
Fixed comment - thanks for reviews @yuwenma and @mengqiy

/hold cancel

PR #635, opened 2021-11-07
✨ New kubebuilder:skipDescriptions CRD marker

### LINKED ISSUE #441: Ability to skip descriptions per field or per package

armsnyder:
Signed-off-by: Adam Snyder <armsnyder@gmail.com>

### Overview

Closes #441 

I added a new marker `kubebuilder:skipDescriptions`. When added to a field it skips adding property descriptions for all nested properties under that field, recursively. When added to a package it skips adding property descriptions for all types in the package, recursively.

The benefit of this marker is it allows a user to embed types from external packages without causing the types to generate descriptions.

### Details

I had trouble implementing this because the marker system currently can only process one type at a time, meaning that markers do not have access to nested properties. This was a problem for removing descriptions in imported packages.

My solution was a new marker interface `PostFlattenMarker` that a marker can implement so that it gets invoked much later, after the JSON schema has been flattened.

### How this was tested

I added integration tests for both the package-type and field-type `kubebuilder:skipDescriptions` marker. The tests are modeled after the existing `gen_integration_test.go` test. They check the controller-gen output against go-generated CRD files.

(Half of this PR diff is testdata files.)

### DIFF
--- a/pkg/crd/markers/skip_descriptions.go
+++ b/pkg/crd/markers/skip_descriptions.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/crd/markers/skip_descriptions.go
+++ b/pkg/crd/markers/skip_descriptions.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package markers
+
+import (
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-tools/pkg/markers"
+)
+
+func init() {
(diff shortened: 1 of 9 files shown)


sathieu:
@armsnyder What is the status of this? You probably need to request a review (see https://github.com/kubernetes-sigs/controller-tools/pull/635#issuecomment-962556381).

armsnyder:
Thanks @sathieu, I wasn't sure that message was for me. The PR should be ready for review.

/assign @pwittrock

pkg/crd/markers/skip_descriptions.go
```diff
@@ -0,0 +1,63 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
christopherhein, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

pkg/crd/markers_integration_test.go
```diff
@@ -0,0 +1,114 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
christopherhein, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

pkg/crd/schema_post_flatten.go
```diff
@@ -0,0 +1,109 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
christopherhein, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

christopherhein (commented):
thanks for the tag @camilamacedo86, added some very basic comments re:copyright dates on the new files, otherwise this lgtm!

christopherhein:
Thanks @armsnyder ! 

/lgtm

armsnyder:
What's the next step for this PR?

armsnyder:
@christopherhein @camilamacedo86 Do we still want this feature?

zuzzas:
@armsnyder 
I've been using your changes to great effect! Thanks!

Unfortunately, the PR does not work on the latest master. Tests fail as well. Can you rebase, please?

andriisoldatenko:
@armsnyder is it something you are still interesting to contribute or are you okey if I try to re-apply your patch in new PR against main branch?

PR #640, opened 2021-11-11
🐛 add generateName in ObjectMeta

cvvz:
Sometimes `GenerateName` will be set to generate a unique name ONLY IF the `Name` field has not been provided.

### DIFF
--- a/pkg/crd/known_types.go
+++ b/pkg/crd/known_types.go
@@ -153,6 +153,9 @@ var ObjectMetaPackages = map[string]PackageOverride{
 						},
 					},
 				},
+				"generateName": {
+					Type: "string",
+				},
 			},
 		}
 	},


pkg/crd/known_types.go
```diff
@@ -153,6 +153,9 @@ var ObjectMetaPackages = map[string]PackageOverride{
 						},
 					},
 				},
+				"generateName": {
+					Type: "string",
+				},
```
camilamacedo86, line 158: > Sometimes GenerateName will be set to generate a unique name ONLY IF the Name field has not been provided.
Could you please describe when/why it is required?
Also, would be possible to ensure a test for your scenario?

cvvz: > Could you please describe when/why it is required?

`generateName` is used widely in Deployment, Replicaset, Daemonst and so on, to make sure the resource's name generated by the owner is unique on the server. In my situation, I want to create a resource with unique name, whose template is embed in my CRD.

> Also, would be possible to ensure a test for your scenario?

 I'v tested it manually, what kind of test would satisfy the requirement? Could you please show me an example?

PR #653, opened 2022-01-28
:sparkles: Add XValidations CRD marker for KEP-2876 support.

benluddy:
This allows generation of the x-kubernetes-validations extension for
expression-based validation rules.


benluddy:
> @benluddy: PR needs rebase.

:+1:

pkg/crd/testdata/cronjob_types.go
```diff
@@ -187,7 +187,12 @@ type CronJobSpec struct {
 
 	// This tests that both unexported and exported inline fields are not skipped in the schema generation
 	unexportedStruct `json:",inline"`
-	ExportedStruct `json:",inline"`
+	ExportedStruct   `json:",inline"`
+
+	// Test of the expression-based validation rule marker, with optional message.
+	// +kubebuilder:validation:XValidation:rule="self.size() % 2 == 0",message="must have even length"
+	// +kubebuilder:validation:XValidation:rule="true"
```
alvaroaleman, line 194: Is there any way for us to check if the `rule` contains a valid expression?

benluddy: It might be practical to check that the expression is at least _syntactically_ valid at generation time, but that would mean taking a dependency on a CEL parser.

The validity of a given expression depends simultaneously on the structure of the CEL expression itself and on where it sits within the schema, so it's complex enough that I'd be wary of rejecting a CRD at generation time that the API would have accepted.

Maybe we can revisit this question once x-k8s-validations graduates to beta (~1.25)?

alvaroaleman: Yeah, by "valid expression" I meant syntastically valid, if it is also semantically valid is likely hard to tell and probably error prone. Deferring for now sgtm.

alvaroaleman (approved):

PR #655, opened 2022-02-03
:seedling: Update cronjob_types.go go:generate invocation and re-generate.

benluddy:
This invocation now requires ignoreUnexportedFields=true. Regenerating
it produces a field-ordering- and whitespace-only manifest diff, but
the integration tests are only concerned with significant differences.


joelanford:
/lgtm
/approve

Thanks!

PR #666, opened 2022-04-28
✨ Bump to k8s.io v1.24.0-rc.1

sbueringer:


pkg/version/version.go
```diff
@@ -23,7 +23,7 @@ import (
 // Version returns the version of the main module
 func Version() string {
 	info, ok := debug.ReadBuildInfo()
-	if !ok {
+	if !ok || (info != nil && info.Main.Version == "") {
```
sbueringer, line 26: A unit test failed without this change

It looks like previously ok was false. Now we get ok==true + an info object with an empty Version

alvaroaleman: okay, interesting. That is IMHO a go bug as it breaks comoatibility.

alvaroaleman: Shouldn't it be
```suggestion
	if !ok || info == nil || info.Main.Version == "" {
```

alvaroaleman (approved):

pkg/version/version.go
AlmogBaku: see #662 for explanation for this issue

PR #673, opened 2022-05-06
✨Added skip build tag in pkg/deepcopy/zz_generated.markerhelp.go to ignore the file in kubebuilder default scaffold for v3 test coverage

NikhilSharmaWe:
### Description
- This PR adds a skip build tag in `pkg/deepcopy/zz_generated.markerhelp.go` to ignore [zz_generated.deepcopy.go](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/api/v1/zz_generated.deepcopy.go) in [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) default scaffold for v3 test coverage. 
- We need to merge this PR and get a new release, so that we can update the new version in [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder).

### Motivation
- Regarding https://github.com/kubernetes-sigs/kubebuilder/issues/2645


NikhilSharmaWe:
@camilamacedo86 Here is the PR adding skip build tag in [https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/deepcopy/zz_generated.markerhelp.go](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/deepcopy/zz_generated.markerhelp.go#L1-L2).

pkg/deepcopy/zz_generated.markerhelp.go
```diff
@@ -1,5 +1,6 @@
 //go:build !ignore_autogenerated
 // +build !ignore_autogenerated
+// +build !skip
```
erikgb, line 3: Further down in this file it states:

> Code generated by helpgen. DO NOT EDIT.

I think this change should be added somewhere else in the codebase?

NikhilSharmaWe: @erikgb Thanks for the review.
Could you please take another look.

cmd/helpgen/main.go
```diff
@@ -38,6 +38,7 @@ const (
 	markerName = "controllertools:marker:generateHelp"
 	header     = `
 // +build !ignore_autogenerated
+// +build !skip
```
erikgb, line 41:
```suggestion
// +build !cover
```

I am not an expert on this, but IMO the command that motivates this change would look better with another build tag? Compare `go test -cover -tags skip` to `go test -cover -tags cover`. This is only relevant for the test coverage, right?

And why doesn't the existing tag (`ignore_autogenerated`)work,` ref. https://github.com/kubernetes-sigs/kubebuilder/issues/2645#issuecomment-1117057817? Will this just work by adding a new build tag?

NikhilSharmaWe: @Sher-Chowdhury Could you please give your thoughts here.

camilamacedo86: @NikhilSharmaWe, the @erikgb suggestion also makes sense for me.
WDYT about you try to reproduce the issue https://github.com/kubernetes-sigs/kubebuilder/issues/2645#issuecomment-1117057817 and see if you can use the tag `ignore_autogenerated` instead of the skip to sort it out?

camilamacedo86, line 41: @NikhilSharmaWe:

If we skip the build files, the tests will fail:

```
r$ go test ./... -cover -tags skip
?   	sigs.k8s.io/kubebuilder/testdata/project-v3	[no test files]
# sigs.k8s.io/kubebuilder/testdata/project-v3/api/v1
api/v1/zz_generated.deepcopy.go:2: +build lines do not match //go:build condition
FAIL	sigs.k8s.io/kubebuilder/testdata/project-v3/api/v1 [build failed]
ok  	sigs.k8s.io/kubebuilder/testdata/project-v3/controllers	11.584s	coverage: 0.0% of statements
FAIL
```

So, I do not think we can address the request made in the Kubebuilder because of the coverage. 
In the comment: https://github.com/kubernetes-sigs/kubebuilder/pull/2668#issuecomment-1119482191, I explained how/where/when it is and if we wish to change what is the process. 

But I do not think that we should address this change here. 
Let's close this one and see how can we address/help who asked it in the Kubebuilder side?

PR #674, opened 2022-05-10
✨Bumped go 1.18 

NikhilSharmaWe:
### Description
Bumped go 1.18 to get a new release and for adding it to https://github.com/kubernetes-sigs/kubebuilder.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/controller-tools
 
-go 1.17
+go 1.18
 
 require (
 	github.com/fatih/color v1.12.0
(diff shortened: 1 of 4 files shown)


camilamacedo86:
/hold

More info: https://kubernetes.slack.com/archives/C02MRBMN00Z/p1652178727902549?thread_ts=1652177106.531319&cid=C02MRBMN00Z

We should follow the same approach in Kubebuilder.
So, we will not bump the go.mod files of the Kubebuilder CLI to allow people to consume.

camilamacedo86:
@NikhilSharmaWe,  

Let's close this one. I do not think that we can do it now. Why?
Same reason of controller-runtime:

> The version in the go.mod file is the minimum go version (https://go.dev/doc/modules/gomod-ref#go). By keeping it on 1.17 we allow consumers  to use Go 1.17+.

PR #681, opened 2022-05-25
🐛 crdgen: compare metav1 pkg by ID

### LINKED ISSUE #680: controller-tools v0.9.0 does not generate CRDs for additional paths
The new controller-gen version only generates CRDs for the first specified path.

An example:

* Checkout https://github.com/kubernetes-sigs/cluster-api
* Run controller-gen:
```bash
controller-gen-v0.9.0 \
                paths=./api/... \
                paths=./exp/api/... \
                crd:crdVersions=v1 \
                output:crd:dir=./config/crd/bases
```
Only CRDs from the `./api/` folder are generated/updated none of `./exp/api/...` (i.e. the MachinePool CRD).

I took a closer look at the code and it looks like the metav1Pkg does not match when checking if MachinePool has ObjectMeta and TypeMeta:
https://github.com/kubernetes-sigs/controller-tools/blob/0fba82b41ad698c536c5b2e3e87ff3fa1908d7ff/pkg/crd/gen.go#L250

Not sure if it would be a correct fix, but by changing the line to
```go
			if fieldPkg != nil && fieldPkg.PkgPath != metav1Pkg.PkgPath {
```
controller-gen was updating the MachinePool CRD again.

I think it's an additional issue, but after introducing this fix controller-gen did update the MachinePool CRD but it ignored the `+kubebuilder:validation:MinLength=1` annotation (i.e. it didn't include minLength in the CRD anymore)

akutz:
This patch updates the way the CRD generator compares an imported metav1 package. Previously the comparison occurred using a Golang equality operator, !=, against two, in-memory data structures. However, this fails when multiple root paths are loaded. Their metav1 packages are identical, just not identical objects in memory. This patch updates the comparison to compare the package IDs, not the instance of the object.

This patch also introduces a filesystem-path specific loader for each unique filesystem path provided as a root. This ensures the AST is loaded correctly and the kubebuilder markers are parsed as intended.

For more information please refer to [this Slack thread](https://kubernetes.slack.com/archives/C02MRBMN00Z/p1653467125115659).

Fixes #680


ahmetb:
needs to be rebased on top of #682

PR #683, opened 2022-05-25
:bug: Add XPreserveUnknownFields to runtime.RawExtension

eddycharly:
This PR adds `XPreserveUnknownFields` to `runtime.RawExtension`.

`runtime.RawExtension` has `type: object` but all fields will be pruned unless `x-kubernetes-preserve-unknown-fields: true` is specified.

It seems reasonable to add `x-kubernetes-preserve-unknown-fields: true` by default, as was done with `apiextensions.JSON` for example https://github.com/kubernetes-sigs/controller-tools/blob/a260f1391b0f722e07ae04fe87e23e32bc1a1ec7/pkg/crd/known_types.go#L106-L111


joelanford:
I think my main concern is that anyone already using runtime.RawExtension with controller-gen is already generating a CRD schema without `x-kubernetes-preserve-unknown-fields`. This PR would cause those users to get a potentially undesired CRD schema change.

At a minimum, I think I'd label this as a breaking change rather than a bug fix. We _might_ also need to consider making this behavior opt-in.

@benluddy was actually chatting with me the other day about a similar request to improve the auto-generated schema for `time.Duration`, which would make the schema more strict to align with the expectations of `time.ParseDuration`.

In this case, we would start allowing fields not mentioned in the schema to be preserved, which seems a little bit different than making a schema more strict.

Maybe @alvaroaleman and @vincepri have opinions?

alvaroaleman:
> I think my main concern is that anyone already using runtime.RawExtension with controller-gen is already generating a CRD schema without x-kubernetes-preserve-unknown-fields. This PR would cause those users to get a potentially undesired CRD schema change.

But if ppl do that without `x-kubernetes-preserve-unknown-fields.` then everything in that field will be pruned today or not, as its schema is unknown at the CRD level?

/ok-to-test

If we merge this, it  will need a test btw.

eddycharly:
I wonder how one can use `runtime.RawExtension` without `x-kubernetes-preserve-unknown-fields`.

It looks like it just doesn’t work to me, all fields get pruned and the object ends up empty 🤔

eddycharly:
> But if ppl do that without x-kubernetes-preserve-unknown-fields. then everything in that field will be pruned today or not, as its schema is unknown at the CRD level?

Exactly what we observed, we ended up with empty objects all fields were pruned.

pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
```diff
@@ -7340,9 +7341,11 @@ spec:
                 minLength: 4
                 type: string
               unprunedEmbeddedResource:
+                allOf:
```
eddycharly, line 7344: This looks strange, not sure where I should look at.

eddycharly: @joelanford @alvaroaleman do you have an idea where I should look in the code to find why this gets transformed into `allOf` ?

eddycharly: From what I understand it happens in the flattener. I’m a bit lost in the code though. Any hint would be appreciated.

eddycharly: I think it happens when `kubebuilder:pruning:PreserveUnknownFields` is present on a field and the target type. I can reproduce with any struct. Is this a bug ? Should both markers get merged when generating the CRD ? cc @alvaroaleman @joelanford

FillZpp:
This PR could solve the question asked on stackoverflow and slack https://stackoverflow.com/questions/72577439/k8s-operator-read-raw-data .

Thanks @eddycharly , and this should be merged after #689 to avoid the `allOf: x-kubernetes-preserve-unknown-fields`. Am I right?

eddycharly:
@FillZpp Yes #689 was a prerequisite for this one.

eddycharly:
Rebased and updated manifest.

alvaroaleman (approved):
/lgtm
/approve

alvaroaleman:
thanks!

eddycharly:
@alvaroaleman do I need to create a cherry pick PR ?

PR #687, opened 2022-05-31
:running: Simplify the LoadRootsWithConfig logic

akutz:
This patch simplifies the LoadRootsWithConfig logic to be closer to the logic that existed prior to the patch that addressed support for multiple, distinct root paths. The fix that was merged was overly complex, and not necessarily so. This update simplifies that patch to the necessary bits.

Please see #680 and #686 for more information.

cc @alvaroaleman


alvaroaleman (approved):
thanks!

PR #689, opened 2022-06-02
:bug: Allow setting PreserveUnknownFields at both type and field level

### LINKED ISSUE #688: Setting PreserverUnknownFields marker on both type and field generates an invalid schema
When setting the `kubebuilder:pruning:PreserveUnknownFields` on a struct and a field using that struct, the generated schema is not valid.

For example:
```go
// +kubebuilder:pruning:PreserveUnknownFields
type Preserved struct {
	ConcreteField string                 `json:"concreteField"`
	Rest          map[string]interface{} `json:"-"`
}

type Spec struct {
	// +kubebuilder:pruning:PreserveUnknownFields
	Preserved Preserved `json:"preserved"`
}

type Test struct {

	Spec Spec `json:"spec,omitempty"`
}
```

Will generate the following schema:
```yaml
apiVersion: apiextensions.k8s.io/v1
metadata:
  annotations:
  creationTimestamp: null
  name: tests.testdata.kubebuilder.io
spec:
  group: testdata.kubebuilder.io
  names:
    kind: Test
    listKind: TestList
    plural: tests
    singular: test
  scope: Namespaced
  versions:
  - name: v1
    schema:
      openAPIV3Schema:
        properties:
          apiVersion:
            type: string
          kind:
            type: string
          metadata:
            type: object
          spec:
            properties:
              preserved:
                allOf:
                - x-kubernetes-preserve-unknown-fields: true
                - x-kubernetes-preserve-unknown-fields: true
                properties:
                  concreteField:
                    type: string
                required:
                - concreteField
                type: object
            required:
            - preserved
            type: object
        type: object
    served: true
    storage: true
```

The problematic part being:
```yaml
             preserved:
                allOf:
                - x-kubernetes-preserve-unknown-fields: true
                - x-kubernetes-preserve-unknown-fields: true
```

From what I observed, this happens when flattening the schema.
I managed to fix the bug by adding a case for `XPreserveUnknownFields` here to prevent hoisting: https://github.com/kubernetes-sigs/controller-tools/blob/f89071af31308871237b805a4d715e8fa812e263/pkg/crd/flatten.go#L105-L156

While this fixed the bug, I'm not too sure about the fix itself.

I can open a PR but would like to confirm it's a bug first and eventually if the fix I tried makes sense.

cc @akutz @sbueringer (per Alvaro's recommendation).

eddycharly:
This PR allows setting `PreserveUnknownFields` at both type and field level.

Fixes #688

### DIFF
--- a/pkg/crd/flatten.go
+++ b/pkg/crd/flatten.go
@@ -143,6 +143,8 @@ func flattenAllOfInto(dst *apiext.JSONSchemaProps, src apiext.JSONSchemaProps, e
 				dstProps.Schema = &apiext.JSONSchemaProps{}
 			}
 			flattenAllOfInto(dstProps.Schema, *srcProps.Schema, errRec)
+		case "XPreserveUnknownFields":
+			dstField.Set(srcField)
 		// NB(directxman12): no need to explicitly handle nullable -- false is considered to be the zero value
 		// TODO(directxman12): src isn't necessarily the field value -- it's just the most recent allOf entry
 		default:
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -134,8 +134,12 @@ type CronJobSpec struct {
 	// +kubebuilder:validation:nullable
 	UnprunedEmbeddedResource runtime.RawExtension `json:"unprunedEmbeddedResource"`
 
-	// This tests that a type-level pruning maker works.
+	// This tests that a type-level pruning marker works.
 	UnprunedFromType Preserved `json:"unprunedFomType"`
+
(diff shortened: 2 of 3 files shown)


eddycharly:
@FillZpp did you have time to look at this PR ?

FillZpp:
> @FillZpp did you have time to look at this PR ?

Sure, it looks good to me. You may need another approver to review it.

/lgtm
/cc @alvaroaleman

alvaroaleman (approved):
/lgtm
/approve
thanks!

FillZpp:
@eddycharly You may have to do this cherry-pick manually.

PR #691, opened 2022-06-13
✨ Support api-approved annotation for CRD with k8s group

### LINKED ISSUE #656: Add support for adding annotations to CRD
For kubernetes projects, if we want to use the `sig.k8s.io` group for CRDs, we need to provide the annotation `api-approved.kubernetes.io`. Otherwise, apiserver rejects the CRD.

As a result, we have to manually edit the CRDs produced by `controller-gen` and continuously edit the files every time there is a type update.

Using kustomize is not possible, because the env-test doesn't load them and it could be problematic to do so (it's an additional dependency).

/kind feature

FillZpp:
Support api-approved annotation for CRD with k8s group

fixes #656

### DIFF
--- a/pkg/crd/markers/crd.go
+++ b/pkg/crd/markers/crd.go
@@ -18,6 +18,7 @@ package markers
 
 import (
 	"fmt"
+	"strings"
 
 	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
@@ -51,6 +52,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:deprecatedversion", markers.DescribesType, DeprecatedVersion{})).
 		WithHelp(DeprecatedVersion{}.Help()),
+
+	must(markers.MakeDefinition("kubebuilder:metadata", markers.DescribesType, Metadata{})).
+		WithHelp(Metadata{}.Help()),
 }
 
 // TODO: categories and singular used to be annotations types
@@ -343,5 +347,41 @@ func (s DeprecatedVersion) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec, 
 		ver.DeprecationWarning = s.Warning
 		break
 	}
+	return nil
+}
+
+// +controllertools:marker:generateHelp:category=CRD
+
+// Metadata configures the additional annotations or labels for this CRD.
+// For example adding annotation "api-approved.kubernetes.io" for a CRD with Kubernetes groups,
+// or annotation "cert-manager.io/inject-ca-from-secret" for a CRD that needs CA injection.
+type Metadata struct {
+	// Annotations will be added into the annotations of this CRD.
+	Annotations []string `marker:",optional"`
+	// Labels will be added into the labels of this CRD.
+	Labels []string `marker:",optional"`
+}
+
+func (s Metadata) ApplyToCRD(crd *apiext.CustomResourceDefinition, version string) error {
+	if len(s.Annotations) > 0 {
+		if crd.Annotations == nil {
+			crd.Annotations = map[string]string{}
+		}
+		for _, str := range s.Annotations {
+			kv := strings.SplitN(str, "=", 2)
+			crd.Annotations[kv[0]] = kv[1]
+		}
+	}
+
+	if len(s.Labels) > 0 {
+		if crd.Labels == nil {
+			crd.Labels = map[string]string{}
+		}
+		for _, str := range s.Labels {
+			kv := strings.SplitN(str, "=", 2)
+			crd.Labels[kv[0]] = kv[1]
+		}
+	}
+
 	return nil
 }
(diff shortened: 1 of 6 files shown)


FillZpp:
/retitle ✨ Support api-approved annotation for CRD with k8s group

pkg/crd/markers/crd.go
```diff
@@ -51,6 +52,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:deprecatedversion", markers.DescribesType, DeprecatedVersion{})).
 		WithHelp(DeprecatedVersion{}.Help()),
+
+	must(markers.MakeDefinition("kubebuilder:apiapprovedk8sgroup", markers.DescribesType, APIApprovedKubernetesGroup{})).
```
alvaroaleman, line 56: Why the`k8sgroup` suffix?

FillZpp: Emm, I supposed it to be short name of the annotation key `api-approved.kubernetes.io`. Do you think `k8sgroupapprovedapi` would be better?

pkg/crd/markers/crd.go
```diff
@@ -345,3 +349,29 @@ func (s DeprecatedVersion) ApplyToCRD(crd *apiext.CustomResourceDefinitionSpec,
...
+// +controllertools:marker:generateHelp:category=CRD
+
+// APIApprovedKubernetesGroup configures the api-approved.kubernetes.io for a CRD
+// with Kubernetes groups, including "k8s.io", "*.k8s.io", "kubernetes.io", "*.kubernetes.io".
+type APIApprovedKubernetesGroup struct {
+	// URL is the value of api-approved.kubernetes.io annotation on CRD,
+	// it should be a valid URL which contains Scheme and Host and
+	// should not start with "unapproved".
+	URL string `marker:"URL"`
+}
```
joelanford, line 362: Any chance we could make this a more generic `Annotation` struct with `Key` and `Value` fields that we parse out of the marker?

I can imagine a few other use cases, for example adding annotation `cert-manager.io/inject-ca-from-secret` to the CRD when it needs a CA injected for a conversion webhook configuration.

FillZpp: Thanks, it makes sense. I will update this later.

FillZpp:
@joelanford @alvaroaleman I have made it more generic for additional annotations or labels. PTAL :)

pkg/crd/testdata/cronjob_types.go
```diff
@@ -499,6 +499,7 @@ type CronJobStatus struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:singular=mycronjob
 // +kubebuilder:storageversion
+// +kubebuilder:additionalmetadata:annotations="api-approved.kubernetes.io=https://github.com/kubernetes-sigs/controller-tools";"cert-manager.io/inject-ca-from-secret=cert-manager/cert-manager-webhook-ca"
```
camilamacedo86, line 502: Why the name is additionalmetada:annotation?

camilamacedo86, line 502:
```suggestion
// +kubebuilder:metadata:annotations="api-approved.kubernetes.io=https://github.com/kubernetes-sigs/controller-tools";"cert-manager.io/inject-ca-from-secret=cert-manager/cert-manager-webhook-ca"
```

Do we need the additional? 
Why not only annotations? 
So, if set we add annotations informed 
If not set, we add only the default annotation

FillZpp: Emm.. okey, I will update it to `// +kubebuilder:metadata:annotations=...`

pkg/crd/markers/crd.go
```diff
@@ -51,6 +52,9 @@ var CRDMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("kubebuilder:deprecatedversion", markers.DescribesType, DeprecatedVersion{})).
 		WithHelp(DeprecatedVersion{}.Help()),
+
+	must(markers.MakeDefinition("kubebuilder:additionalmetadata", markers.DescribesType, AdditionalMetadata{})).
+		WithHelp(AdditionalMetadata{}.Help()),
```
camilamacedo86, line 57: `Are we not adding annotations only?
What about the name here be kubebuilder:annotations

And we Annotations{}

camilamacedo86: Ps is possible to add annotations outside of the metadata? 
Do we need to use this term here or is that redundant?

FillZpp: Now it supports to add annotations and labels

```
// +kubebuilder:metadata:annotations=...
// +kubebuilder:metadata:labels=...
```

Not sure if there will be some other fields in metadata have to be added there. So we'd better put all of them in `kubebuilder:metadata`.

pkg/crd/testdata/cronjob_types.go
FillZpp: Updated

camilamacedo86 (approved):
That seems great for me 🥇 
/lgtm

alvaroaleman (approved):

Issue #435, opened 2020-04-30
Flaky behaviour using custom slice types and listType marker


avorima:
`controller-gen` will sometimes fail to create the CRD manifests when using a custom slice type, e.g. `type Foos []Foo` and the `listType` and `listMapKey` markers.

To reproduce this create a sample project with kubebuilder v2.3.1:
```
kubebuilder init --repo temp
kubebuilder create api --group ship --version v1beta1 --kind Frigate
```

and then add the following to the `api/v1beta1/frigate_types.go` file:

```go
// FrigateStatus defines the observed state of Frigate
type FrigateStatus struct {
        // +kubebuilder:validation:Type=array
        // +optional
        // +listType=map
        // +listMapKey=type
        Conditions Conditions `json:"conditions,omitempty"`
}

type Conditions []Condition

type Condition struct {
        Type   string `json:"type,omitemtpy"`
        Status string `json:"status,omitempty"`
}
```

Running `make manifests` will then on occasion yield the following errors:

```
/tmp/test/api/v1beta1/frigate_types.go:35:2: must apply listType to an array
/tmp/test/api/v1beta1/frigate_types.go:35:2: must apply listMapKey to an associative-list
```

If the `+kubebuilder:validation:Type=array` marker is above the type definition it fails consistently.

I have verified that this occurs with both `v0.2.5` and `v0.3.0`.

The issue seem to come the fact that both [`listType`](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/topology.go#L114) and [`kubebuilder:validation:Type`](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/validation.go#L324) markers implement `ApplyFirst` to ensure they get applied before all other markers. But since [applyMarkers](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/schema.go#L111) loops over a map, the order is random.

avorima:
This issues still is present in 0.6.0

/reopen

chrischdi:
/repoen

The issue is still reproducible on the latest commit (`529c857f74b225412836567200cd0ae003fbb3b3`).

In my case I just want to add the two markers `+listType=map` and `+listMapKey=id` to allow the slice to get used together with Server-Side Apply.

In fact, when using a custom slice type it is not even possible to add the two markers.

The following variant simply get ignored by `controller-gen`:
```go
type BarSpec struct {
[...]

PR #692, opened 2022-06-15
🐛 Make topology markers also valid for type definitions

chrischdi:
This PR allows setting `listMapKey`, `listType`, `mapType` and `structType` at both type and field level to allow setting them at typed slices too.

Currently the markers are only valid at the field level.

The markers are important for defining CRDs which should get used in combination with Server-Side Apply ([xref SSA](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy)).

Fixes: #435

**Example**

When a struct references a typed slice it is currently not possible to set the markers.
Example:

```go
type BarSpec struct {
	Foos Foos `json:"foos,omitempty"`
}

type Foos []Foo
```

* When setting it at field level, controller-gen will throw errors because the detected `type` is empty / `""`
  ```
  /tmp/test-ct/api/v1beta1/bar_types.go:32:2: must apply listType to an array, found
  /tmp/test-ct/api/v1beta1/bar_types.go:32:2: must apply listMapKey to an array, found
  ```
* At type level the markers are currently ignored.

After this PR, setting the markers will be possible at type level too, where the type information will be available:

```go
type BarSpec struct {
	Foos Foos `json:"foos,omitempty"`
}

// +listType=map
// +listMapKey=id
type Foos []Foo
```


FillZpp:
@chrischdi Thanks. Please add some test cases in integration test, and you may also have to test the case that the markers defined on both type and field, like this:

```golang
type BarSpec struct {
        // +listType=map
        // +listMapKey=id
	Foos Foos `json:"foos,omitempty"`
}

// +listType=map
// +listMapKey=id
type Foos []Foo
```

chrischdi:
Hi @FillZpp , thank you for the feedback 👍 

I added a test case similar to
```go
// +listType=map
// +listMapKey=id
type Foos []Foo
```

However it is not possible to add:
```go
type BarSpec struct {
	// +listType=map
	// +listMapKey=name
	// +listMapKey=secondary
	Foos Foos `json:"foos,omitempty"`
}
```

Because this will fail at the following marker checks: 
* https://github.com/kubernetes-sigs/controller-tools/blob/9e2b90e643404d33a7c8f18fbd19e6e8a00f2834/pkg/crd/markers/topology.go#L112
* https://github.com/kubernetes-sigs/controller-tools/blob/9e2b90e643404d33a7c8f18fbd19e6e8a00f2834/pkg/crd/markers/topology.go#L126

The information if it is an array or not is not available there (it is empty string in this case), which is why this PR introduces the possibility to set the markers at the type definition instead.

alvaroaleman (approved):

PR #697, opened 2022-06-29
🐛 Fix crd flattening for XMapType to not duplicate entries

chrischdi:
Bug was introduced in #693.

As outlined in https://github.com/kubernetes-sigs/controller-tools/pull/693#issuecomment-1169433662 , #693 introduced the bug that `x-kubernetes-map-type` may now get set twice and thus result in two entries in `allOf`.

This is because before #693 the marker at the type definition was simply ignored but now they got combined.

This PR fixes the flattening to handle this case for the `structType` marker.

A similar fix for the markers `mapType`, `listType` and `listMapKey` is not necessary because setting them where the type is referenced would result in an error anyway. Before #693 this was also the case, but since #693 it is possible to set them at the type definition.


chrischdi:
cc @robbie-demuth : could you give this a try?

alvaroaleman (approved):
/lgtm
/approve

Issue #676, opened 2022-05-12
[struct,list,map]Type markers on types are ignored


jan--f:
Let me preface this by admitting that I'm not exactly sure if this is a bug or expected behavior. I might also just not use the tooling correctly to get the outcome I want.

Server-side-apply makes use of markers (like `structType=atomic`) to merge resources and partial updates to them. controller-tools already [respects those markers on fields](https://github.com/kubernetes-sigs/controller-tools/blob/a260f1391b0f722e07ae04fe87e23e32bc1a1ec7/pkg/crd/testdata/cronjob_types.go#L151) and produces the [respective yaml annotations](https://github.com/kubernetes-sigs/controller-tools/blob/a260f1391b0f722e07ae04fe87e23e32bc1a1ec7/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml#L7321).

When I mark the type `NestedObject` though the yaml annotation is not generated. To test this one can apply the following patch and just run `make`, which will now produce an error.
``` patch
diff --git a/pkg/crd/testdata/cronjob_types.go b/pkg/crd/testdata/cronjob_types.go
index 66222f6e..5d28063d 100644
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -148,7 +148,6 @@ type CronJobSpec struct {
        MapOfInfo map[string][]byte `json:"mapOfInfo"`

        // A struct that can only be entirely replaced
-       // +structType=atomic
        StructWithSeveralFields NestedObject `json:"structWithSeveralFields"`

        // This tests that type references are properly flattened
@@ -258,6 +257,7 @@ func (p *Preserved) MarshalJSON() ([]byte, error) {
        return json.Marshal(full)
 }

+// +structType=atomic
 type NestedObject struct {
        Foo string `json:"foo"`
        Bar bool   `json:"bar"`
```

Now I think the two variants I described should produce the same results. For example if I use `metav1.LabelSelector` in a type declaration (from which a CRD manifest is generated), I currently have to explicitly add the marker to the field of that type. However the [type already has the marker that I want](https://github.com/kubernetes/apimachinery/blob/97e5df2d0258ac077093867fabf508c1fc31f53b/pkg/apis/meta/v1/types.go#L1181) and I think there's an argument to be made for letting the type define this behavior.

[...]

PR #704, opened 2022-07-06
:sparkles: testdata/cronjob_types.go: add NestedStructWithSeveralFieldsDoubleMarked

jan--f:
This adds a field to the cronjob test data, that ensures that fields
which are marked twice (once through a type marker, one through a field
marker) still generate the correct yaml structure.

Signed-off-by: Jan Fajerski <jfajersk@redhat.com>

### DIFF
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -162,7 +162,6 @@ type CronJobSpec struct {
 	StructWithSeveralFields NestedObject `json:"structWithSeveralFields"`
 
 	// A struct that can only be entirely replaced via a nested type.
-	// +structType=atomic
 	NestedStructWithSeveralFields NestedStructWithSeveralFields `json:"nestedStructWithSeveralFields"`
 
 	// This tests that type references are properly flattened
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -164,6 +164,11 @@ type CronJobSpec struct {
 	// A struct that can only be entirely replaced via a nested type.
 	NestedStructWithSeveralFields NestedStructWithSeveralFields `json:"nestedStructWithSeveralFields"`
 
+	// A struct that can only be entirely replaced via a nested type and
+	// field markers.
+	// +structType=atomic
+	NestedStructWithSeveralFieldsDoubleMarked NestedStructWithSeveralFields `json:"nestedStructWithSeveralFieldsDoubleMarked"`
+
 	// This tests that type references are properly flattened
 	// +kubebuilder:validation:optional
 	JustNestedObject *JustNestedObject `json:"justNestedObject,omitempty"`
--- a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
+++ b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
@@ -7290,6 +7290,19 @@ spec:
                 - foo
                 type: object
                 x-kubernetes-map-type: atomic
+              nestedStructWithSeveralFieldsDoubleMarked:
+                description: A struct that can only be entirely replaced via a nested
+                  type and field markers.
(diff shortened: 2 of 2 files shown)


FillZpp:
I'm not sure, should we keep this test to verify that annotated on both field and type, to avoid bugs like these #688 and https://github.com/kubernetes-sigs/controller-tools/pull/693#issuecomment-1169433662

jan--f:
> I'm not sure, should we keep this test to verify that annotated on both field and type, to avoid bugs like these #688 and [#693 (comment)](https://github.com/kubernetes-sigs/controller-tools/pull/693#issuecomment-1169433662)

Yes that makes sense, I added a commit to that effect. WDYT?

jan--f:
ping @FillZpp Any further concerns about this?

FillZpp:
@jan--f I don't understand what's the difference between `NestedStructWithSeveralFields` and `NestedStructWithSeveralFieldsDoubleMarked `?

jan--f:
> @jan--f I don't understand what's the difference between `NestedStructWithSeveralFields` and `NestedStructWithSeveralFieldsDoubleMarked `?

@FillZpp Right, because I forgot to add the field marker. Added it now, thank you!

jan--f:
@chrischdi anything still blocking this?

chrischdi:
/lgtm

from my side. But requires a maintainer for approval (which I am not).

jan--f:
Thanks, let me try @pwittrock and @mengqiy for an `approve` then.

pkg/crd/testdata/cronjob_types.go
```diff
@@ -162,9 +162,13 @@ type CronJobSpec struct {
 	StructWithSeveralFields NestedObject `json:"structWithSeveralFields"`
 
 	// A struct that can only be entirely replaced via a nested type.
-	// +structType=atomic
 	NestedStructWithSeveralFields NestedStructWithSeveralFields `json:"nestedStructWithSeveralFields"`
 
+	// A struct that can only be entirely replaced via a nested type and
+	// field markers.
+	// +structType=atomic
+	NestedStructWithSeveralFieldsDoubleMarked NestedStructWithSeveralFields `json:"nestedStructWithSeveralFieldsDoubleMarked"`
```
camilamacedo86, line 170: It seems a workarround.  Could you please raise an issue or add in the description:

a) what you were doing when you face the error ( with an example code )
b) what do you expected to see ?
c) what was the behaviour faced instead? 

Also, if we move forward here I think we need a better name than `FieldsDoubleMarked`. DoubleMarket with what? (i.e. NestedStructWithSeveralFieldsMarkedWithAB)

pkg/crd/testdata/cronjob_types.go
```diff
@@ -162,9 +162,13 @@ type CronJobSpec struct {
 	StructWithSeveralFields NestedObject `json:"structWithSeveralFields"`
 
 	// A struct that can only be entirely replaced via a nested type.
-	// +structType=atomic
```
camilamacedo86, line 165: Why it was removed?

camilamacedo86 (changes requested):
It seems requires more info first.

pkg/crd/testdata/cronjob_types.go
jan--f: The context of this is https://github.com/kubernetes-sigs/controller-tools/issues/676, which is a duplicate of https://github.com/kubernetes-sigs/controller-tools/issues/435.

jan--f: @camilamacedo86 so this just adds an additional test case. `NestedStructWithSeveralFields` ensures that the resulting manifest carries the `atomic` marker from the type. `NestedStructWithSeveralFieldsDoubleMarked` ensures the same thing only that the marker is applied to the type, as well as the to the field of the struct.

alvaroaleman (approved):

alvaroaleman:
Thanks!

PR #706, opened 2022-07-15
:sparkles: Markers can now indicate their priority when applying

dprotaso:
There's a need for more control over when SchemaMarkers are applied to a Schema. Right now there are two levels - 'First' and everything else. The iteration over the markers within this level is non-deterministic because we're iterating over a golang map type.

As an alternative we now collect all the markers and sort them based on a new priority. Markers can signal their priority by implementing the optional interface

```go

type ApplyPriority int64

type ApplyPriorityMarker interface {
 	ApplyPriority() ApplyPriority
 }
```

This change is backwards compatible - downstream markers that implement `ApplyFirst()` will behave the same as before.

### DIFF
--- a/pkg/crd/markers/doc.go
+++ b/pkg/crd/markers/doc.go
@@ -31,15 +31,15 @@ limitations under the License.
 // ApplyPriorityFirst constants. Following is an example of how to
 // implement such a marker:
 //
-//  type MyCustomMarker string
+//	type MyCustomMarker string
 //
-//  func (m MyCustomMarker) ApplyPriority() ApplyPriority {
-//  	return ApplyPriorityFirst
-//  }
+//	func (m MyCustomMarker) ApplyPriority() ApplyPriority {
+//		return ApplyPriorityFirst
+//	}
 //
-//  func (m MyCustomMarker) ApplyToSchema(schema *apiext.JSONSchemaProps) error {
-//    ...
-//  }
+//	func (m MyCustomMarker) ApplyToSchema(schema *apiext.JSONSchemaProps) error {
+//	  ...
+//	}
 //
 // All validation markers start with "+kubebuilder:validation", and
 // have the same name as their type name.
--- a/pkg/crd/markers/doc.go
+++ b/pkg/crd/markers/doc.go
@@ -47,7 +47,7 @@ limitations under the License.
 // # CRD Markers
 //
 // Markers that modify anything in the CRD itself *except* for the schema
-// implement ApplyToCRD (crd.CRDMarker).  They are expected to detect whether
+// implement ApplyToCRD (crd.SpecMarker).  They are expected to detect whether
 // they should apply themselves to a specific version in the CRD (as passed to
 // them), or to the root-level CRD for legacy cases.  They are applied *after*
 // the rest of the CRD is computed.
--- a/pkg/crd/markers/doc.go
+++ b/pkg/crd/markers/doc.go
@@ -19,7 +19,7 @@ limitations under the License.
 //
 // All markers related to CRD generation live in AllDefinitions.
 //
-// Validation Markers
+// # Validation Markers
 //
 // Validation markers have values that implement ApplyToSchema
 // (crd.SchemaMarker).  Any marker implementing this will automatically
@@ -28,7 +28,8 @@ limitations under the License.
 // implement ApplyFirst, but this is discouraged and may change
 // in the future. It is recommended to implement the ApplyPriority
 // interface in combination with ApplyPriorityDefault and
-// ApplyPriorityFirst constants.
+// ApplyPriorityFirst constants. Following is an example of how to
+// implement such a marker:
 //
 //  type MyCustomMarker string
 //
@@ -39,7 +40,7 @@ limitations under the License.
 //  func (m MyCustomMarker) ApplyToSchema(schema *apiext.JSONSchemaProps) error {
 //    ...
 //  }
-
+//
 // All validation markers start with "+kubebuilder:validation", and
 // have the same name as their type name.
 //
--- a/pkg/crd/markers/doc.go
+++ b/pkg/crd/markers/doc.go
@@ -25,8 +25,8 @@ limitations under the License.
 // (crd.SchemaMarker).  Any marker implementing this will automatically
 // be run after the rest of a given schema node has been generated.
 // Markers that need to be run before any other markers can also
-// implement ApplyFirst, but this is discouraged and may change
-// in the future.
+// implement ApplyFirst or markers.ApplyPriority, but this is discouraged
+// and may change in the future.
 //
 // All validation markers start with "+kubebuilder:validation", and
 // have the same name as their type name.
--- a/pkg/crd/markers/doc.go
+++ b/pkg/crd/markers/doc.go
@@ -25,13 +25,25 @@ limitations under the License.
 // (crd.SchemaMarker).  Any marker implementing this will automatically
 // be run after the rest of a given schema node has been generated.
 // Markers that need to be run before any other markers can also
-// implement ApplyFirst or markers.ApplyPriority, but this is discouraged
-// and may change in the future.
+// implement ApplyFirst, but this is discouraged and may change
+// in the future. It is recommended to implement the ApplyPriority
+// interface in combination with ApplyPriorityDefault and
+// ApplyPriorityFirst constants.
 //
+//  type MyCustomMarker string
+//
+//  func (m MyCustomMarker) ApplyPriority() ApplyPriority {
+//  	return ApplyPriorityFirst
+//  }
+//
+//  func (m MyCustomMarker) ApplyToSchema(schema *apiext.JSONSchemaProps) error {
+//    ...
+//  }
+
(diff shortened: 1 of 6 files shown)


dprotaso:
Hey folks, just following up - is there anything else needed for this PR?

pkg/crd/markers/doc.go
```diff
@@ -25,8 +25,8 @@ limitations under the License.
 // (crd.SchemaMarker).  Any marker implementing this will automatically
 // be run after the rest of a given schema node has been generated.
 // Markers that need to be run before any other markers can also
-// implement ApplyFirst, but this is discouraged and may change
-// in the future.
+// implement ApplyFirst or markers.ApplyPriority, but this is discouraged
+// and may change in the future.
```
camilamacedo86, line 29: Could we add here an example over how to use both?
Beyond that all fine 🥇 
Great work

dprotaso:
@camilamacedo86 done - take a look

pkg/crd/markers/doc.go
```diff
@@ -26,20 +26,32 @@ limitations under the License.
...
 //
+//  type MyCustomMarker string
+//
+//  func (m MyCustomMarker) ApplyPriority() ApplyPriority {
+//  	return ApplyPriorityFirst
+//  }
+//
+//  func (m MyCustomMarker) ApplyToSchema(schema *apiext.JSONSchemaProps) error {
+//    ...
+//  }
```
camilamacedo86, line 41: Could we change it for. something like

interface in combination with ApplyPriorityDefault and
// ApplyPriorityFirst constants. Following an example over how to use them:

camilamacedo86 (commented):
I am ok with but would be nice get thoughts from others like @alvaroaleman @varshaprasad96 

Otherwise, lgtm

pkg/crd/markers/doc.go
dprotaso: Done

dprotaso:
Hi - folks just following up

camilamacedo86 (commented):
It seems good or me. 🥇 
Let's see what others think about

/lgtm

dprotaso:
bump @vincepri @alvaroaleman

pkg/crd/markers/priority.go
```diff
@@ -0,0 +1,37 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package markers
+
+import "math"
+
+type ApplyPriority int64
```
vincepri, line 21: Add a comment on what this type does?

pkg/crd/markers/priority.go
```diff
@@ -0,0 +1,37 @@
...
+
+const (
+	// ApplyPriorityDefault is the default priority for markers
+	// that don't implement ApplyPriorityMarker
+	ApplyPriorityDefault ApplyPriority = math.MaxInt64 >> 2
+
+	// ApplyPriorityFirst is the priority value assigned to markers
+	// that implement the ApplyFirst() method
+	ApplyPriorityFirst ApplyPriority = ApplyPriorityDefault - 10
```
vincepri, line 30: This is a bit hard to understand imo, could we consider an alternative where we define constants that are clear on where they're used.

For example:
- `ApplyPriorityFirst` = 1
- `ApplyPriorityTypes` = 5
-  `ApplyPriorityDefault` = 10

And so on?

dprotaso:
Seems like prow is broken?

/retest

dprotaso:
Prow started working again

@vincepri and/or @alvaroaleman PTAL

dprotaso:
@vincepri and/or @alvaroaleman PTAL

dprotaso:
@vincepri @alvaroaleman can you take a look please?

pkg/crd/schema.go
```diff
@@ -126,39 +127,44 @@ func infoToSchema(ctx *schemaContext) *apiext.JSONSchemaProps {
 
 // applyMarkers applies schema markers to the given schema, respecting "apply first" markers.
 func applyMarkers(ctx *schemaContext, markerSet markers.MarkerValues, props *apiext.JSONSchemaProps, node ast.Node) {
+	markers := make([]SchemaMarker, 0, len(markerSet))
+
 	// apply "apply first" markers first...
```
sbueringer, line 132:
```suggestion
```

Seems oudated

pkg/crd/schema.go
```diff
@@ -126,39 +127,44 @@ func infoToSchema(ctx *schemaContext) *apiext.JSONSchemaProps {
 
 // applyMarkers applies schema markers to the given schema, respecting "apply first" markers.
```
sbueringer, line 128: Let's please update this comment to match the new implementation

pkg/crd/schema_test.go
```diff
@@ -110,3 +111,58 @@ func Test_Schema_MapOfStringToArrayOfFloat32(t *testing.T) {
...
+			&testapplyFirstMarker{callback: func() {
+				invocations = append(invocations, "applyFirst")
+			}},
+			&testPriorityMarker{priority: crdmarkers.ApplyPriorityDefault, callback: func() {
+				invocations = append(invocations, "default")
+
+			}},
+		}}, props, nil)
+
+	g.Expect(invocations).To(gomega.Equal([]string{"0", "1", "applyFirst", "default"}))
```
sbueringer, line 143: I think "1" and "applyFirst" are not guaranteed to be in that order. Both end up with priority 1 and right now we are only using sort.Slice.
```go
// The sort is not guaranteed to be stable: equal elements
// may be reversed from their original order.
// For a stable sort, use SliceStable.
```

I assume we don't want to guarantee an order for the same priority. So using sort.Slice should be fine. I just wonder if we should use distinct priorities for the test here

pkg/crd/schema_test.go
```diff
@@ -110,3 +111,58 @@ func Test_Schema_MapOfStringToArrayOfFloat32(t *testing.T) {
...
+				},
+			},
+			&testPriorityMarker{priority: 1, callback: func() {
+				invocations = append(invocations, "1")
+
+			}},
+			&testapplyFirstMarker{callback: func() {
+				invocations = append(invocations, "applyFirst")
+			}},
+			&testPriorityMarker{priority: crdmarkers.ApplyPriorityDefault, callback: func() {
```
sbueringer, line 137: Would it make sense to use a different priority here and add an additional marker that fallsback to the default?

pkg/crd/markers/priority.go
```diff
@@ -0,0 +1,37 @@
...
+	// that don't implement ApplyPriorityMarker
+	ApplyPriorityDefault ApplyPriority = 10
+
+	// ApplyPriorityFirst is the priority value assigned to markers
+	// that implement the ApplyFirst() method
+	ApplyPriorityFirst ApplyPriority = 1
+)
+
+// ApplyPriorityMarker designates the order markers should be applied.
+// Lower priority indicates it should be applied first
```
sbueringer, line 34: Should we mention that this only applies to Validation markers?

sbueringer (commented):
Thx looks good. Just a few minor findings otherwise good to go from my side

pkg/crd/schema_test.go
dprotaso: I added a new marker to test the default priority. Likewise I updated the test so all the priorities are unique.

pkg/crd/schema_test.go
dprotaso: done

pkg/crd/markers/priority.go
dprotaso: ApplyPriority can apply to any marker - not just validation

sbueringer: I was looking at the doc and there we distinguish between "Validation Markers" (implementing crd.SchemaMarker) and "CRD Markers" (implementing crd.SpecMarker). I only saw ApplyPriorityMarker being handled for SchemaMarker.

I might be missing something, but looks like it's not implemented for SpecMarker

pkg/crd/markers/doc.go
```diff
@@ -19,27 +19,40 @@ limitations under the License.
...
+//	}
 //
 // All validation markers start with "+kubebuilder:validation", and
 // have the same name as their type name.
 //
-// CRD Markers
+// # CRD Markers
 //
 // Markers that modify anything in the CRD itself *except* for the schema
 // implement ApplyToCRD (crd.CRDMarker).  They are expected to detect whether
```
sbueringer, line 50:
```suggestion
// implement ApplyToCRD (crd.SpecMarker).  They are expected to detect whether
```

Just saw when looking at this again that this comment is wrong. Would be nice if you can fix it :)

pkg/crd/markers/priority.go
dprotaso: Good catch - yeah it's not in the spec marker - I'll update the comment

sbueringer:
Thank you!

/lgtm

/assign @alvaroaleman @vincepri

sbueringer:
/approve

Thank you very much!

Issue #641, opened 2021-11-15
tag for reinvocationPolicy


mythi:
[Admission webhooks' documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy) says `reinvocationPolicy` can be set to `Never` or `IfNeeded`.

My webhook needs `IfNeeded` but `controller-gen` resets the `MutatingWebhookConfiguration` YAML to `reinvocationPolicy: Never`. 

Can we have a tag to control this?

FillZpp:
> Can we have a tag to control this?

@mythi  Of course, it should be a new field in https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/webhook/parser.go#L60 . Do you have time doing this? Or I can help you on this.

mythi:
> > Can we have a tag to control this?
> 
> @mythi Of course, it should be a new field in https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/webhook/parser.go#L60 . Do you have time doing this? Or I can help you on this.

I need to prioritize other topics right now but I could indeed contribute to this.

mythi:
@FillZpp thanks for giving the pointer where to start! I have the PR ready for review.

PR #709, opened 2022-07-26
:sparkles: pkg/webhook: add support for reinvocationPolicy marker

mythi:
Mutating webhooks using controller-gen markers for generating
MutatingWebhookConfigration cannot control reinvocationPolicy.
This field must always be added manually afterwards should the
webhook need to set its value.

Add support for reinvocationPolicy marker for Mutating Webhooks.

Fixes: #641

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -19,7 +19,7 @@ limitations under the License.
 //
 // The markers take the form:
 //
-//  +kubebuilder:webhook:webhookVersions=<[]string>,failurePolicy=<string>,matchPolicy=<string>,groups=<[]string>,resources=<[]string>,verbs=<[]string>,versions=<[]string>,name=<string>,path=<string>,mutating=<bool>,sideEffects=<string>,admissionReviewVersions=<[]string>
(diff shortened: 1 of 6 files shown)


pkg/webhook/parser.go
```diff
@@ -263,6 +272,22 @@ func (c Config) sideEffects() *admissionregv1.SideEffectClass {
...
 
+// reinvocationPolicy returns the reinvocationPolicy config for a webhook.
+func (c Config) reinvocationPolicy() *admissionregv1.ReinvocationPolicyType {
+	var reinvocationPolicy admissionregv1.ReinvocationPolicyType
+	switch strings.ToLower(c.ReinvocationPolicy) {
+	case strings.ToLower(string(admissionregv1.NeverReinvocationPolicy)):
+		reinvocationPolicy = admissionregv1.NeverReinvocationPolicy
+	case strings.ToLower(string(admissionregv1.IfNeededReinvocationPolicy)):
+		reinvocationPolicy = admissionregv1.IfNeededReinvocationPolicy
+	case "":
```
FillZpp, line 283: Is this necessary since we already have the `default`?

mythi: Good question. I blindly followed `sideEffects` here but `matchPolicy` has it too but returns error for `default`. I'll take a closer look.

mythi: Dropped `case "":` is it did not have any effect.

alvaroaleman (approved):

alvaroaleman:
thanks!

PR #712, opened 2022-08-16
📖 Update controller-gen CLI documentation for better info regards paths argument  

padlar:
## Description

Add paths description on how to specify multiple paths.

## Motivation

<TODO: Would be nice add here the problem faced and from where you got the info to update this one>


camilamacedo86 (approved):
It is a small one and shows good to go.
Thank you for the contribution 🥇 

/approoved

Issue #718, opened 2022-08-25
Regression from v0.9.0+ errors with `stderr: go: updates to go.mod needed`


pjbgf:
Command `controller-gen schemapatch:manifests="./" paths="./..."` breaks after upgrading to `v0.9.0+`.
Error:
```
err: exit status 1: stderr: go: updates to go.mod needed; to update it:
        go mod tidy
```

Note that `go mod tidy` does not fix the problem.
Only downgrading to `v0.8.0` or `v0.7.2` fixes the issue (as per https://github.com/fluxcd/pkg/pull/324).

Full error by running [fluxcd/pkg](https://github.com/fluxcd/pkg)'s `make test-oci`.
```
$ make test-oci

go: creating new go.mod: module tmp
Downloading sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0
go: downloading sigs.k8s.io/controller-tools v0.9.0
mkdir -p /home/levi/go/src/github.com/fluxcd/pkg/testbin
/home/levi/go/src/github.com/fluxcd/pkg/bin/setup-envtest use latest --arch=amd64 --bin-dir=/home/levi/go/src/github.com/fluxcd/pkg/testbin
Version: 1.24.2
OS/Arch: linux/amd64
md5: ZViX8vb02m8ULK0fkiNNGw==
Path: /home/levi/go/src/github.com/fluxcd/pkg/testbin/k8s/1.24.2-linux-amd64
cd oci; go mod tidy -compat=1.17
# Run schemapatch to validate all the kubebuilder markers before generation
cd oci; CGO_ENABLED=0 /home/levi/go/src/github.com/fluxcd/pkg/bin/controller-gen schemapatch:manifests="./" paths="./..."
Error: err: exit status 1: stderr: go: updates to go.mod needed; to update it:
        go mod tidy

Usage:
  controller-gen [flags]

Examples:
        # Generate RBAC manifests and crds for all types under apis/,
        # outputting crds to /tmp/crds and everything else to stdout
        controller-gen rbac:roleName=<role name> crd paths=./apis/... output:crd:dir=/tmp/crds output:stdout

        # Generate deepcopy/runtime.Object implementations for a particular file
        controller-gen object paths=./apis/v1beta1/some_types.go

        # Generate OpenAPI v3 schemas for API packages and merge them into existing CRD manifests
        controller-gen schemapatch:manifests=./manifests output:dir=./manifests paths=./pkg/apis/... 

[...]

PR #725, opened 2022-09-18
:bug: return a bit more verbosy error messages for go-package errors

AlmogBaku:
When having sub-packages, and one of them is out of sync, we might get an error such as:
```
err: exit status 1: stderr: go: updates to go.mod needed; to update it:
        go mod tidy
```

Although this error is true, it was a nightmare to debug it and figure out that it was related to a sub-package.

This PR introduces a bit more verbosy error message that can help you to debug the issue:
```
Error: error in /Users/AlmogBaku/projects/myproject/subpackage: err: exit status 1: stderr: go: updates to go.mod needed; to update it:
        go mod tidy
```

### DIFF
--- a/pkg/loader/loader.go
+++ b/pkg/loader/loader.go
@@ -404,7 +404,11 @@ func LoadRootsWithConfig(cfg *packages.Config, roots ...string) ([]*Package, err
 	loadPackages := func(roots ...string) ([]*Package, error) {
 		rawPkgs, err := packages.Load(l.cfg, roots...)
 		if err != nil {
-			return nil, fmt.Errorf("error in %s: %w", l.cfg.Dir, err)
+			loadRoot := l.cfg.Dir
+			if l.cfg.Dir == "" {
+				loadRoot, _ = os.Getwd()
+			}
+			return nil, fmt.Errorf("load packages in root %q: %w", loadRoot, err)
 		}
 		var pkgs []*Package
 		for _, rp := range rawPkgs {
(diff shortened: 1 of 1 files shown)


AlmogBaku:
might be related to #718

camilamacedo86 (approved):
/lgtm

pkg/loader/loader.go
```diff
@@ -404,7 +404,7 @@ func LoadRootsWithConfig(cfg *packages.Config, roots ...string) ([]*Package, err
 	loadPackages := func(roots ...string) ([]*Package, error) {
 		rawPkgs, err := packages.Load(l.cfg, roots...)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error in %s: %w", l.cfg.Dir, err)
```
joelanford, line 407: `l.cfg.Dir` is allowed to be empty string (in which case the loader loads from the current working directory). I'm not sure that's possible in c-t's use of this package as it is currently written, but either way, I think we should account for this possibility if we're going to print its value.

Also, maybe even more context would have been helpful here?
```suggestion
			loadRoot := l.cfg.Dir
			if l.cfg.Dir == "" {
				loadRoot, _ = os.Getwd()
			}
			return nil, fmt.Errorf("load packages in root %q: %w", loadRoot, err)
```

deepankarm:
Any updates on when this PR can be merged? It indeed is a nightmare to debug.

camilamacedo86:
The requested changes by @joelanford are addressed so we can : 

/lgtm

PR #736, opened 2022-11-22
🌱 : dependabot addition

kannon92:
From a request on controller-runtime slack, there was an ask to add dependabot to both controller-runtime (https://github.com/kubernetes-sigs/controller-runtime/pull/2050) and controller-tools.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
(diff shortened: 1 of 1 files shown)


camilamacedo86 (approved):
/ok-to-test

/approved

/lgtm 

Thank you 🥇

camilamacedo86:
PS. I just udate the title to use the right emoji so that this change does not end up in the release notes which is generated automatically.

kannon92:
> PS. I just udate the title to use the right emoji so that this change does not end up in the release notes which is generated automatically.

BTW, you may want to update that in your PULL_REQUEST template for this repo.  

I picked that one because the following code didn't let me know that seedling was an option:

```
<!-- please add a icon to the title of this PR (see VERSIONING.md), and delete this line and similar ones -->
<!-- the icon will be either ⚠ (:warning:, major), ✨ (:sparkles:, minor), 🐛 (:bug:, patch), 📖 (:book:, docs), or 🏃 (:running:, other) -->

<!-- What does this do, and why do we need it? -->
```
Do you want me to add the seedling to this for future contributors?

camilamacedo86:
moving forward with this one since it is just for us to begin to be able to get the deps automatically.

PR #754, opened 2022-12-20
:bug:  Revert Fix the generation of listType=set

vincepri:
Reverts kubernetes-sigs/controller-tools#753

Which could have side-effects with existing CRDs and should be reverted, upstream fix is incoming for the same issue


alvaroaleman (approved):

jdef:
is there a link somewhere to a list of side effects?

PR #758, opened 2023-01-05
Create some simple golden tests for CRD generation

justinsb:
Adds some test coverage to CRDs with HeaderFile generation.


pkg/crd/testdata/golden/with-header-file/bar.example.com_foos.yaml
```diff
@@ -0,0 +1,66 @@
+/*
```
camilamacedo86, line 1: It is an yaml file. So, would not the comments need to be with `#` as prefix?

justinsb: lol - good point!  I'll fix :-)

justinsb: Fixed - thanks for spotting this!

camilamacedo86 (approved):
All fine. I just added the right mojo in the title.

/lgtm

vincepri (approved):
/approve

camilamacedo86:
Hi @justinsb,

I know that it is open for a while. Note that it is no longer passing in the tests.
Could you please check it out?

camilamacedo86:
/hold 

Until fixing the test, which is now longer passing

camilamacedo86:
@dependabot rebase

PR #765, opened 2023-01-19
add support for oneOf

harveyxia:
Cargo culted from github.com/kubernetes-sigs/controller-tools/pull/298 (with some vetting), it looks correct.

The only concern is described in point C of [this comment](https://github.com/kubernetes-sigs/controller-tools/pull/298#issuecomment-518826515), but I don't know if a non-structural schema can actually be generated with this logic.


harveyxia:
(apologies, I opened this PR against the wrong repo)

PR #774, opened 2023-02-03
✨ Add support for example marker

sbueringer:
Adds an example marker.


sbueringer:
Assuming the change is fine in general. Would be great if we can cherry-pick this into v0.11.x

vincepri (approved):
/approve
/lgtm

PR #792, opened 2023-04-08
:bug: Change the type of `markers.Collector.byPackage`'s key from `string` to `*loader.Package`

### LINKED ISSUE #783: Missing `x-kubernetes-map-type` when multiple `paths=` are specified
### What happened

When I executed `controller-gen` with multiple `paths=` option specified, several `x-kubernetes-map-type` was missing.

### How to reproduce

For example, I cloned https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/ to generate CRDs from `apis/` in that repository.

And I prepared script to check some field in generated CRD.

```
$ cat << "EOF" > check_script.sh
heredoc> python -c 'import sys, json, yaml; print(json.dumps(yaml.safe_load(sys.stdin)));' < /tmp/test/infrastructure.cluster.x-k8s.io_vspherevms.yaml | jq '.spec.versions[] | [.name, .schema.openAPIV3Schema.properties.spec.properties.bootstrapRef."x-kubernetes-map-type"] | @tsv' -r
EOF

$ chmod +x check_script.sh
```

Running controller-gen against every single API version generated CRD including `x-kubernetes-map-type` correctly.

```
$ controller-gen paths=./apis/v1alpha3 crd:crdVersions=v1 output:crd:dir=/tmp/test && ./check_script.sh 
v1alpha3	atomic

$ controller-gen paths=./apis/v1alpha4 crd:crdVersions=v1 output:crd:dir=/tmp/test && ./check_script.sh 
v1alpha4	atomic

$ controller-gen paths=./apis/v1beta1 crd:crdVersions=v1 output:crd:dir=/tmp/test && ./check_script.sh 
v1beta1	atomic   
```

And `paths=./apis/...` also worked well.

```
$ controller-gen paths=./apis/... crd:crdVersions=v1 output:crd:dir=/tmp/test && ./check_script.sh 
v1alpha3	atomic
v1alpha4	atomic
v1beta1	atomic
```

But specifying `paths` per API versions generated CRDs with several `x-kubernetes-map-type` field missing, and the version whose `x-kubernetes-map-type` disappeared would change when running many times.

```
$ controller-gen paths=./apis/v1alpha3 paths=./apis/v1alpha4 paths=./apis/v1beta1 crd:crdVersions=v1 output:crd:dir=/tmp/test && ./check_script.sh                                                                     
v1alpha3	atomic
v1alpha4	
v1beta1

$ controller-gen paths=./apis/v1alpha3 paths=./apis/v1alpha4 paths=./apis/v1beta1 crd:crdVersions=v1 output:crd:dir=/tmp/test && ./check_script.sh                                                                     
v1alpha3	
v1alpha4	atomic
v1beta1	
```

### Environment

```
$ controller-gen --version
Version: v0.11.3
```

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/commit/1bae0f90efcceaff2c4aa51e04dc560f9932a07d

### Expected behavior

Both specifying paths individually (such as `paths=./apis/v1alpha3 paths=./apis/v1alpha4 paths=./apis/v1beta1`) and with wildcard (such as `paths=./apis/...`) generates the same CRD.

### Related Issue

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1757

ntoofu:
Fix #783.

As I described in https://github.com/kubernetes-sigs/controller-tools/issues/783#issuecomment-1487356924 , this issue is caused by using `Package.ID` as a key for cache of markers. This PR adds a testcase to reproduce the issue and fixes it by using `Package` itself instead.

### DIFF
--- a/pkg/markers/collect.go
+++ b/pkg/markers/collect.go
@@ -31,7 +31,7 @@ import (
 type Collector struct {
 	*Registry
 
-	byPackage map[string]map[ast.Node]MarkerValues
+	byPackage map[*loader.Package]map[ast.Node]MarkerValues
 	mu        sync.Mutex
 }
 
@@ -53,7 +53,7 @@ func (c *Collector) init() {
 		c.Registry = &Registry{}
 	}
 	if c.byPackage == nil {
-		c.byPackage = make(map[string]map[ast.Node]MarkerValues)
+		c.byPackage = make(map[*loader.Package]map[ast.Node]MarkerValues)
 	}
 }
 
@@ -75,7 +75,7 @@ func (c *Collector) init() {
 func (c *Collector) MarkersInPackage(pkg *loader.Package) (map[ast.Node]MarkerValues, error) {
 	c.mu.Lock()
 	c.init()
-	if markers, exist := c.byPackage[pkg.ID]; exist {
+	if markers, exist := c.byPackage[pkg]; exist {
 		c.mu.Unlock()
 		return markers, nil
 	}
@@ -91,8 +91,7 @@ func (c *Collector) MarkersInPackage(pkg *loader.Package) (map[ast.Node]MarkerVa
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	c.byPackage[pkg.ID] = markers
-
+	c.byPackage[pkg] = markers
 	return markers, nil
 }
 
--- a/pkg/crd/parser_integration_test.go
+++ b/pkg/crd/parser_integration_test.go
@@ -266,4 +266,52 @@ var _ = Describe("CRD Generation From Parsing to CustomResourceDefinition", func
 		By("checking that no errors occurred along the way (expect for type errors)")
 		Expect(packageErrors(cronJobPkg, packages.TypeError)).NotTo(HaveOccurred())
 	})
+
+	It("should generate markers properly among several package versions", func() {
+		By("switching into testdata to appease go modules")
+		cwd, err := os.Getwd()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(os.Chdir("./testdata/multiple_versions")).To(Succeed())
+		defer func() { Expect(os.Chdir(cwd)).To(Succeed()) }()
+
+		By("loading the roots")
+		pkgs, err := loader.LoadRoots("./v1beta1", "./v1beta2")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(pkgs).To(HaveLen(2))
+
+		By("setting up the parser")
+		reg := &markers.Registry{}
+		Expect(crdmarkers.Register(reg)).To(Succeed())
+		parser := &crd.Parser{
+			Collector: &markers.Collector{Registry: reg},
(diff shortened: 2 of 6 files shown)


randomvariable:
cc @sbueringer fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1928

sbueringer:
Thx for the extensive explanation here: https://github.com/kubernetes-sigs/controller-tools/issues/783#issuecomment-1487356924

/lgtm

pkg/markers/collect.go
```diff
@@ -31,7 +31,7 @@ import (
 type Collector struct {
 	*Registry
 
-	byPackage map[string]map[ast.Node]MarkerValues
+	byPackage map[*loader.Package]map[ast.Node]MarkerValues
```
vincepri, line 34: This stores the package based on the pointer's value, is there any other way that doesn't rely on where the memory is allocated?

randomvariable: There is a unique ID in https://pkg.go.dev/golang.org/x/tools/go/packages#Package

ntoofu: Thank you for your reviewing!

There is possibly another way, but it may be complicated.

`byPackage` has two keys, package ID (`string`) and node (`ast.Node`). Package ID is not a pointer's value, while node is an interface and can be a struct containing pointers. That inconsistency is the root cause of this issue.

In the problematic situation, exactly the same package is loaded several times. Among each load, the package ID is the same but nodes are different objects. [Parsing markers is skipped](https://github.com/kubernetes-sigs/controller-tools/blob/73391d1c6f3edfed27bbd82c8cc7bfb05f635bad/pkg/markers/collect.go#L78) when the same package ID is already processed. Therefore, `byPackage`'s node keys are created only at the first time. As a result, access `byPackage` using node as a key fails from the second time.

To avoid to use pointer's value as `byPackage`'s key, I think we have to do one of the following.

* Stop using `ast.Node` as a key (but I have no idea about the alternative)
* Do not skip parsing markers even if a package with the same ID is already parsed
* Stop loading the same package multiple times (but I have no idea why the multiple load occurs)

sbueringer: Yup agree. The problem is definitely that we cache the markers (with pointers) by package.ID and then try to use them for another package (which has different ast nodes).

sbueringer: I think it should be not worse to use *loader.Package as a map key then it already is to use the ast.Node interface (aka e.g. *ast.File) as a map key one layer deeper)

vincepri: That's fair, it's a bit triggering though in review 😄

sbueringer: Yeah 100% :)

sbueringer:
Oh don't need the fix in release-0.11 (although no objection to merge there).

Just missed that we have a v0.12 release because we don't have a release branch for v0.12 yet

sbueringer:
I created a release-0.12 branch based on v0.12.0

/cherry-pick release-0.12

PR #800, opened 2023-04-21
🐛 remove metadata.creationTimestamp when generating CRDs, RBAC, and webhooks

### LINKED ISSUE #402: Remove creationTimestamp from generated CRD
controller-gen currently adds the metadata field `creationTimestamp: null`.  I think this field should not be included in the CRD yaml since this value should be added by the server.

joelanford:
Closes #402 

This issue has been open for awhile, but seemingly not _that_ big of a deal until kustomize v5 was released with a bug that converts `null` to `"null"`. And _that_ causes manifests to fail to apply.

https://github.com/kubernetes-sigs/kustomize/issues/5031

We can stem the tide in controller-tools by removing `creationTimestamp: null` just before writing the generated manifests. This was always unsightly/superfluous anyways.


camilamacedo86 (approved):
/lgtm

PR #802, opened 2023-04-22
🐛 Fix marker parsing

inteon:
Markers that have number-like parts (like `0B7LDeoZta`) trigger parsing errors.
This causes `controller-gen` to fail (see https://github.com/cert-manager/cert-manager/issues/5955).
This PR updates the scanner, so it does not try to parse a string as numbers when we know it is a string instead of a number.


sbueringer:
Seems okay to me

/lgtm

vincepri (approved):
/approve
/lgtm

PR #804, opened 2023-04-26
⚠ Bump github.com/gobuffalo/flect from 0.3.0 to 1.0.2

sbueringer:
This PR bumps flect to 1.0.2

Please be aware that the pluarlize logic in flect changed

Links:
* https://github.com/kubernetes-sigs/controller-tools/pull/780#issuecomment-1523658715
* https://github.com/gobuffalo/flect/releases/tag/v1.0.0


pkg/crd/testdata/gen/zoo/zoo_types.go
```diff
@@ -33,6 +33,7 @@ type ZooSpec struct {
 }
 type ZooStatus struct{}
 
+// +kubebuilder:resource:path=zooes
```
sbueringer, line 36: Without overwriting the plural here the new flect version would generate the plural "zoo"

sbueringer:
/assign @vincepri @alvaroaleman 

Not sure if there is a better way to deal with the breaking change in flect.

pkg/crd/testdata/gen/zoo/zoo_types.go
vincepri: Isn't the plural `zoos`?

sbueringer: That is what flect changed. They are now generating zoos instead of zooes.

I could have either changed the test file or pinned the path. I picked pinning the path/plural.

I'll change the test file now instead

sbueringer:
@vincepri done

vincepri (approved):
/lgtm

PR #399, opened 2020-02-05
 🐛 Provide error when CRD has multiple versions marked for storage

theishshah:
This addresses the bug found in Issue #390 

In order to verify that only one storage version exists, I replaced the boolean flag with a counter and provide errors when the count is equal to 0 or greater than 1.

PR #820, opened 2023-05-22
✨ Return error when CRD has multiple versions marked for storage

lauchokyip:
Continue the work from https://github.com/kubernetes-sigs/controller-tools/pull/399, which will be beneficial for this kubebuilder issue https://github.com/kubernetes-sigs/kubebuilder/issues/2589 too

### DIFF
--- a/pkg/crd/spec.go
+++ b/pkg/crd/spec.go
@@ -142,6 +142,15 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind, maxDescLen *int) {
 	// Otherwise, crd.Spec.Version may point to different CRD versions across different runs.
 	sort.Slice(crd.Spec.Versions, func(i, j int) bool { return crd.Spec.Versions[i].Name < crd.Spec.Versions[j].Name })
 
+	for _, err := range ValidateStorageVersionAndServe(crd, groupKind) {
+		packages[0].AddError(err)
+	}
+
+	p.CustomResourceDefinitions[groupKind] = crd
+}
+
+func ValidateStorageVersionAndServe(crd apiext.CustomResourceDefinition, groupKind schema.GroupKind) []error {
+	errs := []error{}
 	// make sure we have *a* storage version
 	// (default it if we only have one, otherwise, bail)
 	if len(crd.Spec.Versions) == 1 {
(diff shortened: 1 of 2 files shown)


pkg/crd/spec.go
```diff
@@ -151,14 +160,18 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind, maxDescLen *int) {
...
+			}
 			hasStorage = true
-			break
 		}
 	}
 	if !hasStorage {
 		// just add the error to the first relevant package for this CRD,
 		// since there's no specific error location
-		packages[0].AddError(fmt.Errorf("CRD for %s has no storage version", groupKind))
+		errs = append(errs, fmt.Errorf("CRD for %s has no storage version", groupKind))
```
vincepri, line 174: why was this changed?

lauchokyip: @vincepri  I was following this PR 
https://github.com/kubernetes-sigs/controller-tools/pull/399/files#diff-05d11052b3234e4c8ce76c9362508cbd5c3f275a76ee2513240234ed879cb2b5L151-R163 

Should I stick with that approach or the one you suggested?

pkg/crd/spec.go
```diff
@@ -151,14 +160,18 @@ func (p *Parser) NeedCRDFor(groupKind schema.GroupKind, maxDescLen *int) {
 	hasStorage := false
 	for _, ver := range crd.Spec.Versions {
 		if ver.Storage {
+			// more than one storage, already set it true on previous iteration
+			if hasStorage {
+				errs = append(errs, fmt.Errorf("CRD for %s has more than one storage version", groupKind))
```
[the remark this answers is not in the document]
vincepri: sets.New[String] should do it?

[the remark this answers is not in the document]
lauchokyip: I think there is a misunderstanding here.

when you loop through different version, it would be something like `v1`, `v2`, `v3`, and if we put them all in the set, it would be something like `set{"v1,"v2", "v3"}`. **Storage Version** is a boolean value, making it a map would be `map{"v1": true,"v2": true, "v3": false}`, I don't think there is need to set or map. We would need to check if the boolean is triggered twice, that should be more than enough

pkg/crd/spec.go
lauchokyip: @vincepri  I wouldn't recommend to pass the package into the function because  it would complicate the unit test writing ( need to setup a package loader ) . I would keep the function loosely coupled with that

PR #839, opened 2023-08-01
:sparkles: Generate role binding based on generated role

### LINKED ISSUE #760: Role and Clusterrole binding are not generated correctly
**CONTROLLER_TOOLS_VERSION=v0.10.0**

command: 
`controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases`

**What I did:**
My controller needs to access both a cluster resource and a namespaced one so I use both annotations to autogenerate rbac rules. 

**What happened:**
When annotating the controller with both namespaced and cluster rbac rules both roles are created but, only one binding.

**What I would expect:**
Both bindings for the clusterrole and the role are generated

**How To Reproduce:**

1.  In a kubebuilder PROJECT annotate a controller with 

```
//clusterrole
// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;list;watch;update;

//role
// +kubebuilder:rbac:groups="",namespace=mynamespace,resources=secrets,verbs=get;list;watch;create;update;patch;delete
```
2. Run `make manifests` or `controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases`

3. Observe that the role.yaml file has both ClusterRole and Role correctly generated
```
---
kind: ClusterRole
metadata:
  creationTimestamp: null
  name: manager-role
rules:
- apiGroups:
  - admissionregistration.k8s.io
  resources:
  - mutatingwebhookconfigurations
  verbs:
  - get
  - list
  - watch

---
kind: Role
metadata:
  creationTimestamp: null
  name: manager-role
  namespace: mynamespace
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
```

4. Check that role_binding.yaml  has only the ClusterRole bindings

```
kind: ClusterRoleBinding
metadata:
  labels:
  name: manager-rolebinding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: manager-role
subjects:
- kind: ServiceAccount
  name: cert-manager
  namespace: system
```
**Temporary fix**
I could manually add the other binding and list it in the kustomization.yaml, but I would expect the tool to generate both in the role_binding.yaml

Anddd7:
Generate `RoleBinding` or `ClusterRoleBinding` based on the generated roles.

In [kubebuilder.role_binding.go#L44](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/rbac/role_binding.go#L44), it initilized the rolebinding as a static template which only contains ClusterRoleBinding. So we need to edit it manually and it's not auto-updated if you add `Role` instead of `ClusterRole`.

fix #760

(BTW, in `kubebuilder.role_binding.go#L44`, there are several labels generated in template. To keep consistent with other resources, do we need to add it for role/role_binding as well?
(One solution is using `kustomize#LabelTransformer` or `kustomize#patches`, which only changes the yaml templates


Anddd7:
/title Create role binding generator based on generated role

Anddd7:
/retitle :sparkles: Generate role binding based on generated role

pkg/rbac/parser.go
```diff
@@ -259,13 +259,107 @@ func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{
...
+
+		if role, ok := role.(rbacv1.Role); ok {
+			objs = append(objs, rbacv1.RoleBinding{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "RoleBinding",
+					APIVersion: rbacv1.SchemeGroupVersion.String(),
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("%sbinding", roleName),
+					Namespace: role.Namespace,
```
aleskandro, line 300: Hi, I just got into the issue you refer to in this PR. Is the kustomize namespace replacement considered here? If I manually generate a role like this in my case, the namespace would be replaced with the one set in the main kustomization.yaml file.

Anddd7: This 'controller-tools' is mainly used to parse the annotations of kubebuilder and generate the manifest, no kustomization.

Ref the kubebuilder, it do have kustomization, https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/rbac/kustomization.go.

If we want to overwrite the namespace, i prefer to use an additional config of https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#_namespacetransformer.

aleskandro: Are you stating that you would change how the namespace is set by the projects consuming controller-runtime (kube-builder, operator-sdk, ...)  as a follow-up that uses something like `unsetOnly: true` and `setRoleBindingSubjects: allServiceAccounts`?

Issue #837, opened 2023-07-26
controller-gen fails to handle individual file in paths 


mmingdai:
controller-gen treats the file as a package, which causes error. This command works in v0.7.0. I think it's due to https://github.com/kubernetes-sigs/controller-tools/pull/663 
Specifically this part
https://github.com/kubernetes-sigs/controller-tools/pull/663/files#diff-19effe657639f31c4b607076be411e82add5f6cc176fc7c26555020ff9663cbdR562

mmingdai:
@akutz

camilamacedo86:
PR to fix thise one: https://github.com/kubernetes-sigs/controller-tools/pull/1389

/assign

JoelSpeed:
What is the motivation here? What benefit does loading a single file have over loading the entire package?

camilamacedo86:
Hi @JoelSpeed 

The motivation here is to solve the scenario:

The loader treated all filesystem paths as directories, causing errors when individual files were specified:

```
# This failed with "not a directory" error
controller-gen object paths=./apis/v1beta1/some_types.go
                                                                              
Error: load packages in root ".../some_types.go": err: fork/exec
/usr/local/go/bin/go: not a directory                                         
```

sbueringer:
I think Joel's question is what is the use case for only loading single files

camilamacedo86:
Hi @sbueringer @JoelSpeed,

> I think Joel's question is what is the use case for only loading single files

Thanks for raising this.

Package-level generation should remain the default, but file-level inputs are useful when the generation boundary is smaller than the Go package.

In complex repositories, users may need to:

* choose exactly which files controller-tools processes;
* generate files in different pipeline stages;
* send outputs to different directories;
* avoid processing unrelated files from the same package.

We should not assume every Go file in a directory belongs to the same generation workflow.

A real example is `operator-controller`. It keeps several CRDs in the same package but distributes them separately. Its Makefile documents the current workaround: generate everything together, then move and filter the outputs.

https://github.com/operator-framework/operator-controller/blob/main/Makefile#L162-L206

PR #663 is another example of why flexible filesystem inputs matter. It added support for repositories with nested Go modules because real generation boundaries do not always match a single package or module.

**Other cases include per-controller RBAC, separate webhook deployments, feature-gated APIs, custom generators, and focused test fixtures.**

This would not replace package-based generation. It would simply allow more precise generation boundaries when projects need them.

sbueringer:
[...]

PR #864, opened 2023-11-30
🐛 Fix controller-tools doesn't support single files as input

yongxiu:
Fix https://github.com/kubernetes-sigs/controller-tools/issues/837

Below command
```bash
go run ../vendor/sigs.k8s.io/controller-tools/cmd/controller-gen webhook paths=./apis/cluster_webhook.go output:webhook:artifacts:config=./webhook/
```
Used to work, but didn't work any more since v0.9.0, which is caused by https://github.com/kubernetes-sigs/controller-tools/pull/663. There is a assumption `we want the base part of the path to be either "..." or "."`, this is untrue if user specified a specific file, I added a logic to exclude golang file in this case.

Another logic is `uniquePkgIDs`, as from comment `there is not harm that comes from this per se, but it makes testing easier when a known number of modules can be asserted`, this is fine however if there are two files coming from same package, second file will be ignored, thus I have to move this logic to test file.


rmohr (commented):
@yongxiu would you mind adding some context on how you are solving this to the PR description and/or commit?

yongxiu:
> @yongxiu would you mind adding some context on how you are solving this to the PR description and/or commit?

Done

vincepri:
`paths` is a list of directories, not files

rmohr (commented):
/lgtm

lgtom to me, but requires a project owner to have a look :)

rmohr (commented):
/remove-lgtm

based on:

> paths is a list of directories, not files

waiting for the out come of the discussion.

yongxiu:
> Why not specify the entire folder even if there is a single file in it?

There could be multiple files.

```
├── apis
│   ├── cluster_webhook.go
│   ├── machine_webhook.go
│   ├── typeBCluster_webhook.go
│   ├── typeBMachine_webhook.go
```

In above case, I want to generate webhook for cluster_webhook and machine_webhook in output directory A, while typeBCluster_webhook and typeBMachine_webhook to output directory B.

```
go run ../vendor/sigs.k8s.io/controller-tools/cmd/controller-gen webhook paths=./apis/cluster_webhook.go paths=./apis/machine_webhook.go  output:webhook:artifacts:config=./webhook/normal/
go run ../vendor/sigs.k8s.io/controller-tools/cmd/controller-gen webhook paths=./apis/typeBCluster_webhook.go paths=./apis/typeBMachine_webhook.go output:webhook:artifacts:config=./webhook/special/
```

> paths is a list of directories, not files

While, at least the command line example is using single file as well: https://github.com/kubernetes-sigs/controller-tools/blob/881ffb4682cb5882f5764aca5a56fe9865bc9ed6/cmd/controller-gen/main.go#L139

rmohr:
@vincepri being able to separate output directories like described above sounds useful and it seemed to work before. Is there a clear idea if that should only be dirs, or is it more an interpretation of the word `path` that it should be a directory?

joelanford:
@yongxiu, can you add a test that covers using a single file within a directory? The test should probably cover:
1. Generation does not occur for types in other files in the directory.
2. If the file that we're generating for references identifiers (types, functions, variables, etc.) in this package, but in another file in this package, generation still works and includes those referenced items as necessary.

My hunch is that the reason that the current code doesn't accept individual files is because the Go AST parser operates on package directories, which makes handling my item (2) above trivial.

yongxiu:
Added webhook integration tests as requested, please take another look, thanks. @joelanford

yongxiu:
@joelanford, added tests for CRDs and rbac as well

For the case you mentioned, I have this case in the test, it needs to pass in both files, otherwise it will only shows `type: object` for that field without detail fields, and if we run controller-tool cmd, it will give error:
```
$ go run cmd/controller-gen/main.go  crd paths=pkg/crd/testdata/multiple_files/file_two.go output:dir=pkg/crd/testdata/
multiple_files/
controller-tools/pkg/crd/testdata/multiple_files/file_two.go:31:7: unknown type TwoResourceSpec
```

To clarify, I also tested controller-tools v0.8.0, it give the same error.

joelanford:
Not suggesting you go and do this before we see what other maintainers think, but I wonder if we could:
- accept a single file as the path
- load the entire package directory that contains the file
- detect that are defined in the specified file and only generate based on those types

Alternately, another potential option is leaving `paths` supporting only package directories, but then add other option that lets a caller pass in a specific list of types they want to generate for?

yongxiu:
IMHO, detecting the types or add one more input of types is over kill. This feature has been broken for 5 releases, which is 1.5 years, and there is only one bug: https://github.com/kubernetes-sigs/controller-tools/issues/837 for this. I don't see much value of adding types, accepting multiple files which is changing back to 0.8.0 behavior is good enough for most of developers.

yongxiu:
I wonder how can we move this forward? Are we waiting for any other stakeholders?

yongxiu:
Friendly ping!

airycanon:
@yongxiu @joelanford Hi,  Is there any progress on this PR? 
My project relies on the single files input, and to use it, I have to keep my Go version at 1.21, as the old version that includes this feature causes a panic when running on Go 1.22.

rmohr:
@joelanford we would still be interested in that change. Would you have some recommendations on how to move this forward?

PR #517, opened 2020-11-11
✨ Trim leading/trailing whitespace on lines in documentation

Porges:
This is important for input files that use `/*…*/`-style comments,
where `go/ast` preserves the leading and trailing (and inner) whitespace
and this ends up in the generated CRD schema descriptions, where it looks simply silly.

PR #870, opened 2023-12-20
🐛 crd: Respect multiline comments at godocs 

qinqon:
Sometimes at type has examples about how to use it embedding something like a yaml on it, yamls should not be truncated and they are white space sensitive. This change keep the new lines and also remove the white space trimming only on /* comments.

Golang struct example with problematic comment.

```golang
// DnsState  DNS resolver state. Example partial yaml output of [NetworkState] with
// static DNS config:                                                           
// ```yaml                                                                      
// ---                                                                          
// dns-resolver:                                                                
//                                                                              
//  running:                                                                    
//     server:                                                                  
//     - 2001:db8:1::250                                                        
//     - 192.0.2.250                                                            
//     search:                                                                  
//     - example.org                                                            
//     - example.net                                                            
//  config:                                                                     
//     search:                                                                  
//     - example.org                                                            
//     - example.net                                                            
//     server:                                                                  
//     - 2001:db8:1::250                                                        
//     - 192.0.2.250                                                            
//     options:                                                                 
//     - trust-ad                                                               
//     - rotate                                                                 
//                                                                              
// ```                                                                          
// To purge all static DNS configuration:                                       
// ```yml                                                                       
// ---                                                                          
// dns-resolver:                                                                
//                                                                              
//  config: {}                                                                  
//                                                                              
// ```                                                                          
// +k8s:deepcopy-gen=true                                                       
type DnsState struct {                                                          
...
}
```

Part of the probem was introduced at https://github.com/kubernetes-sigs/controller-tools/pull/517

### DIFF
--- a/pkg/crd/testdata/gen/bar.example.com_foos.yaml
+++ b/pkg/crd/testdata/gen/bar.example.com_foos.yaml
@@ -19,14 +19,19 @@ spec:
       openAPIV3Schema:
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: |-
+              APIVersion defines the versioned schema of this representation of an object.
+              Servers should convert recognized schemas to the latest internal value, and
+              may reject unrecognized values.
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this
-              object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: |-
+              Kind is a string value representing the REST resource this object represents.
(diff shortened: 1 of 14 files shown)


varshaprasad96 (commented):
This change looks good to me, but it would break a lot of existing docs. @qinqon Some of the integration tests are failing due to formating changes, could you please fix them? (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-tools/870/pull-controller-tools-test-master/1737421873815228416#)

cc: @alvaroaleman any thoughts on this? Making this change optional may be an overkill. Having this introduced in a new release is still fine?

qinqon:
> This change looks good to me, but it would break a lot of existing docs. @qinqon Some of the integration tests are failing due to formating changes, could you please fix them? (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-tools/870/pull-controller-tools-test-master/1737421873815228416#)
> 
> cc: @alvaroaleman any thoughts on this? Making this change optional may be an overkill. Having this introduced in a new release is still fine?

Ok, let me fix stuff.

qinqon:
@varshaprasad96 @alvaroaleman test fixed, I have regenerate all the testing CRDs inpacted so you can check how "description" field looks now.

varshaprasad96 (approved):
/lgtm

alvaroaleman (approved):

Issue #728, opened 2022-09-26
Hide `// TODO:` from generated CRD.


kaovilai:
Would be awesome to be able to annotate each `_types.go` field if there is a TODO but without having it live in a CRD for end users to see.

cbandy:
Related: #649

camilamacedo86:
Hi @kaovilai,

Would you like to work on a PR to propose this solution?

kaovilai:
I'm open to it :)

kaovilai:
unstale

sbueringer:
/ifecycle frozen

I think it makes sense to strip TODO's like upstream. We don't want TODOs showing up in CRDs, espeically if they are on types from k/k.

mythi:
seems to be fixed in 0.16.x?

sbueringer:
Yeah, I think so! 

I didn't remember this issue exists :)

/close

PR #877, opened 2024-01-24
:warning: Strip comments from CRD descriptions

### LINKED ISSUE #649: Imported descriptions do not match upstream
`controller-gen crd` puts `TODO` in imported descriptions which don't appear in Kubernetes descriptions.

For example, [corev1.LocalObjectReference](https://github.com/kubernetes/kubernetes/blob/v1.23.0/staging/src/k8s.io/api/core/v1/types.go#L5606-L5612) has a `TODO` on its one Name field. This type is embedded in [corev1.ConfigMapEnvSource](https://github.com/kubernetes/kubernetes/blob/v1.23.0/staging/src/k8s.io/api/core/v1/types.go#L2088), so it appears in any CRD that uses [corev1.EnvFromSource](https://github.com/kubernetes/kubernetes/blob/v1.23.0/staging/src/k8s.io/api/core/v1/types.go#L2075) including the CronJob example.

```go
type LocalObjectReference struct {
	// Name of the referent.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
	// TODO: Add other useful fields. apiVersion, kind, uid?
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
}
```

https://github.com/kubernetes-sigs/controller-tools/blob/c796d037386bc209fde0ca657df6b6face0ec5b7/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml#L1067-L1072

This `TODO` does _not_ appear, however, in the OpenAPI Specification for [io.k8s.api.core.v1.ConfigMapEnvSource](https://github.com/kubernetes/kubernetes/tree/v1.23.0/api/openapi-spec) nor the [Kubernetes documentation](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#environment-variables).

```json
    "io.k8s.api.core.v1.ConfigMapEnvSource": {
      "description": "ConfigMapEnvSource selects a ConfigMap to populate the environment variables with.\n\nThe contents of the target ConfigMap's Data field will represent the key-value pairs as environment variables.",
      "properties": {
        "name": {
          "description": "Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names",
          "type": "string"
        },
        "optional": {
          "description": "Specify whether the ConfigMap must be defined",
          "type": "boolean"
        }
      },
      "type": "object"
    },
```

Some difference between these two processes is adding `TODO` to my CRD, and I would like `controller-gen` to not do that.

tsaarni:
This PR strips comments from the generated CRDs for clarity of the public API documentation.

The change aligns controller-tools with Kubernetes conventions for type descriptions:

- Line that only contains `---` separates the API description from comments. The leading lines are part of public documentation, the trailing lines are internal instructions such as code examples etc. As an example, see widely used type [`Condition`](https://github.com/kubernetes/apimachinery/blob/02a41040d88da08de6765573ae2b1a51f424e1ca/pkg/apis/meta/v1/types.go#L1508-L1522) that includes some example code that should not be considered as part of API documentation.
- Line that begins with TODO is internal note for implementer, and not part of public documentation of the type. See [`LocalObjectReference`](https://github.com/kubernetes/kubernetes/blob/a1ffdedf782edf1472102b0b99c1467d4ed39753/staging/src/k8s.io/api/core/v1/types.go#L6418-L6423) for example.

**Background information**

When Kubernetes itself generates OpenAPI docs for its internal resources, it will strip the implementation related comments from the go docs. I believe it is done by [this code](https://github.com/kubernetes/apimachinery/blob/02a41040d88da08de6765573ae2b1a51f424e1ca/pkg/runtime/swagger_doc_generator.go#L55-L93). 

However with controller-tools the implementation comments "leak" to public documentation. This causes confusion and usability issues since the users have no context to understand the implementation comments. 

Google search for ["Represents the observations of a foo's current state"](https://www.google.com/search?q=%22Represents+the+observations+of+a+foo%27s+current+state%22) shows that the problem is wide spread.

**Limited compatibility with markdown formatting**

I've marked the PR as :warning: breaking with following reasoning:

While I first thought it is [very unlikely](https://github.com/kubernetes-sigs/controller-tools/issues/875#issuecomment-1895595757) to bump into `---` in type's documentation, I'm proven wrong by #870 which shows that sometimes markdown formatting is used inside go docs.  `kubectl explain` will not understand how to format markdown, but the reasoning for markdown inside go docs might come from [OpenAPI spec](https://swagger.io/specification/#rich-text-formatting) which supports "rich text formatting" with [CommonMark 0.27](https://spec.commonmark.org/0.27/)

There is a limitation:  There is no reliable way to know if go doc is in plain text or markdown, therefore `---` could mean comment separator (according to Kubernetes convention) or alternatively [thematic break](https://spec.commonmark.org/0.27/#thematic-break) or [heading](https://spec.commonmark.org/0.27/#setext-heading) (according to markdown). The implementation takes an approach where `---` inside markdown [fenced code blocks](https://spec.commonmark.org/0.27/#fenced-code-blocks) is kept, but when seen in plain, it is considered as comment separator and rest of the doc is stripped. This might be wrong assumption in some corner cases. As result the description in the generated CRD will be cut short.

Fixes #649, #728, #875

### DIFF
--- a/pkg/markers/zip.go
+++ b/pkg/markers/zip.go
@@ -69,13 +69,15 @@ func extractDoc(node ast.Node, decl *ast.GenDecl) string {
 	// split lines, and re-join together as a single
 	// paragraph, respecting double-newlines as
 	// paragraph markers.
-	outLines := strings.Split(outGroup.Text(), "\n")
-	if outLines[len(outLines)-1] == "" {
+	lines := strings.Split(outGroup.Text(), "\n")
+	if lines[len(lines)-1] == "" {
 		// chop off the extraneous last part
-		outLines = outLines[:len(outLines)-1]
+		lines = lines[:len(lines)-1]
 	}
 
-	for i, line := range outLines {
+	var outLines []string
+	var insideCodeBlock bool
+	for i, line := range lines {
 		if isAsteriskComment {
 			// Trim any extranous whitespace,
 			// for handling /*…*/-style comments,
@@ -86,10 +88,33 @@ func extractDoc(node ast.Node, decl *ast.GenDecl) string {
 		// Respect that double-newline means
 		// actual newline:
 		if line == "" {
-			outLines[i] = "\n"
+			lines[i] = "\n"
 		} else {
-			outLines[i] = line
+			lines[i] = line
 		}
+
+		// Recognize markdown code blocks (``` or ~~~)
+		// https://spec.commonmark.org/0.27/#fenced-code-blocks
+		if strings.HasPrefix(line, "```") || strings.HasPrefix(line, "~~~") {
+			insideCodeBlock = !insideCodeBlock
+		}
+
+		if !insideCodeBlock {
+			// If we are not inside markdown code block, follow the Kubernetes formatting conventions:
+			// - Lines after --- are comments and should be ignored.
+			// - Lines starting with TODO are comments and should be ignored.
+			// See function fmtRawDoc() in https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/swagger_doc_generator.go
+
+			if strings.HasPrefix(line, "TODO") {
+				continue
+			}
+
+			if strings.HasPrefix(line, "---") {
+				break
+			}
+		}
+
+		outLines = append(outLines, line)
 	}
 	return strings.Join(outLines, "\n")
 }
--- a/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
+++ b/pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml
@@ -1111,7 +1111,6 @@ spec:
                                                     description: |-
                                                       Name of the referent.
                                                       More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-                                                      TODO: Add other useful fields. apiVersion, kind, uid?
                                                     type: string
                                                   optional:
                                                     description: Specify whether the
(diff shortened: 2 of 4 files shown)


tsaarni:
I'm cc'ing @qinqon, @cbandy, in case you are interested to comment this PR as well.

kaovilai (approved):
For my use this is lgtm

pkg/markers/zip.go
```diff
@@ -86,10 +88,33 @@ func extractDoc(node ast.Node, decl *ast.GenDecl) string {
...
+
+		// Recognize markdown code blocks (``` or ~~~)
+		// https://spec.commonmark.org/0.27/#fenced-code-blocks
+		if strings.HasPrefix(line, "```") || strings.HasPrefix(line, "~~~") {
+			insideCodeBlock = !insideCodeBlock
+		}
+
+		if !insideCodeBlock {
+			// If we are not inside markdown code black, follow the Kubernetes formatting conventions:
+			// - Lines after --- are comments and should be ignored.
```
vincepri, line 104:
```suggestion
			// If we are not inside markdown code block, follow the Kubernetes formatting conventions:
			// - Lines after --- are comments and should be ignored.
```

vincepri (commented):
One minor comment, otherwise lgtm!

pkg/markers/zip.go
tsaarni: Fixed!

vincepri (approved):

PR #892, opened 2024-03-11
:bug: Fix SIGSERV when RawSpec is nil (due to anonymous struct)

rkettelerij:
Under some circumstances `RawSpec` can be nil. I've encountred this when generating a CRD of a complex set of structs. Currently the `structToSchema` func fails hard on this (SIGSERV). This PR fixes the latter.

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -341,7 +341,7 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiext.JSON
 		Properties: make(map[string]apiext.JSONSchemaProps),
 	}
 
-	if ctx.info.RawSpec.Type != structType {
+	if ctx.info.RawSpec != nil && ctx.info.RawSpec.Type != structType {
 		ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("encountered non-top-level struct (possibly embedded), those aren't allowed"), structType))
 		return props
 	}


pkg/crd/schema.go
```diff
@@ -341,7 +341,7 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiext.JSON
 		Properties: make(map[string]apiext.JSONSchemaProps),
 	}
 
-	if ctx.info.RawSpec.Type != structType {
+	if ctx.info.RawSpec != nil && ctx.info.RawSpec.Type != structType {
```
vincepri, line 344: could we add a test case?

rkettelerij: Hi, I've managed to break the complex set of structs down to the following test case. `controller-get crd
paths="<path/to/testcase/below>/..." output:stdout` currently fails on this one. With the fix it doesn't fail. It's due to the use of an anonymous struct.

```
// +groupName=test
package config

import (
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

type ExampleSpec struct {
	Service Service `json:"service"`
}

type Example struct {

	Spec ExampleSpec `json:"spec,omitempty"`
}

type Service struct {

	// +optional
	Foobar struct {
		// placeholder
	} `yaml:"foobar" json:"foobar"`
}
```

alvaroaleman: @rkettelerij please update pkg/crd/testdata/cronjob_types.go with an anonymous struct to test this

/ok-to-test

sbueringer:
@rkettelerij Do you have time to address the finding?

PR #897, opened 2024-03-26
:bug: pkg/crd: support validating internal list items on list types

### LINKED ISSUE #342: Support validating internal list items on list types
Per https://book.kubebuilder.io/reference/markers/crd-validation.html, validating tags like `+kubebuilder:validation:MinLength` and `kubebuilder:validation:Pattern` must be specified on `string` fields or `string` types. If one has a field of type `[]string`, however, it is impossible to use such tags on the field directly to enforce validation on the individual list items. This forces users to alias the list item type and place the validation on the type itself like so:

```go
type NonEmptyString string

// +k8s:openapi-gen=true
type CRSpec struct {
  Field []NonEmptyString `json:"field"`
}
```

This leads to undesired consequences like having to export the type so that it can be referenced in controller packages and having to cast list items since there is no easy way to cast `[]NonEmptyString` to `[]string`.

It used to previously be possible to add the validation tag to the field itself like so:

```go
// +k8s:openapi-gen=true
type CRSpec struct {
  Field []string `json:"field"`
}
```

Doing so, however, would generate the following CRD where the validation was present on both the list itself as well as its items:

```yaml
field:
  items:
    minLength: 1
    type: string
  minLength: 1
  type: array
```

Now, the CRD is generated as follows:

```yaml
field:
  items:
    type: string
  type: array
```

It'd be great if the validating tags could be specified on the field itself and result in the following CRD:

```yaml
field:
  items:
    minLength: 1
    type: string
  type: array
```

Previously, I was using `v0.9.0` of the Operator SDK which used `controller-tools` `v0.1.11-0.20190411181648-9d55346c2bde`. Now, I'm using `v0.11.0` of the Operator SDK which uses `controller-tools` `v0.2.0`

AlexanderYastrebov:
For #342

### DIFF
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -252,6 +252,24 @@ type CronJobSpec struct {
 
 	// Checks that arrays work when the type contains a composite literal
 	ArrayUsingCompositeLiteral [len(struct{ X [3]int }{}.X)]string `json:"arrayUsingCompositeLiteral,omitempty"`
+
+	// This tests string slice item validation.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?([.][a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	Hosts []string `json:"hosts,omitempty"`
+
+	HostsAlias Hosts `json:"hostsAlias,omitempty"`
+
+	// This tests string alias slice item validation.
+	LongerStringArray []LongerString `json:"longerStringArray,omitempty"`
+
+	// This tests that a slice of IntOrString can also have a pattern attached to it.
+	// This can be useful if you want to limit the string to a perecentage or integer.
+	// The XIntOrString marker is a requirement for having a pattern on this type.
+	// +kubebuilder:validation:XIntOrString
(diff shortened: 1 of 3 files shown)


pkg/crd/testdata/cronjob_types.go
```diff
@@ -252,6 +252,24 @@ type CronJobSpec struct {
 
 	// Checks that arrays work when the type contains a composite literal
 	ArrayUsingCompositeLiteral [len(struct{ X [3]int }{}.X)]string `json:"arrayUsingCompositeLiteral,omitempty"`
+
+	// This tests string slice item validation.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?([.][a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	Hosts []string `json:"hosts,omitempty"`
```
alvaroaleman, line 260: I think we have to find a way to differentiate if the validation is for the list as a whole or for the individual items in the list. This change just changes things to assume that its for the items. I don't think that is correct and its a breaking change

AlexanderYastrebov: > I don't think that is correct and its a breaking change

I don't think its a breaking change as previously these markers were rejected for slice types.
I also do not see a solid reason to introduce a new set of "items" markers as they will be mutually exclusive with those I change in this PR.

AlexanderYastrebov:
Closing in favour of https://github.com/kubernetes-sigs/controller-tools/pull/898

PR #903, opened 2024-04-05
:sparkles: Build and publish Kubernetes envtest tools as packages

vincepri:

### DIFF
--- a/.github/workflows/tools-releases.yml
+++ b/.github/workflows/tools-releases.yml
@@ -1,0 +1,38 @@
+name: Kubernetes Tools Releases for controller-runtime envtest
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        kubernetes_version: [ 1.28.0 ]
+        os: [ linux darwin windows ]
+        arch: [ amd64 arm64 ]
+        include:
+          - arch: ppc64le
+            os: linux
+          - arch: s390x
+            os: linux
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Set up Docker
+        uses: docker/setup-buildx-action@v3
+      - name: Build and push Docker images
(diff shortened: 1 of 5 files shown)


.github/workflows/tools-releases.yml
```diff
@@ -0,0 +1,38 @@
+name: Kubernetes Tools Releases for controller-runtime envtest
+
+on:
+  push:
+    branches:
+      - main
```
vincepri, line 6: We might want to switch to a release eventually, this needs to be tested, so for now it might be ok

hack/tools/bin/golangci-lint
```diff
@@ -0,0 +1 @@
+/Users/vince/go/src/sigs.k8s.io/controller-tools/hack/tools/bin/golangci-lint-v1.57.2
```
sbueringer, line 1: Looks like the binary & symlink is checked in

.github/workflows/tools-releases.yml
```diff
@@ -0,0 +1,38 @@
...
+      matrix:
+        kubernetes_version: [ 1.28.0 ]
+        os: [ linux darwin windows ]
+        arch: [ amd64 arm64 ]
+        include:
+          - arch: ppc64le
+            os: linux
+          - arch: s390x
+            os: linux
+    steps:
```
sbueringer, line 22: eventually we should pin to the sha's

.github/workflows/tools-releases.yml
```diff
@@ -0,0 +1,38 @@
...
+      fail-fast: true
+      matrix:
+        kubernetes_version: [ 1.28.0 ]
+        os: [ linux darwin windows ]
+        arch: [ amd64 arm64 ]
+        include:
+          - arch: ppc64le
+            os: linux
+          - arch: s390x
+            os: linux
```
sbueringer, line 21: I think this combinations won't get kubernetes_version (if I read this here correctly: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)

PR #607, opened 2021-08-31
⚠ bump to kubernetes 1.22.1, remove support for legacy v1beta1 CRDs and webhooks

joelanford:
This PR bumps controller-tools k8s dependencies to 1.22.1

This is a breaking change. It removes support for v1beta1 CRDs and webhooks. These APIs are no longer served in 1.22.

PR #912, opened 2024-04-10
:sparkles: crd: allow specifying spec.preserveUnknownFields

### LINKED ISSUE #476: preserveUnknownFields=false doesn't work with crdVersions=v1

AlexanderYastrebov:
Reinstate crd:preserveUnknownFields option removed by #607 to allow specifying the value of deprecated spec.preserveUnknownFields CRD field.

This is useful for updating CRDs that were automatically converted from v1beta1 version which had true as a default value and resulted in:
```
$ kubectl get crd foo.bar -o yaml | grep preserveUnknownFields
  preserveUnknownFields: true
    message: 'spec.preserveUnknownFields: Invalid value: true: must be false'
```

For #476

### DIFF
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -85,6 +85,13 @@ type Generator struct {
 
 	// Year specifies the year to substitute for " YEAR" in the header file.
 	Year string `marker:",optional"`
+
+	// PreserveUnknownFields indicates whether or not we should turn off pruning.
+	//
+	// Specifies spec.preserveUnknownFields value that is false and omitted by default.
+	//
+	// It's required to be false for v1 CRDs.
+	PreserveUnknownFields *bool `marker:",optional"`
 }
 
 func (Generator) CheckFilter() loader.NodeFilter {
@@ -98,6 +105,16 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
 func transformRemoveCRDStatus(obj map[string]interface{}) error {
 	delete(obj, "status")
 	return nil
+}
+
+// transformPreserveUnknownFields adds spec.preserveUnknownFields=value.
+func transformPreserveUnknownFields(value bool) func(map[string]interface{}) error {
+	return func(obj map[string]interface{}) error {
+		if spec, ok := obj["spec"].(map[interface{}]interface{}); ok {
+			spec["preserveUnknownFields"] = value
+		}
+		return nil
+	}
 }
 
 func (g Generator) Generate(ctx *genall.GenerationContext) error {
@@ -146,6 +163,14 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	}
 	headerText = strings.ReplaceAll(headerText, " YEAR", " "+g.Year)
 
+	yamlOpts := []*genall.WriteYAMLOptions{
+		genall.WithTransform(transformRemoveCRDStatus),
+		genall.WithTransform(genall.TransformRemoveCreationTimestamp),
+	}
+	if g.PreserveUnknownFields != nil {
+		yamlOpts = append(yamlOpts, genall.WithTransform(transformPreserveUnknownFields(*g.PreserveUnknownFields)))
+	}
+
 	for _, groupKind := range kubeKinds {
 		parser.NeedCRDFor(groupKind, g.MaxDescLen)
 		crdRaw := parser.CustomResourceDefinitions[groupKind]
@@ -171,7 +196,7 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 			} else {
 				fileName = fmt.Sprintf("%s_%s.%s.yaml", crdRaw.Spec.Group, crdRaw.Spec.Names.Plural, crdVersions[i])
 			}
-			if err := ctx.WriteYAML(fileName, headerText, []interface{}{crd}, genall.WithTransform(transformRemoveCRDStatus), genall.WithTransform(genall.TransformRemoveCreationTimestamp)); err != nil {
+			if err := ctx.WriteYAML(fileName, headerText, []interface{}{crd}, yamlOpts...); err != nil {
 				return err
 			}
 		}
--- a/pkg/crd/gen.go
+++ b/pkg/crd/gen.go
@@ -86,12 +86,19 @@ type Generator struct {
 	// Year specifies the year to substitute for " YEAR" in the header file.
 	Year string `marker:",optional"`
 
-	// PreserveUnknownFields indicates whether or not we should turn off pruning.
+	// DeprecatedV1beta1CompatibilityPreserveUnknownFields indicates whether
+	// or not we should turn off field pruning for this resource.
 	//
 	// Specifies spec.preserveUnknownFields value that is false and omitted by default.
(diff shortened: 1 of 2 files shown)


pkg/crd/gen_integration_test.go
```diff
@@ -119,6 +119,32 @@ var _ = Describe("CRD Generation proper defaulting", func() {
 		expectedOut := string(expectedFileFoos) + string(expectedFileZoos)
 		Expect(out.buf.String()).To(Equal(expectedOut), cmp.Diff(out.buf.String(), expectedOut))
 	})
+
+	It("should add preserveUnknownFields=false when specified", func() {
```
sbueringer, line 123: Let's also add a case that it doesn't show up with false if PreserveUnknownFields is nil

sbueringer (commented):
Makes sense to me as the alternative seems to be that users have to modify the CRD objects manually

pkg/crd/gen_integration_test.go
AlexanderYastrebov: Added

sbueringer:
Thx!

/ok-to-test

/lgtm

Let's see if someone else has an opinion on it

(cc @vincepri @alvaroaleman )

pkg/crd/gen_integration_test.go
```diff
@@ -119,6 +119,43 @@ var _ = Describe("CRD Generation proper defaulting", func() {
...
+		By("searching preserveUnknownFields")
+		Expect(out.buf.String()).To(ContainSubstring("preserveUnknownFields: false"))
+	})
+
+	It("should add preserveUnknownFields=true when specified", func() {
+		By("calling Generate")
+		yes := true
+		gen := &crd.Generator{
+			CRDVersions:           []string{"v1"},
+			PreserveUnknownFields: &yes,
```
vincepri, line 141: Shouldn't we disallow adding true for v1?

sbueringer: As we don't support generating v1beta1 CRDs anymore, I think it's as easy as "always" disallowing true

AlexanderYastrebov: Is it really necessary? API server does not allow `preserveUnknownFields: true` anyway:
```console
kubectl apply -f pkg/crd/testdata/gen/bar.example.com_foos.yaml
The CustomResourceDefinition "foos.bar.example.com" is invalid: 
* spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema
* spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-kubernetes-preserve-unknown-fields to true in spec.versions[*].schema instead
```

AlexanderYastrebov: I think we could let user specify `true` even though it is not supported but I added a check that prohibits that as suggested.

sbueringer: I prefer that we block generation of invalid CRD YAML like in this case. The earlier it fails the better (CRD generation instead of CRD deployment)

sbueringer:
Thx!

/lgtm

/assign @vincepri

pkg/crd/gen_integration_test.go
alvaroaleman: @sbueringer it is only invalid if you create the CRD as v1. If it was originally created as v1beta1, then it is valid.

sbueringer: You can still deploy a v1 CRD with the field set to true if the CRD was before created with v1beta1?

alvaroaleman: > You can still deploy a v1 CRD with the field set to true if the CRD was before created with v1beta1?

Yes

AlexanderYastrebov: Let me know if I should remove the check to allow `true` value.

pkg/crd/gen.go
```diff
@@ -85,6 +85,13 @@ type Generator struct {
 
 	// Year specifies the year to substitute for " YEAR" in the header file.
 	Year string `marker:",optional"`
+
+	// PreserveUnknownFields indicates whether or not we should turn off pruning.
+	//
+	// Specifies spec.preserveUnknownFields value that is false and omitted by default.
+	//
+	// It's required to be false for v1 CRDs.
+	PreserveUnknownFields *bool `marker:",optional"`
```
vincepri, line 94:
```suggestion
	// DeprecatedV1beta1CompatibilityPreserveUnknownFields indicates whether 
	// or not we should turn off field pruning for this resource.
	//
	// Specifies spec.preserveUnknownFields value that is false and omitted by default.
	// This value can only be specified for CustomResourceDefinitions that were created with
	// `apiextensions.k8s.io/v1beta1`.
	//
	// The field can be set for compatiblity reasons, although strongly discouraged, resource
	// authors should move to a structural OpenAPI schema instead.
	//
	// See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning 
	// for more information about field pruning and v1beta1 resources compatibility.
	DeprecatedV1beta1CompatibilityPreserveUnknownFields *bool `marker:",optional"`
```

Something like that?

pkg/crd/gen_integration_test.go
vincepri: Added a suggestion above to rename the exposed field and prevent possible footguns

sbueringer: > Let me know if I should remove the check to allow true value.

Yup, we should allow true. Sorry I thought that v1 CRDs cannot be deployed with true

pkg/crd/gen.go
AlexanderYastrebov: Is this necessary? By default the marker is not used, it will be used by those who solve specific problem and know what they want.

pkg/crd/gen_integration_test.go
AlexanderYastrebov: Removed the check to allow `true`.

pkg/crd/gen.go
AlexanderYastrebov: The link is useful as well as different name from previously existing marker, fixed.

PR #937, opened 2024-05-02
✨ rbac: deduplicate having the same groups, resourceNames, URLs and Verbs

chrischdi:
This improves the output of controller-gens rbac generator to further deduplicate the output.

It now:
* first separates the rules by namespace **and resource**
* and later deduplicates on the set resources if the rules are otherwise the same (excluding resource).
* and also deduplicates based on apiGroups if possible.

The first commit only adds test data.

The second commit adds the change for deduplicating based on resources

The third commit deduplicates on apiGroups.


pkg/rbac/parser_integration_test.go
```diff
@@ -52,7 +52,7 @@ var _ = Describe("ClusterRole generated by the RBAC Generator", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			By("parsing the desired YAML")
-			for i, expectedRoleBytes := range bytes.Split(expectedFile, []byte("\n---\n"))[1:] {
+			for i, expectedRoleBytes := range bytes.Split(expectedFile, []byte("\n---\n")) {
```
chrischdi, line 55: Note: this change allows to use the directly generated output as golden file. The `[1:]` is not necessary anymore because the document starts with `---\n`

pkg/rbac/parser.go
```diff
@@ -93,6 +93,18 @@ func (r *Rule) key() ruleKey {
 	}
 }
 
+func (r *Rule) KeyWithoutResources() string {
```
sbueringer, line 96: Not sure if the name is ideal. It also adds verbs compared to the key

pkg/rbac/parser.go
```diff
@@ -93,6 +93,18 @@ func (r *Rule) key() ruleKey {
...
+	key := r.key()
+	verbs := strings.Join(r.Verbs, "&")
+	return fmt.Sprintf("%s + %s + %s + %s", key.Groups, key.ResourceNames, key.URLs, verbs)
+}
+
+func (r *Rule) KeyWithoutGroups() string {
```
sbueringer, line 102: Same here

pkg/rbac/parser.go
```diff
@@ -168,22 +180,32 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
+					Groups:        rule.Groups,
+					Resources:     []string{resource},
+					ResourceNames: rule.ResourceNames,
+					URLs:          rule.URLs,
+					Namespace:     rule.Namespace,
+					Verbs:         rule.Verbs,
+				}
+				namespace := r.Namespace
+				if _, ok := rulesByNSResource[namespace]; !ok {
+					rules := make([]*Rule, 0)
```
sbueringer, line 204: Creating an empty slice should not be necessary I think (Just saw it was the same before, but still I guess we can drop the entire if?)

pkg/rbac/parser.go
```diff
@@ -168,22 +180,32 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
-	rulesByNS := make(map[string][]*Rule)
+	rulesByNSResource := make(map[string][]*Rule)
 	for _, root := range ctx.Roots {
 		markerSet, err := markers.PackageMarkers(ctx.Collector, root)
 		if err != nil {
 			root.AddError(err)
 		}
 
-		// group RBAC markers by namespace
+		// group RBAC markers by namespace and resource
```
sbueringer, line 190: If that comment refres to the map key it is still only grouped by namespace

pkg/rbac/parser.go
sbueringer: Same below

pkg/rbac/parser.go
sbueringer: Maybe let's enumerate what is in the key (KeyWith...)? (+unexport the func if possible)

pkg/rbac/parser.go
```diff
@@ -200,6 +222,53 @@ func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{
...
+				ruleMapWithoutResources[key] = rules
+			}
+			ruleMapWithoutResources[key] = append(ruleMapWithoutResources[key], rule)
+		}
+		// 2. merge to ruleMap
+		ruleMap = make(map[ruleKey]*Rule)
+		for _, rules := range ruleMapWithoutResources {
+			rule := rules[0]
+			for _, mergeRule := range rules[1:] {
+				rule.Resources = append(rule.Resources, mergeRule.Resources...)
```
sbueringer, line 242: Isn't it a problem that we are losing Namespace of rules[1:]?

chrischdi: NormalizeRules gets executed below for a single namespace. So all rules passed to this function act on the same namespace.

pkg/rbac/parser.go
```diff
@@ -200,6 +218,49 @@ func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{
...
+
+			key := rule.key()
+			ruleMap[key] = rule
+		}
+
+		// deduplicate groups
+		// 1. create map based on key without group
+		ruleMapWithoutGroup := make(map[string][]*Rule)
+		for _, rule := range ruleMap {
+			// unset Resources on the key
```
sbueringer, line 245:
```suggestion
			// get key without Group
```
"resources" was a copy&paste error from above. But maybe this comment would be better (same above)

pkg/rbac/parser.go
```diff
@@ -200,6 +218,49 @@ func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{
...
+			key := rule.keyWithResourcesResourceNamesURLsVerbs()
+			if _, ok := ruleMapWithoutGroup[key]; !ok {
+				rules := make([]*Rule, 0)
+				ruleMapWithoutGroup[key] = rules
+			}
```
sbueringer, line 250:
```suggestion
```

sbueringer (commented):
A few smaller findings and some ideas to test more before merge

chrischdi:
> Could you also open a PR against core CAPI using this PR? (doesn't have to be a clean install of the binary, just want to validate the differences in generated roles)
> 
> (I would also bump to CT main once this PR merges in CAPI)

Example PR: https://github.com/kubernetes-sigs/cluster-api/pull/10612

sbueringer:
Thx!

CAPI PR looks entirely as expected

Really nice optimization, Roles look much cleaner now

/lgtm
/approve

sbueringer:
Ah interesting. Can you open an issue for that, please?

We should be able to improve this further :)

(cc @chrischdi, maybe you have some time to look into it (let's also add test coverage for this case))

pmalek:
> Ah interesting. Can you open an issue for that, please?
> 
> We should be able to improve this further :)
> 
> (cc @chrischdi, maybe you have some time to look into it (let's also add test coverage for this case))

@sbueringer #1036

rakesh-garimella:
Hi @chrischdi I tested the change with 0.16.1. Found one possible issue. We have some codemarker for `nonResourceURLs`
```
//+kubebuilder:rbac:urls=/metrics,verbs=get
//+kubebuilder:rbac:urls=/metrics/cadvisor,verbs=get
```

But after generating the manifests these RBAC are missing.

PR #938, opened 2024-05-03
✨ Add support for +default markers

liggitt:
This adds annotation support for `+default` markers for specifying default values in JSON format, as is done in core Kubernetes types and described in https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1929-built-in-default#summary

If both `+default` and `+kubebuilder:default` are specified, the kubebuilder-specific annotation wins.

Fixes https://github.com/kubernetes-sigs/controller-tools/issues/939

### DIFF
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -38,6 +38,8 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+const DefaultRefValue = "defaultRefValue"
+
 // CronJobSpec defines the desired state of CronJob
 // +kubebuilder:validation:XValidation:rule="has(oldSelf.forbiddenInt) || !has(self.forbiddenInt)",message="forbiddenInt is not allowed",fieldPath=".forbiddenInt",reason="FieldValueForbidden"
 type CronJobSpec struct {
@@ -116,9 +118,9 @@ type CronJobSpec struct {
 	// +kubebuilder:example={a,b}
 	DefaultedSlice []string `json:"defaultedSlice"`
 
-	// This tests that object defaulting can be performed.
-	// +kubebuilder:default={{nested: {foo: "baz", bar: true}},{nested: {bar: false}}}
-	// +kubebuilder:example={{nested: {foo: "baz", bar: true}},{nested: {bar: false}}}
+	// This tests that slice and object defaulting can be performed.
+	// +kubebuilder:default={{nested: {foo: "baz", bar: true}},{nested: {foo: "qux", bar: false}}}
+	// +kubebuilder:example={{nested: {foo: "baz", bar: true}},{nested: {foo: "qux", bar: false}}}
 	DefaultedObject []RootObject `json:"defaultedObject"`
 
 	// This tests that empty slice defaulting can be performed.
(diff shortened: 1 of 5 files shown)


pkg/markers/parse.go
```diff
@@ -824,6 +831,49 @@ func parserScanner(raw string, err func(*sc.Scanner, string)) *sc.Scanner {
...
+	if structFieldName != "" {
+		// it's a struct field mapped to an anonymous marker
+		outTarget = out.FieldByName(structFieldName)
+		if !outTarget.CanSet() {
+			errs = append(errs, fmt.Errorf("cannot set field %q (might not exist)", structFieldName))
+			return out.Interface(), loader.MaybeErrList(errs)
+		}
+	}
+
+	fmt.Println("parse", fields, "into", outTarget)
```
alvaroaleman, line 864: Debugging leftover?

liggitt: heh, oops

liggitt: fixed

liggitt:
/hold

forgot about the ref(type) support in the KEP... that's going to be a pain to reimplement :-/

alexzielenski:
> cc @alexzielenski for thoughts on the best way to add support for +default=ref(...) to kubebuilder

In our code generators we use [`ParseSymbolReference`](https://github.com/kubernetes/gengo/blob/050f4528535e704ce42d5a9ca6b3974e6bed8fd0/examples/defaulter-gen/generators/defaulter.go#L519-L537) to parse the ref string into an symbol name and package path. But our output is not the final YAML schema [like it is for kubebuilder](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/gen.go#L181-L209) - it is Go code to express a `spec.Schema`.

To add `ref` support to this I don't see a way except a large refactor to generate code that expresses a spec.Schema like kube-openapi, or if there is a way to statically evaluate a constant using Go tools which I have not seen.

pkg/crd/testdata/cronjob_types.go
```diff
@@ -133,6 +133,35 @@ type CronJobSpec struct {
...
+	// This tests that primitive defaulting can be performed.
+	// +default="forty-two"
+	KubernetesDefaultedString string `json:"kubernetesDefaultedString"`
+
+	// This tests that slice defaulting can be performed.
+	// +default=["a","b"]
+	KubernetesDefaultedSlice []string `json:"kubernetesDefaultedSlice"`
+
+	// This tests that object defaulting can be performed.
+	// +default=[{"nested": {"foo": "baz", "bar": true}},{"nested": {"bar": false}}]
```
sbueringer, line 150: nit: technically this default value is invalid because foo is a required field

liggitt: ha, that's fun... this just copied the existing default example... fixed up here and the existing one

liggitt:
> To add `ref` support to this I don't see a way except a large refactor to generate code that expresses a spec.Schema like kube-openapi, or if there is a way to statically evaluate a constant using Go tools which I have not seen.

yeah... it's really thorny... I might start by limiting support here to inlined json-formatted values. We don't actually use `ref()` in any in-tree objects... we might not want to as long as its difficult for CRD generators to make use of them

alexzielenski:
> I might start by limiting support here to inlined json-formatted values. 

Sounds reasonable. I've looked around a little, and closest thing I found to static evaluation of constants from symbol path is `go doc`:

```console
❯ go doc k8s.io/api/core/v1.TerminationMessageReadFile
package v1 // import "k8s.io/kubernetes/staging/src/k8s.io/api/core/v1"

const (
        // TerminationMessageReadFile is the default behavior and will set the container status message to
        // the contents of the container's terminationMessagePath when the container exits.
        TerminationMessageReadFile TerminationMessagePolicy = "File"
        // TerminationMessageFallbackToLogsOnError will read the most recent contents of the container logs
        // for the container status message when the container exits with an error and the
        // terminationMessagePath has no contents.
        TerminationMessageFallbackToLogsOnError TerminationMessagePolicy = "FallbackToLogsOnError"
)
```

Unfortunately it is not machine readable, not isolated to what was asked for, and returns go source so if the constant was an expression like `24 * time.Hour` then we'd not be able to handle it simply by parsing this.

pkg/crd/testdata/cronjob_types.go
```diff
@@ -133,6 +135,39 @@ type CronJobSpec struct {
...
+	// +default="forty-two"
+	KubernetesDefaultedString string `json:"kubernetesDefaultedString"`
+
+	// This tests that slice defaulting can be performed.
+	// +default=["a","b"]
+	KubernetesDefaultedSlice []string `json:"kubernetesDefaultedSlice"`
+
+	// This tests that object defaulting can be performed.
+	// +default=[{"nested": {"foo": "baz", "bar": true}},{"nested": {"foo": "qux", "bar": false}}]
+	KubernetesDefaultedObject []RootObject `json:"kubernetesDefaultedObject"`
```
sbueringer, line 153: This technically tests defaulting of a slice :)

pkg/crd/testdata/cronjob_types.go
```diff
@@ -133,6 +135,39 @@ type CronJobSpec struct {
...
+
+	// This tests that empty slice defaulting can be performed.
+	// +default=[]
+	KubernetesDefaultedEmptySlice []string `json:"kubernetesDefaultedEmptySlice"`
+
+	// This tests that an empty object defaulting can be performed on a map.
```
sbueringer, line 159:
```suggestion
	// This tests that an empty map defaulting can be performed on a map.
```
nit

sbueringer (commented):
Just a few nits

pkg/crd/testdata/cronjob_types.go
liggitt: heh, copied from `DefaultedObject []RootObject `json:"defaultedObject"`` above ... I guess it defaults a slice of objects, so... both... but sure :)

pkg/crd/testdata/cronjob_types.go
liggitt: the default is a json object (`{}`) that applies to a map... I think the current message is correct (also was copied from above)

pkg/crd/testdata/cronjob_types.go
liggitt: do you want me to change this and the existing one above, or leave them, or just change the new one?

sbueringer: Maybe let's change both. I guess this works anyway, but maybe safer to default an object to get the actual coverage for this case

sbueringer:
Thx!

/lgtm
/assign @alvaroaleman

alvaroaleman (approved):
thanks!

PR #959, opened 2024-05-21
✨ Release envtest v1.28.3

sbueringer:


hack/envtest/_matrix/v1.28.3.yaml
```diff
@@ -0,0 +1,2 @@
+go: 1.21
```
chrischdi, line 1: Should be 1.20: https://github.com/kubernetes/kubernetes/blob/v1.28.3/.go-version
```suggestion
go: 1.20
```

sbueringer: Good point. I would keep 1.21 here, as we already used it for 1.28.0

It wasn't correct, but it would be strange to "downgrade" in a higher version now

chrischdi (commented):
note: https://github.com/kubernetes-sigs/controller-tools/pull/959/files#r1608526958

/lgtm
/hold

PR #968, opened 2024-05-22
✨ Release envtest v1.25.0 (II)

sbueringer:

Now using etcd v3.5.5. darwin arm64 etcd binaries are only available starting with v3.5.5.

I think it's okay to not use 100% the correct version


chrischdi (commented):
/lgtm

PR #979, opened 2024-05-23
✨ Release envtest v1.23.5

sbueringer:


sbueringer:
/assign @fabriziopandini 

etcd version is here: https://github.com/kubernetes/kubernetes/blob/v1.23.5/cmd/kubeadm/app/constants/constants.go#L304

but we use v3.5.5 instead because patch version doesn't matter that much and <v3.5.5 doesn't provide darwin arm64 binaries

go version is here:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md
(grep for "go 1.17")

PR #1000, opened 2024-07-04
:sparkles: Make the anonymous field of the groupName marker optional

twz123:
Upstream introduced the use of empty groupName markers in kubernetes/kubernetes#125162. This breaks groupName parsing because its field is non-optional. Add an additional `mustOptional` function that takes anonymous field definitions and turns them into optional ones.

### DIFF
--- a/pkg/crd/markers/package.go
+++ b/pkg/crd/markers/package.go
@@ -22,7 +22,7 @@ import (
 
 func init() {
 	AllDefinitions = append(AllDefinitions,
-		must(markers.MakeDefinition("groupName", markers.DescribesPackage, "")).
+		mustOptional(markers.MakeDefinition("groupName", markers.DescribesPackage, "")).
 			WithHelp(markers.SimpleHelp("CRD", "specifies the API group name for this package.")),
 
 		must(markers.MakeDefinition("versionName", markers.DescribesPackage, "")).
--- a/pkg/crd/markers/register.go
+++ b/pkg/crd/markers/register.go
@@ -17,6 +17,7 @@ limitations under the License.
 package markers
 
 import (
+	"fmt"
 	"reflect"
 
 	"sigs.k8s.io/controller-tools/pkg/markers"
@@ -53,6 +54,19 @@ func (d *definitionWithHelp) clone() *definitionWithHelp {
 func must(def *markers.Definition, err error) *definitionWithHelp {
 	return &definitionWithHelp{
 		Definition: markers.Must(def, err),
+	}
+}
+
+func mustOptional(def *markers.Definition, err error) *definitionWithHelp {
+	def = markers.Must(def, err)
+	if !def.AnonymousField() {
+		def = markers.Must(def, fmt.Errorf("not an anonymous field: %v", def))
+	}
+	field := def.Fields[""]
+	field.Optional = true
+	def.Fields[""] = field
+	return &definitionWithHelp{
+		Definition: def,
 	}
 }
 
--- a/pkg/crd/parser_integration_test.go
+++ b/pkg/crd/parser_integration_test.go
@@ -100,9 +100,9 @@ var _ = Describe("CRD Generation From Parsing to CustomResourceDefinition", func
 			}
 		})
 
-		assertCRD := func(pkg *loader.Package, kind, fileName string) {
-			By(fmt.Sprintf("requesting that the %s CRD be generated", kind))
-			groupKind := schema.GroupKind{Kind: kind, Group: "testdata.kubebuilder.io"}
+		assertCRDForGroupKind := func(pkg *loader.Package, groupKind schema.GroupKind, fileName string) {
+			kind := groupKind.Kind
+			By(fmt.Sprintf("requesting that the %s CRD be generated in group %s", kind, groupKind.Group))
 			parser.NeedCRDFor(groupKind, nil)
 
 			By(fmt.Sprintf("fixing top level ObjectMeta on the %s CRD", kind))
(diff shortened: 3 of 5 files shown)


sbueringer:
@twz123 Implementation looks good. Let's please add some test coverage.  (I think somewhere below pkg/crd/testdata something that uses groupName=)

pkg/crd/markers/package.go
```diff
@@ -22,7 +22,7 @@ import (
 
 func init() {
 	AllDefinitions = append(AllDefinitions,
-		must(markers.MakeDefinition("groupName", markers.DescribesPackage, "")).
+		mustOptional(markers.MakeDefinition("groupName", markers.DescribesPackage, "")).
```
sbueringer, line 25: Is making this optional enough to handle groupName= correctly? (or will we e.g. just ignore the marker?)

(Sorry stupid question, but not really familiar with this)

twz123: > Is making this optional enough to handle groupName= correctly? (or will we e.g. just ignore the marker?)

It was at least enough to make it generate CRDs again with the expected outcome after an upgrade to 1.31.0-alpha.3.
 
> (Sorry stupid question, but not really familiar with this)

Me neither, trial and error ftw!!1 :sweat_smile:

twz123:
> @twz123 Implementation looks good. Let's please add some test coverage. (I think somewhere below pkg/crd/testdata something that uses groupName=)

Done!

pkg/crd/markers/package.go
twz123: > > Is making this optional enough to handle groupName= correctly? (or will we e.g. just ignore the marker?)
> 
> It was at least enough to make it generate CRDs again with the expected outcome after an upgrade to 1.31.0-alpha.3.

That said, it was generating the expected outcome for _me_, i.e. CRDs holding core/v1 structs could be generated again. The errors I saw w/o this look like "unable to locate schema for type "k8s.io/api/core/v1".PullPolicy" ...  Not sure what "handle groupName= correctly" entails from your point of view, though.

sbueringer: > entails from your point of view, though.

Yeah, same :)

sbueringer: I'm going to test this with Cluster API. We have a lot of CRDs in there. I'll check if there is a relevant diff after this PR

pkg/crd/markers/package.go
```diff
@@ -22,7 +22,7 @@ import (
 
 func init() {
 	AllDefinitions = append(AllDefinitions,
-		must(markers.MakeDefinition("groupName", markers.DescribesPackage, "")).
```
sbueringer, line 25: @sttts Does it sound reasonable for you to just adjust the marker here to accept an "empty" groupName marker ("// +groupName=") without any additional handling for the empty string

sbueringer: Context: Without this PR controller-gen fails like this now (with API types that use types from k8s.io/api@v0.31.0-alpha.3)

> /home/prow/go/pkg/mod/k8s.io/api@v0.31.0-alpha.3/core/v1/doc.go:21:1: missing argument "" (at <input>)
> /home/prow/go/pkg/mod/k8s.io/api@v0.31.0-alpha.3/core/v1/doc.go:21:1: missing argument "" (at <input>)
> https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api/10803/pull-cluster-api-verify-main/1811372226943913984

pkg/crd/markers/package.go
sbueringer: Tested with CAPI, no unexpected diffs

pkg/crd/markers/package.go
sttts: > without any additional handling for the empty string?

This is about embedding corev1 types in some CRD, right? So the groupName of the corev1 package does not really play a role for the CRD, does it? So yes, I think just parsing it without falling over, and then ignoring it outside of actual CRD API package sounds right.

sbueringer: > This is about embedding corev1 types in some CRD, right?

Yes exactly

> So the groupName of the corev1 package does not really play a role for the CRD, does it?

This is my understanding, yes

sbueringer: I'm not sure if anyone is doing it, but theoretically folks could use controller-gen to generate CRD for types of the k/k core API. In that case they would end up with an empty group in the CRD with `groupName=` (see the test in this PR: pkg/crd/testdata/testdata._cronjobs.yaml)

Not sure which group would be correct in this (probably hypothetical) case. But folks could just specify whatever group they want via groupName, so I assume it's fine

sttts: Empty string would be correct for the group name, `"core"` for the group in the CRD name. At least that's what kcp does: https://github.com/kcp-dev/kcp/blob/main/test/e2e/fixtures/kube/core.k8s.io_pods.yaml#L9

sbueringer: Okay, thx! 

I think then let's do the same in controller-gen for now. We can still modify this behavior later (or offer some additional markers) if people want something else for CRD generation for the core API group

sbueringer: @twz123 WDYT about handling the empty group here: https://github.com/kubernetes-sigs/controller-tools/blob/b370e9b1e3394f945fe796accc424569a1a62892/pkg/crd/spec.go#L73 to use "core" for the CRD name suffix if .Group is an empty string?

sbueringer: @sttts I was trying to deploy a CRD with name & group like this, but I'm getting these errors with a regular kube-apiserver (kindest/node:v1.30.0)
```
The CustomResourceDefinition "cronjobs.core" is invalid:
* metadata.name: Invalid value: "cronjobs.core": must be spec.names.plural+"."+spec.group
* spec.group: Required value
```

CRD:
```yaml
apiVersion: apiextensions.k8s.io/v1
metadata:
  annotations:
  name: cronjobs.core
spec:
  group: ""
  names:
    kind: CronJob
    listKind: CronJobList
    plural: cronjobs
    singular: cronjob
  scope: Namespaced
...
```

sbueringer:
Let's go ahead with this fix to unblock folks that just want to generate CRD for "non-core" apiGroups. I think this should be 99,9% of controller-gen users. Let's see if we want to do a follow-up based on the discussion here: https://github.com/kubernetes-sigs/controller-tools/pull/1000#discussion_r1677570370

pkg/crd/markers/package.go
sttts: In Kube does is not allowed today. You cannot mix native and CRD api groups.

PR #1002, opened 2024-07-10
✨ Allow customizing generated webhook configuration's name

### LINKED ISSUE #865: allow customizing generated webhook's name and its service's namespace and name
I'm using kubebuilder which generates webhooks with [these markers](https://book.kubebuilder.io/reference/markers/webhook.html). The generated mutating webhook has a hardcoded name of `mutating-webhook-configuration` [here](https://github.com/kubernetes-sigs/controller-tools/blob/881ffb4682cb5882f5764aca5a56fe9865bc9ed6/pkg/webhook/parser.go#L422), and the validating webhook has a hardcoded name of `validating-webhook-configuration` [here](https://github.com/kubernetes-sigs/controller-tools/blob/881ffb4682cb5882f5764aca5a56fe9865bc9ed6/pkg/webhook/parser.go#L453).

The webhook's `.webhooks[].clientConfig.service` also have their [namespace hardcoded](https://github.com/kubernetes-sigs/controller-tools/blob/881ffb4682cb5882f5764aca5a56fe9865bc9ed6/pkg/webhook/parser.go#L281C7-L282) to `system` and name to `webhook-service`, respectively.

I'd like to be able to customize these three parameters with the markers. My webhook Service might be in a different namespace and name. I might also have different webhooks that need different [Validating|Mutating]WebhookConfiguration names. It'd be nice to not have to generate these webhook YAMLs and then have to use kustomize to alter them.

Is this a feature request you're open to?

related to https://github.com/kubernetes-sigs/controller-tools/issues/490
```[tasklist]
### Tasks
- [ ] https://github.com/kubernetes-sigs/controller-tools/pull/1002
- [ ] https://github.com/kubernetes-sigs/controller-tools/pull/1128
```

davidxia:
## example usage

```
❯ GOBIN=(pwd)/bin go install ./cmd/*

❯ ./bin/controller-gen webhook -w

Webhook

+kubebuilder:webhook:admissionReviewVersions=<[]string>,failurePolicy=<string>,groups=<[]string>[,matchPolicy=<string>],mutating=<bool>,name=<string>[,path=<string>][,reinvocationPolicy=<string>],resources=<[]string>[,sideEffects=<string>][,timeoutSeconds=<int>][,url=<string>],verbs=<[]string>,versions=<[]string>[,webhookVersions=<[]string>]  package  specifies how a webhook should be served.  
+kubebuilder:webhookconfiguration:mutating=<bool>[,name=<string>]                                                                                                                                                                                                                                                                                        package  specifies how a webhook should be served.  
```

contributes to https://github.com/kubernetes-sigs/controller-tools/issues/865

### DIFF
--- a/pkg/webhook/parser.go
+++ b/pkg/webhook/parser.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-tools/pkg/genall"
 	"sigs.k8s.io/controller-tools/pkg/markers"
 )
@@ -42,14 +43,29 @@ const (
 )
 
 var (
-	// ConfigDefinition s a marker for defining Webhook manifests.
+	// ConfigDefinition is a marker for defining Webhook manifests.
 	// Call ToWebhook on the value to get a Kubernetes Webhook.
 	ConfigDefinition = markers.Must(markers.MakeDefinition("kubebuilder:webhook", markers.DescribesPackage, Config{}))
+	// WebhookConfigDefinition is a marker for defining MutatingWebhookConfiguration or ValidatingWebhookConfiguration manifests.
+	WebhookConfigDefinition = markers.Must(markers.MakeDefinition("kubebuilder:webhookconfiguration", markers.DescribesPackage, WebhookConfig{}))
 )
 
 // supportedWebhookVersions returns currently supported API version of {Mutating,Validating}WebhookConfiguration.
 func supportedWebhookVersions() []string {
 	return []string{defaultWebhookVersion}
+}
+
+// +controllertools:marker:generateHelp
+
+type WebhookConfig struct {
+	// Mutating marks this as a mutating webhook (it's validating only if false)
+	//
+	// Mutating webhooks are allowed to change the object in their response,
+	// and are called *before* all validating webhooks.  Mutating webhooks may
+	// choose to reject an object, similarly to a validating webhook.
+	Mutating bool
+	// Name indicates the name of the K8s MutatingWebhookConfiguration or ValidatingWebhookConfiguration object.
+	Name string `marker:"name,optional"`
 }
 
 // +controllertools:marker:generateHelp:category=Webhook
@@ -152,6 +168,32 @@ func verbToAPIVariant(verbRaw string) admissionregv1.OperationType {
 	default:
 		return admissionregv1.OperationType(verbRaw)
 	}
+}
+
+// ToMutatingWebhookConfiguration converts this WebhookConfig to its Kubernetes API form.
+func (c WebhookConfig) ToMutatingWebhookConfiguration() (admissionregv1.MutatingWebhookConfiguration, error) {
+	if !c.Mutating {
+		return admissionregv1.MutatingWebhookConfiguration{}, fmt.Errorf("%s is a validating webhook", c.Name)
+	}
+
+	return admissionregv1.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: c.Name,
+		},
+	}, nil
+}
+
+// ToValidatingWebhookConfiguration converts this WebhookConfig to its Kubernetes API form.
+func (c WebhookConfig) ToValidatingWebhookConfiguration() (admissionregv1.ValidatingWebhookConfiguration, error) {
+	if c.Mutating {
+		return admissionregv1.ValidatingWebhookConfiguration{}, fmt.Errorf("%s is a mutating webhook", c.Name)
+	}
+
+	return admissionregv1.ValidatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: c.Name,
+		},
+	}, nil
 }
 
 // ToMutatingWebhook converts this rule to its Kubernetes API form.
@@ -362,7 +404,11 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
 	if err := into.Register(ConfigDefinition); err != nil {
 		return err
 	}
+	if err := into.Register(WebhookConfigDefinition); err != nil {
+		return err
+	}
 	into.AddHelp(ConfigDefinition, Config{}.Help())
+	into.AddHelp(WebhookConfigDefinition, Config{}.Help())
 	return nil
 }
 
@@ -370,10 +416,41 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	supportedWebhookVersions := supportedWebhookVersions()
 	mutatingCfgs := make(map[string][]admissionregv1.MutatingWebhook, len(supportedWebhookVersions))
 	validatingCfgs := make(map[string][]admissionregv1.ValidatingWebhook, len(supportedWebhookVersions))
+	var mutatingWebhookCfgs admissionregv1.MutatingWebhookConfiguration
+	var validatingWebhookCfgs admissionregv1.ValidatingWebhookConfiguration
+
 	for _, root := range ctx.Roots {
 		markerSet, err := markers.PackageMarkers(ctx.Collector, root)
 		if err != nil {
 			root.AddError(err)
+		}
+
+		webhookCfgs := markerSet[WebhookConfigDefinition.Name]
+		var hasValidatingWebhookConfig, hasMutatingWebhookConfig bool = false, false
+		for _, webhookCfg := range webhookCfgs {
+			webhookCfg := webhookCfg.(WebhookConfig)
+
+			if webhookCfg.Mutating {
+				if hasMutatingWebhookConfig {
+					return fmt.Errorf("duplicate mutating %s with name %s", WebhookConfigDefinition.Name, webhookCfg.Name)
+				}
+
+				if mutatingWebhookCfgs, err = webhookCfg.ToMutatingWebhookConfiguration(); err != nil {
+					return err
+				}
+
+				hasMutatingWebhookConfig = true
+			} else {
+				if hasValidatingWebhookConfig {
+					return fmt.Errorf("duplicate validating %s with name %s", WebhookConfigDefinition.Name, webhookCfg.Name)
+				}
+
+				if validatingWebhookCfgs, err = webhookCfg.ToValidatingWebhookConfiguration(); err != nil {
+					return err
+				}
+
+				hasValidatingWebhookConfig = true
+			}
 		}
 
 		cfgs := markerSet[ConfigDefinition.Name]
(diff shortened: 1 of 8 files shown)


davidxia:
@sbueringer thanks for [your tip](https://github.com/kubernetes-sigs/controller-tools/issues/865#issuecomment-2219636396). How does this look? If this looks good overall, I can write a test for it. Should the test be [similar to this](https://github.com/kubernetes-sigs/controller-tools/pull/824/files#diff-ebe2f285c3166a90e416bf40390a511d3ccb7b41b12181a3efcca518ac5942c3)?

pkg/webhook/parser.go
```diff
@@ -102,6 +102,9 @@ type Config struct {
 	// Name indicates the name of this webhook configuration. Should be a domain with at least three segments separated by dots
 	Name string
 
+	// K8sName indicates the K8s name of this webhook configuration.
+	K8sName string `marker:"k8sName,optional"`
```
davidxia, line 106: open to naming suggestions

davidxia (commented):
tested manually by

1. `GOBIN=(pwd)/bin go install ./cmd/*`
2. copy the `./bin/controller-gen` executable into a kubebuilder-scaffolded project's `bin/controller-gen-vX.Y.Z` path
3. remove the `kubectl apply` from the kubebuilder project's `Makefile`'s `deploy` target
4. add `k8sName=DEADBEEF` to `+kubebuilder:webhook` marker in `api/v1/cronjob_webhook.go` (after having done [Implementing defaulting/validating webhooks](https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation#implementing-defaultingvalidating-webhooks))
5. `make deploy` and check that `kind: MutatingWebhookConfiguration` has the following.
   ```
   apiVersion: admissionregistration.k8s.io/v1
   kind: MutatingWebhookConfiguration
   metadata:
     name: project-DEADBEEF
   ```

   instead of
   ```
   apiVersion: admissionregistration.k8s.io/v1
   kind: MutatingWebhookConfiguration
   metadata:
     name: project-mutating-webhook-configuration
   ```

sbueringer:
> @sbueringer thanks for [your tip](https://github.com/kubernetes-sigs/controller-tools/issues/865#issuecomment-2219636396). How does this look? If this looks good overall, I can write a test for it. Should the test be [similar to this](https://github.com/kubernetes-sigs/controller-tools/pull/824/files#diff-ebe2f285c3166a90e416bf40390a511d3ccb7b41b12181a3efcca518ac5942c3)?

Yup. I think something similar to the test coverage in this PR would be good

Oh I think I missed something before. The existing kubebuilder:webhook marker is used to define individual webhooks in the object. While the name is global for the entire MutatingWebhookConfiguration / ValidatingWebhookConfiguration configuration. Might be a bit awkward to define the name of the overall obect on the marker that defines a single element in the webhook array.

WDYT about introducing a separate marker called `kubebuilder:webhookconfiguration` with a name field instead? 

Looking at the issue, the service namespace (& maybe also name?) can be added to the Config struct though (just like Path)

pkg/webhook/parser.go
```diff
@@ -362,20 +404,54 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
+			} else {
+				w, err := webhookCfg.ToValidatingWebhookConfiguration()
+				if err != nil {
+					return err
+				}
+				for _, webhookVersion := range supportedWebhookVersions {
+					validatingWebhookCfgs[webhookVersion] = w
+				}
+			}
+		}
```
davidxia, line 453: copied the same pattern as below

pkg/webhook/parser.go
```diff
@@ -362,20 +404,54 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
+		for _, webhookCfg := range webhookCfgs {
+			webhookCfg := webhookCfg.(WebhookConfig)
+
+			if webhookCfg.Mutating {
+				w, err := webhookCfg.ToMutatingWebhookConfiguration()
+				if err != nil {
+					return err
+				}
+				for _, webhookVersion := range supportedWebhookVersions {
+					mutatingWebhookCfgs[webhookVersion] = w
```
davidxia, line 442: Last one wins instead of appending to an array because there'll only be one MutatingWebhookConfiguration and one ValidatingWebhookConfiguration per supported version.

pkg/webhook/parser.go
```diff
@@ -362,20 +404,54 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
 	into.AddHelp(ConfigDefinition, Config{}.Help())
+	into.AddHelp(WebhookConfigDefinition, Config{}.Help())
 	return nil
 }
 
 func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	supportedWebhookVersions := supportedWebhookVersions()
 	mutatingCfgs := make(map[string][]admissionregv1.MutatingWebhook, len(supportedWebhookVersions))
 	validatingCfgs := make(map[string][]admissionregv1.ValidatingWebhook, len(supportedWebhookVersions))
+	mutatingWebhookCfgs := make(map[string]admissionregv1.MutatingWebhookConfiguration, len(supportedWebhookVersions))
```
davidxia, line 419: map where values are the object instead of array or objects since there's only one MutatingWebhookConfiguration and one ValidatingWebhookConfiguration per supported version

pkg/webhook/parser.go
```diff
@@ -410,16 +486,22 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
-			objRaw := &admissionregv1.MutatingWebhookConfiguration{}
+			var objRaw *admissionregv1.MutatingWebhookConfiguration
+			if mutatingWebhookCfg, ok := mutatingWebhookCfgs[version]; ok {
+				objRaw = &mutatingWebhookCfg
+			} else {
+				// The only possible version in supportedWebhookVersions is v1,
+				// so use it for all versioned types in this context.
+				objRaw = &admissionregv1.MutatingWebhookConfiguration{}
+				objRaw.SetName("mutating-webhook-configuration")
+			}
```
davidxia, line 497: for backwards-compat

pkg/webhook/parser.go
```diff
@@ -441,16 +523,22 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
...
-			objRaw := &admissionregv1.ValidatingWebhookConfiguration{}
+			var objRaw *admissionregv1.ValidatingWebhookConfiguration
+			if validatingWebhookCfg, ok := validatingWebhookCfgs[version]; ok {
+				objRaw = &validatingWebhookCfg
+			} else {
+				// The only possible version in supportedWebhookVersions is v1,
+				// so use it for all versioned types in this context.
+				objRaw = &admissionregv1.ValidatingWebhookConfiguration{}
+				objRaw.SetName("validating-webhook-configuration")
+			}
```
davidxia, line 534: for backwards-compat

davidxia:
> WDYT about introducing a separate marker called `kubebuilder:webhookconfiguration` with a name field instead?

done, PTAL and let me know if that looks good

> Looking at the issue, the service namespace (& maybe also name?) can be added to the Config struct though (just like Path)

Good idea. I can make another PR for that once we get something merged for this.

davidxia:
@joelanford @vincepri do you have time to give this a quick initial review?

pkg/webhook/parser.go
```diff
@@ -42,16 +43,31 @@ const (
 )
 
 var (
-	// ConfigDefinition s a marker for defining Webhook manifests.
+	// ConfigDefinition is a marker for defining Webhook manifests.
 	// Call ToWebhook on the value to get a Kubernetes Webhook.
 	ConfigDefinition = markers.Must(markers.MakeDefinition("kubebuilder:webhook", markers.DescribesPackage, Config{}))
+	// WebhookConfigDefinition is a marker for defining MutatingWebhookConfiguration or ValidatingWebhookConfiguration manifests.
+	WebhookConfigDefinition = markers.Must(markers.MakeDefinition("kubebuilder:webhookconfiguration", markers.DescribesPackage, WebhookConfig{}))
```
sbueringer, line 50: Let's add test coverage for this (see webhook/testdata folder and let's add a new simple case that also uses this marker)

pkg/webhook/parser.go
```diff
@@ -362,20 +404,54 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
+	into.AddHelp(WebhookConfigDefinition, Config{}.Help())
 	return nil
 }
 
 func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	supportedWebhookVersions := supportedWebhookVersions()
 	mutatingCfgs := make(map[string][]admissionregv1.MutatingWebhook, len(supportedWebhookVersions))
 	validatingCfgs := make(map[string][]admissionregv1.ValidatingWebhook, len(supportedWebhookVersions))
+	mutatingWebhookCfgs := make(map[string]admissionregv1.MutatingWebhookConfiguration, len(supportedWebhookVersions))
+	validatingWebhookCfgs := make(map[string]admissionregv1.ValidatingWebhookConfiguration, len(supportedWebhookVersions))
```
sbueringer, line 420: len(supportedWebhookVersions) probably doesn't make sense here

pkg/webhook/parser.go
sbueringer: Q: This whole thing today only supports to generate one MutatingWebhookConfiguration and ValidatingWebhookConfiguration? (even before your change, I"m pretty sure just want to double check)

sbueringer (commented):
Sorry for the delay, looks good so far!

pkg/webhook/parser.go
```diff
@@ -362,20 +404,54 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
...
+
 	for _, root := range ctx.Roots {
 		markerSet, err := markers.PackageMarkers(ctx.Collector, root)
 		if err != nil {
 			root.AddError(err)
 		}
 
+		webhookCfgs := markerSet[WebhookConfigDefinition.Name]
```
sbueringer, line 428: WDYT about returning an error if we find more than 1 of the "kubebuilder:webhookconfiguration" markers? (one each for mutating / validating of course)

davidxia: sgtm

pkg/webhook/parser.go
davidxia: yes, I think so

pkg/webhook/parser.go
davidxia: done

pkg/webhook/parser.go
davidxia: done

davidxia:
@sbueringer updated. Lmk if I'm missing anything!

sbueringer:
@davidxia Looks good. PTAL at https://github.com/kubernetes-sigs/controller-tools/pull/1002#issuecomment-2341625931

Otherwise happy to merge

davidxia:
@sbueringer thanks, but the bot message doesn't say why the commit message is invalid or how to fix. Are there instructions?

sbueringer:
I would just drop everything that contains fixes or an issue number in the entire commit message

davidxia:
@sbueringer that did the trick. Ready for review and/or merge now.

sbueringer:
Thank you!

/lgtm
/approve

PR #1006, opened 2024-07-15
:sparkles: Handle word boundaries and add ellipsis for `MaxDescLen`

dhaiducek:
Both of the items here are noted in a TODO in the code:
- Implement word boundary handling
- Add ellipsis after truncated text that doesn't end in a sentence terminator.
- (not in TODO) Skips processing if the description is already less than `MaxDescLen`

Outstanding questions about the initial implementation:
- Whether the ellipsis should be included in the `MaxDescLen`--in the implementation I've added it on to the end, so the total length could technically be more than `MaxDescLen` in some cases.
- Whether an ellipsis should be present after the sentence terminator (this implementation skips the ellipsis in this case).

### DIFF
--- a/pkg/crd/desc_visitor.go
+++ b/pkg/crd/desc_visitor.go
@@ -33,7 +33,7 @@ func TruncateDescription(schema *apiext.JSONSchemaProps, maxLen int) {
 // descVisitor recursively visits all fields in the schema and truncates the
 // description of the fields to specified maxLen.
 type descVisitor struct {
-	// maxLen is the maximum allowed length for decription of a field
+	// maxLen is the maximum allowed length for description of a field
 	maxLen int
 }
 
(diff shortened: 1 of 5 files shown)


alvaroaleman (commented):
Thanks for your contribution. Could you please extend pkg/crd/testdata/cronjob_types.go to demonstrate the change in behavior end to end?

dhaiducek:
Sorry for the churn here--inadvertently added a commit message from a different repo. I've reset the commit and will add some integration tests in a subsequent commit. 😬

dhaiducek:
> ... extend `pkg/crd/testdata/cronjob_types.go` ...

Hi @alvaroaleman! I've updated the `gen_integration_test.go` instead. It looked like the `cronjob_types.go` was for the CRD itself whereas the `MaxDescLen` is a generation configuration that needed a test at a higher level? If I'm mistaken, let me know!

alvaroaleman (approved):
thanks!

PR #1015, opened 2024-07-30
:warning: Add support for encoding.TextMarshaler

twz123:
Whenever a type is encountered that implements `encoding.TextMarshaler` but not `encoding/json.Marshaler`, assume that it will be encoded as a string.

This is a breaking change, as types that implement `TextMarshaler` are now handled differently. On the other hand, this probably restores reality for all but the most customized setups, as Go's JSON package will also check for the presence of this interface, and output a JSON string.

### DIFF
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -27,6 +27,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"strconv"
 	"time"
 
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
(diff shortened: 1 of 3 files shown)


pkg/crd/testdata/cronjob_types.go
```diff
@@ -552,6 +553,30 @@ func (u *URL2) String() string {
...
+
+// MarshalText implements [encoding.TextMarshaler].
+func (u *URL3) MarshalText() (text []byte, err error) {
+	return u.MarshalBinary()
+}
+
+// URL4 is an alias of [net/url.URL]. It implements [encoding.TextMarshaler] so
+// that it can be used in K8s CRDs such that the CRD resource will have the URL
+// but operator code can can work with the URL struct.
+type URL4 url.URL
```
alvaroaleman, line 571: I do not understand the difference between the two, could you explain why they both exist?

What seems to be missing as a testcase is something that implements both json marshaller and text marshaller, in that case the json marshalling must take precedence

alvaroaleman (commented):
/ok-to-test

Thanks for your contribution, the change makes sense. One comment re testing

pkg/crd/testdata/cronjob_types.go
twz123: > I do not understand the difference between the two, could you explain why they both exist?

It's been a while since I wore that piece of code. Need to figure out myself 🙃

> What seems to be missing as a testcase is something that implements both json marshaller and text marshaller, in that case the json marshalling must take precedence

Will do a rebase and add the test in the next few days I hope.

twz123: > I do not understand the difference between the two, could you explain why they both exist?

You mean `URL4` and `URL4`? I mimicked the `URL` and `URL2` types. One is a regular struct, the other is a newtype. I guess it makes sense to have both, as they might behave differently when using reflection?

> What seems to be missing as a testcase is something that implements both json marshaller and text marshaller, in that case the json marshalling must take precedence

Added.

alvaroaleman (approved):
Thanks!

PR #1026, opened 2024-08-09
✨ Publish controller-gen binaries on release

### LINKED ISSUE #500: Any plans to support binary releases for `controller-gen`?
First, want to say thanks for the great work on the controller-tools repo. It's generally awesome. :)

I work on a project (github.com/aws/aws-controllers-k8s) that makes heavy use of `controller-gen` to augment our own code generation. However, because there are no binary releases of `controller-gen` we've seen one issue crop up.

Since the only way to use the controller-gen tool is to "install" it with `go get`, if a contributor to our project doesn't have `controller-tools` locally and they run `go get` to install it, that invariably ends up modifying the `go.mod/sum` file in our source repository. These are `// indirect` entries in the `go.mod` but still, it's kind of annoying to have to tell contributors to undo the changes to their `go.mod/sum` files due to this.

I was wondering if there are any plans to produce binary artifacts for the `controller-gen` tool? This would certainly make our lives a bit easier on the downstream consumer side of things.

If not binary artifacts, perhaps publishing Docker images containing pre-built `controller-gen` binaries?

sbueringer:

This PR adds a GitHub action which will build & upload binaries whenever a tag/release is created
(this requires no changes to our release workflow, simply creating a GitHub release still works)

Example: https://github.com/sbueringer/controller-tools/releases/tag/v0.16.0-beta.0

Fixes #500

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -23,12 +23,14 @@ jobs:
         working-directory:
           - ""
     steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # tag=v4.1.7
+      - name: Calculate go version
+        id: vars
+        run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
       - name: Set up Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # tag=v5.0.2
         with:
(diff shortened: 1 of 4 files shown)


sbueringer:
cc @camilamacedo86 (with this it should be easy to just download the setup-envtest binaries instead of building them locally)

chrischdi (commented):
/lgtm

PR #1044, opened 2024-08-29
🐛 rbac: fix adding nonResourceURLs including normalisation

chrischdi:
Fixes the rbac generator for nonResourceURLs / when using `kubebuilder:rbac:urls` marker.

Kudos to @rakesh-garimella for reporting it at https://github.com/kubernetes-sigs/controller-tools/pull/937

This propably broke when implementing #937

@sbueringer : I think this is ok to cherry-pick?

### DIFF
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -105,6 +105,12 @@ func (r *Rule) keyWithResourcesResourceNamesURLsVerbs() string {
 	return fmt.Sprintf("%s + %s + %s + %s", key.Resources, key.ResourceNames, key.URLs, verbs)
 }
 
+func (r *Rule) keyWitGroupResourcesResourceNamesVerbs() string {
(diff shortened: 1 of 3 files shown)


sbueringer:
Thx!

/lgtm
/approve

/test pull-controller-tools-test-master

sbueringer:
@rakesh-garimella Do you have some time to verify the fix on your side as well? I would then cut a new patch release

sbueringer:
Just remembered that we don't always use release branches in CT, which is also the reason why we don't have one yet.

I would also cut this patch release directly from main, because apart from a few bug fixes nothing was changed on CT main since v0.16 & v0.16.1.

rakesh-garimella:
The fix works for me, thanks again @chrischdi @sbueringer

sbueringer:
Thx again for reporting!

Issue #1036, opened 2024-08-15
Core group can be used as "" or "core" but deduplication doesn't take that into account


pmalek:
### Problem statement

0.16.0 introduced deduplication in #937.

Since core can be referred to by `""` or `"core"` this should be taken into account but it isn't causing 2 apigroups to be added for the same core group.

### How to reproduce

Have 2 different but logically identical codemarkers:

```
//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
```

and 

```
// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
```

Get the following generated with 0.16.1:

```
rules:
- apiGroups:
  - ""
  - ""
  resources:
  - events
  verbs:
  - create
  - patch
```

The workaround is to use the same apigroup everywhere.

sbueringer:
Thx!

/help

PR #1045, opened 2024-08-29
🐛 rbac: fix deduplication of core group and add test coverage

chrischdi:
Fixes: #1036

Deduplication happened before renaming.

Now renaming before doing deduplication to fix this case for the core group.

### DIFF
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -146,12 +146,6 @@ func removeDupAndSort(strs []string) []string {
 
 // ToRule converts this rule to its Kubernetes API form.
 func (r *Rule) ToRule() rbacv1.PolicyRule {
-	// fix the group names first, since letting people type "core" is nice
-	for i, group := range r.Groups {
-		if group == "core" {
-			r.Groups[i] = ""
-		}
-	}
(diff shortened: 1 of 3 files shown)


pkg/rbac/parser.go
```diff
@@ -230,6 +224,11 @@ func GenerateRoles(ctx *genall.GenerationContext, roleName string) ([]interface{
 		ruleMap := make(map[ruleKey]*Rule)
 		// all the Rules having the same ruleKey will be merged into the first Rule
 		for _, rule := range rules {
+			// fix the group name first, since letting people type "core" is nice
+			if len(rule.Groups) == 1 && rule.Groups[0] == "core" {
```
sbueringer, line 228: Why only if len(rule.Groups) == 1 instead of for all r.Groups like before?

chrischdi:
/test pull-controller-tools-test-master

Dunno why it did not run

chrischdi:
weird, why is there a `pull-controller-tools-test-master`, I guess it was moved to main during the pr

sbueringer:
Thank you very much!!

/lgtm
/approve

Issue #1087, opened 2024-11-05
Bug: Plural for words end with "ea"


vietanhtwdk:
Any crd with kind end with "ea" has wrong plural form. This is because the "github.com/gobuffalo/flect" package so either fix from this package or replace it with other lib like "github.com/jinzhu/inflection" will fix it.

Sample code:
```go
package main

import (
	"fmt"
	"strings"

	"github.com/gobuffalo/flect"
	"github.com/jinzhu/inflection"
)

func main() {
	fmt.Println(strings.ToLower(flect.Pluralize("area")))
	fmt.Println(strings.ToLower(inflection.Plural("area")))
}
```
Output:
```
area
areas
```

sbueringer:
/close

See: https://github.com/kubernetes-sigs/controller-tools/pull/1090#discussion_r1835633577

PR #1090, opened 2024-11-10
🐛 Fix plural

vietanhtwdk:
🐛Fix plural

Fix: https://github.com/kubernetes-sigs/controller-tools/issues/1087

### DIFF
--- a/pkg/typescaffold/resource.go
+++ b/pkg/typescaffold/resource.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/gobuffalo/flect"
+	"github.com/jinzhu/inflection"
 )
 
 // Resource contains the information required to scaffold files for a resource.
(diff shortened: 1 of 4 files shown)


pkg/typescaffold/resource.go
```diff
@@ -42,7 +43,7 @@ func (r *Resource) Validate() error {
 	}
 
 	if len(r.Resource) == 0 {
-		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
+		r.Resource = inflection.Plural(strings.ToLower(r.Kind))
```
camilamacedo86, line 46: It can broke the projects built so far and kubebuilder,
see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/model/resource/utils.go#L59-L62

camilamacedo86: /hold

sbueringer: Agree. We can't make a change that leads to any different names here

camilamacedo86:
I think we should close this one as referred.
based on : https://github.com/kubernetes-sigs/controller-tools/pull/1090#discussion_r1835633577

WDYT?

@vietanhtwdk @sbueringer

sbueringer:
Agree

/close

PR #415, opened 2020-03-10
🐛Fix panic on anonymous nested struct in CRD

hypnoglow:
Anonymous nested structs are not allowed according to the error message, but the code was panicking because for such struct `ctx.info.RawSpec` is nil.

This change adds extra checks so that struct name must be present and spec must not be nil.

Fixes https://github.com/kubernetes-sigs/controller-tools/issues/340 (which was closed recently by bot)

---

Not sure if it is possible to add a test for this somewhere 🤔. I tested locally:

```go
type ZugaSpec struct {
	NestedStruct ZugaNestedStruct `json:"nestedStruct,omitempty"`

	NestedAnonymousStruct struct {
		Bar string `json:"bar"`
	} `json:"nestedAnonymousStruct"`
}

type ZugaNestedStruct struct {
	Foo string `json:"foo"`
}
```

Running `make manifests` on master:

```
λ make manifests
/Users/hypnoglow/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x16678cc]

goroutine 1 [running]:
sigs.k8s.io/controller-tools/pkg/crd.structToSchema(0xc001263440, 0xc0003e4460, 0x0)
	/Users/hypnoglow/sources/github.com/kubernetes-sigs/controller-tools/pkg/crd/schema.go:322 +0x9c
sigs.k8s.io/controller-tools/pkg/crd.typeToSchema(0xc001263440, 0x19aaf60, 0xc0003e4460, 0x0)
	/Users/hypnoglow/sources/github.com/kubernetes-sigs/controller-tools/pkg/crd/schema.go:164 +0x3a6
sigs.k8s.io/controller-tools/pkg/crd.structToSchema(0xc000a860b0, 0xc0003e44a0, 0x0)
	/Users/hypnoglow/sources/github.com/kubernetes-sigs/controller-tools/pkg/crd/schema.go:375 +0x491
sigs.k8s.io/controller-tools/pkg/crd.typeToSchema(0xc000a860b0, 0x19aaf60, 0xc0003e44a0, 0x0)
	/Users/hypnoglow/sources/github.com/kubernetes-sigs/controller-tools/pkg/crd/schema.go:164 +0x3a6
sigs.k8s.io/controller-tools/pkg/crd.infoToSchema(...)
[...]

Issue #340, opened 2019-10-11
Nested anonymous structs cause panic when generating CRDs


jameshochadel:
This issue was found while using `controller-tools` with Kubebuilder. 

**To reproduce:**

Create a new Kubebuilder v2 project:

```sh
go mod init example.com/m
kubebuilder init --domain example.com --license apache2 --owner "Test Owner"
kubebuilder create api --group ship --version v1 --kind Frigate
```

In `./api/v1/frigate_types.go`, modify `FrigateSpec` to be:

```go
type FrigateSpec struct {
	NestedAnonymousStruct struct {
		SomeField string `json:"someField"`
	} `json:"nestedAnonymousStruct"`
}
```

And run `make manifests`.

_Expected behavior_: `controller-get` is fetched, and the CRDs are generated.
_Actual behavior_: I get the following panic:

```
go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.1
go: finding sigs.k8s.io v0.2.1
go: finding sigs.k8s.io/controller-tools/cmd v0.2.1
go: finding sigs.k8s.io/controller-tools/cmd/controller-gen v0.2.1
/Users/james/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x15e6f4c]

goroutine 1 [running]:
sigs.k8s.io/controller-tools/pkg/crd.structToSchema(0xc000f9cfc0, 0xc0006d4200, 0x0)
        /Users/james/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.1/pkg/crd/schema.go:322 +0x9c
sigs.k8s.io/controller-tools/pkg/crd.typeToSchema(0xc000f9cfc0, 0x18f9420, 0xc0006d4200, 0x0)
        /Users/james/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.1/pkg/crd/schema.go:164 +0x3a6
sigs.k8s.io/controller-tools/pkg/crd.structToSchema(0xc0007ae1e8, 0xc0006d42a0, 0x0)
        /Users/james/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.1/pkg/crd/schema.go:375 +0x491
[...]

PR #1098, opened 2024-11-28
🐛 Fix SIGSERV when RawSpec is nil (due to anonymous struct)

rkettelerij:
This is the same PR as https://github.com/kubernetes-sigs/controller-tools/pull/892 but that one was automatically closed. This time testdata is included as requested so I'm hoping to get this merged.

From the original PR:

> Under some circumstances RawSpec can be nil. I've encountred this when generating a CRD of a complex set of structs. Currently the structToSchema func fails hard on this (SIGSERV). This PR fixes the latter.

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -404,7 +404,7 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiext.JSON
 		Properties: make(map[string]apiext.JSONSchemaProps),
 	}
 
-	if ctx.info.RawSpec.Type != structType {
+	if ctx.info.RawSpec != nil && ctx.info.RawSpec.Type != structType {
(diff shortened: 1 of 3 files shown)


rkettelerij:
Apparently this was also reported before in #340 and #415 but never merged due to staleness.

pkg/crd/schema.go
```diff
@@ -404,7 +404,7 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiext.JSON
 		Properties: make(map[string]apiext.JSONSchemaProps),
 	}
 
-	if ctx.info.RawSpec.Type != structType {
+	if ctx.info.RawSpec != nil && ctx.info.RawSpec.Type != structType {
```
chrischdi, line 407: But this would allow embedded structs / anonymous structs, which does not seem to be allowed according to the error below?

Shouldn't this then be:

```suggestion
	if ctx.info.RawSpec == nil || ctx.info.RawSpec.Type != structType {
```

PR #1118, opened 2025-01-03
🌱 Allow string validation on XIntOrString

cbandy:
This allows `maxLength` and `MinLength` on `x-kubernetes-int-or-string` fields. The former helps limit CEL validation costs.

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -316,6 +316,10 @@ func hasNumericType(schema *apiext.JSONSchemaProps) bool {
 	return schema.Type == "integer" || schema.Type == "number"
 }
 
+func hasTextualType(schema *apiext.JSONSchemaProps) bool {
+	return schema.Type == "string" || schema.XIntOrString
+}
+
 func isIntegral(value float64) bool {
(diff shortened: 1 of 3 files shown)


JoelSpeed (commented):
This seems reasonable, thanks!

One question, if I were to use this feature, and provide a value that is a long integer, does the length of my integer, say 12 characters, also get validated with the same maximum length constraints? I assume the answer is yes but I'm not sure I see how the tests might demonstrate that this works

cbandy:
> One question, if I were to use this feature, and provide a value that is a long integer, does the length of my integer, say 12 characters, also get validated with the same maximum length constraints? I assume the answer is yes but I'm not sure I see how the tests might demonstrate that this works

In my testing with 1.31 and 1.32, the answer is no, `pattern` and `maxLength` do not affect integer values.

Go markers affect the YAML this project generates, and the Kubernetes API interprets that.

sbueringer:
Thx!

/ok-to-test
/lgtm
/approve

cbandy:
> The former helps limit CEL validation costs.

😅 Except that it didn't until https://github.com/kubernetes/kubernetes/pull/132837

PR #1119, opened 2025-01-04
🐛 Guard before type assertion

### LINKED ISSUE #813: Basic type alias panic on shouldBeCopied()
When generating `DeepCopy`, `DeepCopyInto` and `DeepCopyObject`, a panic occurs with the following error:

```
panic: interface conversion: types.Type is *types.Basic, not *types.Named

goroutine 1 [running]:
sigs.k8s.io/controller-tools/pkg/deepcopy.shouldBeCopied(0x140002584c0, 0x14002327080)
	/Users/me/go/pkg/mod/sigs.k8s.io/controller-tools@v0.12.0/pkg/deepcopy/traverse.go:622
```

This occurs when using an alias with a basic type:

```
type MyAlias = string
```

Things pass when using the following:
```
type MyAlias string
```

While checking the [internals](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/deepcopy/traverse.go#L609-L650) it _seems_ like the intention is to skip types that are _eventually_ basic but nothing about types that are "immediately" basic.

I wanted to propose adding a check for if `typeInfo` is a basic type before getting the underlying pointer of the named type to support an alias.

```
func shouldBeCopied(pkg *loader.Package, info *markers.TypeInfo) bool {
	if !ast.IsExported(info.Name) {
		return false
	}

	typeInfo := pkg.TypesInfo.TypeOf(info.RawSpec.Name)
	if typeInfo == types.Typ[types.Invalid] {
		pkg.AddError(loader.ErrFromNode(fmt.Errorf("unknown type: %s", info.Name), info.RawSpec))
		return false
	}

+	if _, isBasic := typeInfo.(*types.Basic); isBasic {
+		return false
+	}

	// according to gengo, everything named is an alias, except for an alias to a pointer,
	// which is just a pointer, afaict.  Just roll with it.
	if asPtr, isPtr := typeInfo.(*types.Named).Underlying().(*types.Pointer); isPtr {
		typeInfo = asPtr
	}

	lastType := typeInfo
	if _, isNamed := typeInfo.(*types.Named); isNamed {
		// if it has a manual deepcopy or deepcopyinto, we're fine
		if hasAnyDeepCopyMethod(pkg, typeInfo) {
			return true
		}

		for underlyingType := typeInfo.Underlying(); underlyingType != lastType; lastType, underlyingType = underlyingType, underlyingType.Underlying() {
			// if it has a manual deepcopy or deepcopyinto, we're fine
			if hasAnyDeepCopyMethod(pkg, underlyingType) {
				return true
			}

			// aliases to other things besides basics need copy methods
			// (basics can be straight-up shallow-copied)
			if _, isBasic := underlyingType.(*types.Basic); !isBasic {
				return true
			}
		}
	}

	// structs are the only thing that's not a basic that's copiable by default
	_, isStruct := lastType.(*types.Struct)
	return isStruct
}

```

cbandy:
Without this, Go type aliases cause a panic.

Fixes #813

### DIFF
--- a/pkg/deepcopy/testdata/cronjob_types.go
+++ b/pkg/deepcopy/testdata/cronjob_types.go
@@ -153,7 +153,12 @@ type SpecificCases struct {
 
 	// Case: kubernetes-sigs/controller-tools#262 part 2 (see type definition)
 	StringMap MapOfStrings `json:"stringMap"`
+
+	// Case: kubernetes-sigs/controller-tools#813
+	StringAlias StringAlias `json:"stringAlias,omitempty"`
 }
+
+type StringAlias = string
 
 // Test aliases to basic types
 type TotallyAString string
--- a/pkg/deepcopy/traverse.go
+++ b/pkg/deepcopy/traverse.go
@@ -618,14 +618,15 @@ func shouldBeCopied(pkg *loader.Package, info *markers.TypeInfo) bool {
 		return false
 	}
 
-	// according to gengo, everything named is an alias, except for an alias to a pointer,
-	// which is just a pointer, afaict.  Just roll with it.
-	if asPtr, isPtr := typeInfo.(*types.Named).Underlying().(*types.Pointer); isPtr {
-		typeInfo = asPtr
-	}
-
 	lastType := typeInfo
 	if _, isNamed := typeInfo.(*types.Named); isNamed {
+		// according to gengo, everything named is an alias, except for an alias to a pointer,
+		// which is just a pointer, afaict.  Just roll with it.
+		if asPtr, isPtr := typeInfo.(*types.Named).Underlying().(*types.Pointer); isPtr {
+			lastType = asPtr
+			typeInfo = asPtr
+		}
+
 		// if it has a manual deepcopy or deepcopyinto, we're fine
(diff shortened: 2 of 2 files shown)


pkg/deepcopy/testdata/cronjob_types.go
```diff
@@ -153,8 +153,13 @@ type SpecificCases struct {
 
 	// Case: kubernetes-sigs/controller-tools#262 part 2 (see type definition)
 	StringMap MapOfStrings `json:"stringMap"`
+
```
sbueringer, line 156: I'm a bit confused that this addition doesn't lead to any necessary change in testdata

cbandy: IIUC, it is covered by this existing assignment because it is a string field.

https://github.com/kubernetes-sigs/controller-tools/blob/12ea13fd95e06878edb3b2febb5674e05bc73151/pkg/deepcopy/testdata/zz_generated.deepcopy.go#L572-L573

sbueringer: Makes sense

sbueringer:
Thank you!

/lgtm
/approve

PR #1120, opened 2025-01-04
🌱 Test DeepCopy of contained non-pointer types

cbandy:
This seems like a copy-pasta from long ago. The name of the test/field is `NonPtr` but the type is `Ptr` like the test/field above it.


sbueringer:
Thx!

/lgtm
/approve

PR #1132, opened 2025-01-22
🌱Bump golangci-lint to v1.63.4

dongjiang1989:
Bump golangci-lint to v1.63.4

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,19 @@ run:
   # while if uncached!
   timeout: 10m
 
+linters:
+  enable:
+  - errcheck
+  - gosimple
+  - govet
+  - ineffassign
+  - staticcheck
+  - unused
+  - testifylint
+  - iface
+  - exptostd
+  - nilnesserr
+
 issues:
   exclude-rules:
     # Dot imports for gomega and ginkgo are allowed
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -4,6 +4,19 @@ run:
   # while if uncached!
   timeout: 10m
 
+linters:
+  enable:
```
sbueringer, line 8: This seems like a short list. How did we decide which ones to enable?

Wondering if we should try to enable the same as in CR (potentially split up in multiple PRs if there would be a lot of findings)

dongjiang1989: Thanks @sbueringer 

In my opinion it should be possible to enable all linters. 

Under the previous default settings, `.golangci.yaml` only enabled `[errcheck gosimple govet ineffassign staticcheck unused] ` linter.
image

I'll add other linters for next PR.

sbueringer: Sounds good, thx!

> In my opinion it should be possible to enable all linters.

Just to avoid misunderstandings. I would only enable the ones enabled in controller-runtime (vs. all :))

sbueringer:
/lgtm
/approve

Thank you!

.golangci.yml
dongjiang1989: > Just to avoid misunderstandings. I would only enable the ones enabled in controller-runtime (vs. all :))

Thanks!
I got it.

PR #1206, opened 2025-05-15
🌱ci: Update golangci-lint to v2.1.6 and adjust configuration

dongjiang1989:
Update golangci-lint to v2.1.6 and adjust configuration

Link #1182 

cc @sbueringer


JoelSpeed (commented):
/lgtm

Issue #1193, opened 2025-04-16
applyconfiguration generator uses a different variable name


terinjokes:
In #818 support for generating ApplyConfiguration was added via the applyconfiguration generator. However, it uses different variable name (`SchemeGroupVersion`) than the name generated by kubebuilder (`GroupVersion`). This results in a build failures out of the box.

```
# sigs.k8s.io/controller-tools/pkg/applyconfiguration/testdata/cronjob/api/v1/applyconfiguration
api/v1/applyconfiguration/utils.go:19:10: undefined: v1.SchemeGroupVersion
```

The variable name from kubebuilder is set in the `groupversion_info.go` template:

https://github.com/kubernetes-sigs/kubebuilder/blob/9932c89b631bb03413afe858c736613c66e07882/pkg/plugins/golang/v4/scaffolds/internal/templates/api/group.go#L68-L69

The variable name in the upstream applyconfiguration-gen is set in the `GenerateType` implementation:

https://github.com/kubernetes/code-generator/blob/2a635fa63960c0d6fd2461819f261ed8b9115f57/cmd/applyconfiguration-gen/generators/util.go#L127-L130

The fix for users is simple (`SchemeGroupVersion = GroupVersion`) but it would be nice if the files generated here work out of the box with those from kubebuilder.

JoelSpeed:
We have no influence over the code in code-generator that we are relying on here. However, we do have influence over the loaded `universe`. It's possible that we could detect whether the required `SchemeGroupVersion` exists, and, should it not exist, mock it up from the `GroupVersion`.

That could be a bit fragile, so we would want to also check that the types match our expectations, but at least if we put this check in earlier, we could also add a warning if we aren't able to detect the `schema.GroupVersion` that we require, and bail out of the generation earlier

alvaroaleman:
@terinjokes what is missing here? The generated applyconfigs do build for some time now

sbueringer:
I think the point of this issue is the following

Usually you would use the following in groupversion_info.go of your API:
```
	// GroupVersion is group version used to register these objects.
	GroupVersion = schema.GroupVersion{Group: "cluster.x-k8s.io", Version: "v1beta2"}
```

But generated applyconfiguration code is importing `SchemeGroupVersion`.

So the current state requires 
```
	SchemeGroupVersion = GroupVersion
```

in groupversion_info.go to make applyconfigurations compile

sbueringer:
[...]

PR #1217, opened 2025-06-18
:bug: Applyconfiguration: Use correct groupVersion

alvaroaleman:
The upstream generator we import infers the groupVersion from the package path: https://github.com/kubernetes/kubernetes/blob/b0b52f4fb29223f829b50ae58323c20a7764dfc9/staging/src/k8s.io/code-generator/cmd/applyconfiguration-gen/generators/targets.go#L263

As a result, if the filesystem layout doesn't match the group version, the result will be incorrect. Fix by updating the package path for the generator to contain the correct groupVersion.

Fixes https://github.com/kubernetes-sigs/controller-tools/issues/1193


alvaroaleman:
Superseded by https://github.com/kubernetes-sigs/controller-tools/pull/1218

PR #1218, opened 2025-06-18
:bug: Applyconfiguration: Use correct groupVersion

alvaroaleman:
The upstream generator we import infers the groupVersion from the package path: https://github.com/kubernetes/kubernetes/blob/b0b52f4fb29223f829b50ae58323c20a7764dfc9/staging/src/k8s.io/code-generator/cmd/applyconfiguration-gen/generators/targets.go#L263

The path might not be reflective of the groupVersion, so use the override mechanism: https://github.com/kubernetes/kubernetes/blob/b0b52f4fb29223f829b50ae58323c20a7764dfc9/staging/src/k8s.io/code-generator/cmd/applyconfiguration-gen/generators/targets.go#L270

Fixes https://github.com/kubernetes-sigs/controller-tools/issues/1194


sbueringer:
Thx!

/lgtm
/approve

PR #1221, opened 2025-06-20
🐛 Applyconfiguration: register CRD markers

alexymantha:
It seems the markers collector will only collect markers which have been explicitly registered to the markers registry. Since `groupName` is never registered, it always fails with
```
could not infer groupVersion for package - Is the `// +groupName` marker set?
run `controller-gen applyconfiguration paths=./api/v1beta1/ -w` to see all available markers, or `controller-gen applyconfiguration paths=./api/v1beta1/ -h` for usage
```
because the marker will not be in the map of markers despite the marker being set on the package.

This registers the CRD markers (`groupName`, `versionName`, etc) so they can be collected and used to infer the group version.

Fixes #1218

### DIFF
--- a/pkg/applyconfiguration/applyconfiguration_integration_test.go
+++ b/pkg/applyconfiguration/applyconfiguration_integration_test.go
@@ -106,11 +106,9 @@ var _ = Describe("ApplyConfiguration generation from API types", func() {
 		By("Initializing the runtime")
 		optionsRegistry := &markers.Registry{}
 		Expect(genall.RegisterOptionsMarkers(optionsRegistry)).To(Succeed())
-		Expect(optionsRegistry.Register(markers.Must(markers.MakeDefinition("crd", markers.DescribesPackage, crd.Generator{})))).To(Succeed())
 		Expect(optionsRegistry.Register(markers.Must(markers.MakeDefinition("applyconfiguration", markers.DescribesPackage, Generator{})))).To(Succeed())
 
 		rt, err := genall.FromOptions(optionsRegistry, []string{
-			"crd:allowDangerousTypes=true,ignoreUnexportedFields=true", // Run another generator first to make sure they don't interfere; see also: the comment on cronjob_types.go:UntypedBlob
 			"applyconfiguration",
 			"paths=./api/v1",
 		})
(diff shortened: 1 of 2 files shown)


pkg/applyconfiguration/gen.go
```diff
@@ -72,6 +72,10 @@ func (Generator) RegisterMarkers(into *markers.Registry) error {
 		return err
 	}
 
+	if err := crdmarkers.Register(into); err != nil {
```
alvaroaleman, line 75: Thanks. Could you please change the test in pkg/applyconfiguration/applyconfiguration_integration_test.go to use this `RegisterMarkers` rather than manually registering markers so we have a chance to find issues like this in CI?

alvaroaleman (commented):
/ok-to-test

pkg/applyconfiguration/gen.go
alexymantha: I think `RegisterMarkers` is already being called indirectly on this line: https://github.com/kubernetes-sigs/controller-tools/blob/9afcc17b313ee6acb64e840a5aefba3414a345fb/pkg/applyconfiguration/applyconfiguration_integration_test.go#L110

because the markers specific to `applyconfiguration` do not seem to be registered manually anywhere in the tests, that's the only registering I could find.

I will try to add some tests to cover this though

alexymantha: On a side note, this didn't get caught by the tests because the tests are also registering the CRD generator in the same registry: https://github.com/kubernetes-sigs/controller-tools/blob/9afcc17b313ee6acb64e840a5aefba3414a345fb/pkg/applyconfiguration/applyconfiguration_integration_test.go#L109C24-L109C25

So the necessary markers were indirectly being registered

pkg/applyconfiguration/applyconfiguration_integration_test.go
```diff
@@ -106,11 +106,9 @@ var _ = Describe("ApplyConfiguration generation from API types", func() {
 		By("Initializing the runtime")
 		optionsRegistry := &markers.Registry{}
 		Expect(genall.RegisterOptionsMarkers(optionsRegistry)).To(Succeed())
-		Expect(optionsRegistry.Register(markers.Must(markers.MakeDefinition("crd", markers.DescribesPackage, crd.Generator{})))).To(Succeed())
 		Expect(optionsRegistry.Register(markers.Must(markers.MakeDefinition("applyconfiguration", markers.DescribesPackage, Generator{})))).To(Succeed())
 
 		rt, err := genall.FromOptions(optionsRegistry, []string{
-			"crd:allowDangerousTypes=true,ignoreUnexportedFields=true", // Run another generator first to make sure they don't interfere; see also: the comment on cronjob_types.go:UntypedBlob
```
alexymantha, line 113: Split this up into another test that will run both generators. That way the tests covers both when `applyconfiguration` runs standalone and when it runs with other generators.

Issue #1219, opened 2025-06-18
Applyconfiguration testdata doesn't build


alvaroaleman:
Trying to build the applyconfiguration testdata in pkg/applyconfiguration/testdata/cronjob/api/v1/applyconfiguration/api/v1 fails with the following error:

```
api/v1/cronjobspec.go:74:2: undefined: unexportedStructApplyConfiguration
api/v1/cronjobspec.go:816:75: cannot use *values[i] (variable of struct type intstr.IntOrString) as *intstr.IntOrString value in argument to append
```

These look like bugs in the upstream AC generation, however we should get these fixed and test that this can be built.

/kind bug

sbueringer:
Yeah, would be great to have some sort of unit test that uses the generated apply configurations (at the very least to ensure they compile)

PR #1226, opened 2025-06-27
:seedling: Fix AC generation so it builds and test all submodules

alvaroaleman:
Currently, we do not test if the generated applyconfiguration builds because it is in a submodule. As a result, we didn't notice that it doesn't.

This change:
* Fixes the AC to build by commenting out the problematic fields until the generation gets fixed
* Updated our test script to run the `go test` command for all folders with a `go.mod` file to prevent this class of issue in the future

FIxes https://github.com/kubernetes-sigs/controller-tools/issues/1219


pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go
```diff
@@ -294,8 +294,10 @@ type CronJobSpec struct {
 	unexportedField string
 
 	// This tests that both unexported and exported inline fields are not skipped in the schema generation
-	unexportedStruct `json:",inline"`
-	ExportedStruct   `json:",inline"`
+	//
+	// This currently does not work
+	//unexportedStruct `json:",inline"`
```
sbueringer, line 299: Are we tracking this limitation and the one below somewhere? (e.g. in an ApplyConfiguration umbrella issue?)

pkg/applyconfiguration/testdata/cronjob/api/v1/groupversion_info.go
```diff
@@ -27,9 +27,10 @@ import (
 )
 
 var (
-	GroupName     = "testdata.kubebuilder.io"
-	GroupVersion  = schema.GroupVersion{Group: GroupName, Version: "v1"}
-	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+	GroupName          = "testdata.kubebuilder.io"
+	GroupVersion       = schema.GroupVersion{Group: GroupName, Version: "v1"}
+	SchemeGroupVersion = GroupVersion
+	SchemeBuilder      = runtime.NewSchemeBuilder(addKnownTypes)
```
sbueringer, line 33: This fixes the issue, but we don't have unit test coverage that this actually compiles, right?

pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go
alvaroaleman: No, but I will have a look into the upstream generator to see if I can fix it easily and create issues if not

pkg/applyconfiguration/testdata/cronjob/api/v1/groupversion_info.go
alvaroaleman: The `go test` command will fail if the code doesn't compile, regardless of if it has tests or not

sbueringer: Perfect

sbueringer: Looking at the change in test.sh only when we call go test directly on the directory? (`go test -race ./pkg/... ./cmd/... -parallel 4` doesn't seem to have tried to compile it before?)

sbueringer:
/lgtm
/approve

Thx!

pkg/applyconfiguration/testdata/cronjob/api/v1/groupversion_info.go
alvaroaleman: The problem is that `go test ./...` will only test packages that are part of the main go module, to test other go modules, we have to first change in the directory, hence the `find` and `cd` in the script

sbueringer: Got it, thx!

PR #1250, opened 2025-07-31
:sparkles:  Add support for k8s:enum tag.

lalitc375:
This PR introduces support for a +k8s:enum marker to automatically generate enum validation for Custom Resource Definitions (CRDs). This tag is used by native types to validate enums using declarative Validations.

This change simplifies the process of defining enums in CRDs. Previously, developers had to manually specify all enum values using the +kubebuilder:validation:Enum marker for every field of an enum type. This was repetitive and prone to errors, especially when enum values changed.

With this new +k8s:enum type-level marker, controller-gen automatically discovers all constant values associated with a string-based enum type and generates the corresponding validation in the OpenAPI schema. This makes the process more maintainable and less error-prone.

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -34,6 +34,9 @@ const (
 
 	ValidationExactlyOneOfPrefix = validationPrefix + "ExactlyOneOf"
 	ValidationAtMostOneOfPrefix  = validationPrefix + "AtMostOneOf"
+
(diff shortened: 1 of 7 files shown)


lalitc375:
Holding this PR, until we gets approval on DV.

JoelSpeed:
/reopen

/remove-lifecycle rotten

/lifecycle frozen

We should try and move this forward

alvaroaleman:
Created a rebase of this with an added test here: https://github.com/kubernetes-sigs/controller-tools/pull/1352

PR #1263, opened 2025-08-22
🐛 Update golang.org/x/tools to 0.24.1

GeorgeMac:
This updates `x/tools` package in the release/0.15 branch because currently it is not possible to install controller-tools with Go 1.25:

```console
go install golang.org/dl/go1.25.0@latest
go1.25.0 download
go1.25.0 install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.15.0
# golang.org/x/tools/internal/tokeninternal
../../../../go/pkg/mod/golang.org/x/tools@v0.20.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
```

I am not exactly sure if this is the correct way to backport an upgrade to a prior release in this project, but thought I would give it a shot.

See: https://github.com/golang/go/issues/74462


sbueringer:
Is there a reason why you can't update to a recent release of controller-tools / controller-gen? :)

GeorgeMac:
@sbueringer yes, we are going to 👍 feel free to close / ignore. I supposed someone else might have a legitimate reason for staying on 0.15 but we’re going forwards to 0.17.

sbueringer:
> yes, we are going to 👍 feel free to close / ignore. I supposed someone else might have a legitimate reason for staying on 0.15 but we’re going forwards to 0.17.

Sounds good. Usually we are not maintaining releases that old anymore.
It's usually quite straightforward to upgrade controller-gen and we also provide compiled binaries since a while.

If you want to continue to compile controller-gen v0.15.x yourself you can also set GOTOOLCHAIN to use an older Go version

/close

Issue #889, opened 2024-03-06
Add verbosity to `controller-gen`


grzesuav:
Recently I stumble upon a situation when running controller gen had no effect
```
controller-gen crd:headerFile="hack/boilerplate.go.txt",year=2024 paths="./pkg/api/..." output:dir="."

```

for the `crd` gen, but worked fine for `object` gen
```
controller-gen object:headerFile="hack/boilerplate.go.txt",year=2024 paths="./pkg/api/..."
```
(it produced output)

When I wanted to dig into it, I realized there is no way to enable `verbose` output in controller-gen, which would help to debug it.

Would be nice to have this feature

grzesuav:
Initial idea would be to add `logr.Logger` interface, configure some initial logger (like zap) and add a flag to enable verbose logging - of course as writing to sdtout it would conflict with `output:stdout` but it is expected

sbueringer:
/reopen 
/remove-lifecycle frozen 
/lifecycle frozen 

TBD if we should use zap or component base which uses zap for json logs

erikgb:
Could we use [slog](https://go.dev/blog/slog) instead, to avoid additional dependencies?

PR #1273, opened 2025-09-01
✨ feat: add verbose logging option to controller-gen

wazery:
Adds verbose logging to controller-gen which is disabled by default, and can be enabled. It uses `slog` as suggested in the issue #889 

Example:

`./controller-gen crd verbose=true paths=./pkg/crd/testdata output:stdout`

Fixes: #889


pkg/crd/gen.go
```diff
@@ -134,30 +134,58 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 		GenerateEmbeddedObjectMeta: g.GenerateEmbeddedObjectMeta != nil && *g.GenerateEmbeddedObjectMeta,
 	}
 
+	if ctx.Logger != nil {
```
JoelSpeed: Is there such an implementation as a null logger? I wonder if it would be better to extract the logger at the start of the function, and init it if nil to a null logger.

Then we don't have to do the nil check at every call site?

pkg/genall/genall.go
```diff
@@ -261,17 +266,31 @@ func (r *Runtime) Run() bool {
 		return true
 	}
 
+	// Set up logging based on verbose setting
+	var logger *slog.Logger
+	if r.Verbose {
```
JoelSpeed: If we aren't doing a verbose logging, should we still have a log at info level?

pkg/genall/genall.go
```diff
@@ -101,6 +102,8 @@ type Runtime struct {
 	OutputRules OutputRules
 	// ErrorWriter defines where to write error messages.
 	ErrorWriter io.Writer
+	// Verbose enables verbose logging for debugging purposes.
+	Verbose bool
```
JoelSpeed: For future expansion of options here, a bool is probably not the best choice. Typically you might see an enum here of Error, Warn, Info and Debug, what does slog expose as levels? Should we mirror their option?

wazery:
@JoelSpeed many thanks for the valuable insights, I reflected them all in the PR now.

pkg/genall/genall.go
```diff
@@ -261,17 +267,37 @@ func (r *Runtime) Run() bool {
...
 	}
 
+	// Set up logging based on log level setting
+	var logger *slog.Logger
+
+	// Use the specified log level, defaulting to Info if not set
+	logLevel := r.LogLevel
+	if logLevel == 0 {
+		logLevel = slog.LevelInfo
+	}
```
JoelSpeed, line 277: Aren't `0` and `slog.LevelInfo` equivalent?

pkg/genall/genall.go
```diff
@@ -261,17 +267,37 @@ func (r *Runtime) Run() bool {
 		return true
 	}
 
+	// Set up logging based on log level setting
+	var logger *slog.Logger
```
JoelSpeed, line 271: Why not declare this inline with `:=`?

PR #1282, opened 2025-09-13
✨ Allow title to be set on a type

cbandy:
We use types to reduce duplication and increase the consistency of our CRDs. This allows the `+kubebuilder:title` marker to be set on a type as well as a field.

```go
// +kubebuilder:title="Secret Name"
// +kubebuilder:validation:MaxLength=253
type SecretName = string
```

Fixes: #883
See: #1175


JoelSpeed (commented):
What happens if I put a title on both a type, and the field in which it is used?

cbandy:
> What happens if I put a title on both a type, and the field in which it is used?

The field is used. I've updated the tests with a type and two fields.

PR #1288, opened 2025-09-24
🌱chore: Update golangci-lint version to v2.5.0

dongjiang1989:
Sync controller-runtime .golangci.yaml

ref: https://github.com/kubernetes-sigs/controller-runtime/pull/3323

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -164,95 +164,94 @@ func init() {
 	AllDefinitions = append(AllDefinitions, ValidationIshMarkers...)
 }
 
-// +controllertools:marker:generateHelp:category="CRD validation"
 // Maximum specifies the maximum numeric value that this field can have.
+// +controllertools:marker:generateHelp:category="CRD validation"
 type Maximum float64
 
 func (m Maximum) Value() float64 {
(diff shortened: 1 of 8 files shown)


pkg/crd/markers/validation.go
```diff
@@ -299,12 +298,12 @@ type KubernetesDefault struct {
...
 // validation will be performed. Full validation of an example requires
 // submission of the containing CRD to an apiserver.
+// +controllertools:marker:generateHelp:category="CRD validation"
 type Example struct {
 	Value interface{}
 }
 
-// +controllertools:marker:generateHelp:category="CRD processing"
-// PreserveUnknownFields stops the apiserver from pruning fields which are not specified.
+// XPreserveUnknownFields stops the apiserver from pruning fields which are not specified.
```
chrischdi, line 306: Not sure if having the X prefix is wanted here / how this results later on in CRD's (if it even does)

chrischdi (commented):
/ok-to-test

pkg/crd/markers/validation.go
sbueringer: I think it has no impact. It's just the godoc of a marker

PR #1280, opened 2025-09-11
🌱Release envtest v1.34.1

dongjiang1989:

PR #1297, opened 2025-10-27
🌱 Add Windows ARM64 in Makefile

bear-redhat:
Add Windows ARM64 to release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,7 @@ release-envtest: clean-release ## Build the envtest binaries by operating system
 	OS=darwin ARCH=amd64 $(MAKE) release-envtest-docker-build
 	OS=darwin ARCH=arm64 $(MAKE) release-envtest-docker-build
 	OS=windows ARCH=amd64 $(MAKE) release-envtest-docker-build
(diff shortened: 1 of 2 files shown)


sbueringer:
Sorry, wrong button

/ok-to-test

@bear-redhat I assume you tested it and the binary generally works on Windows arm64? (because we will have no way to test it)

bear-redhat:
@sbueringer heyy thank you for looking into the PR!
I am not sure how to systematically test the binary, but it does generate the executable. I tried to use the `envtest` with `openshift-installer` and it seems working.
I missed one part yesterday about the `Dockerfile` change - `etcd` current does not release a Windows on ARM64 binary, and I am not sure how to make them generating that binary (seems they are not using GitHub workflow for releasing). Thus, I changed the `Dockerfile` to download amd64 version of the `etcd` instead, which will work on Windows on ARM as well.

sbueringer:
Alright, sounds good. Let's just give it a try.

Can you open a PR like [this](https://github.com/kubernetes-sigs/controller-tools/pull/1280) for the latest Kubernetes 1.35 alpha? So we can test the entire workflow once end to end

/lgtm
/approve

PR #1321, opened 2026-01-04
✨ deepcopy - do not generate empty import

ivankatliarchuk:
Controller-gen was generating empty `import ()` blocks in DeepCopy files when types didn't require external dependencies

Modified the writeHeader function to conditionally generate import blocks:
  - Added logic to check if importSpecs has any entries before writing the import block
  - Prevents generation of empty import () statements
  
Create unit test Unit Tests. 

 Added comprehensive E2E test that:
  - Generates actual deepcopy code from real Go source files
  - Verifies no empty import blocks exist in output
  - Confirms no import statement appears when unnecessary
  - Compares generated output with expected golden file
  
  Before (with empty imports):

```go
  package simpletypes

  import ()

  // DeepCopyInto is an autogenerated deepcopy function...
```
  After (clean output):
```go
  package simpletypes

  // DeepCopyInto is an autogenerated deepcopy function...
```


pkg/deepcopy/deepcopy_integration_test.go
```diff
@@ -104,4 +105,64 @@ var _ = Describe("CRD Generation From Parsing to CustomResourceDefinition", func
...
+
+		By("loading the expected code")
+		expectedFile, err := os.ReadFile("zz_generated.deepcopy.go")
+		Expect(err).NotTo(HaveOccurred())
+
+		By("comparing the generated output with expected")
+		Expect(outContents).To(Equal(string(expectedFile)), "generated code not as expected\n\nDiff:\n\n%s", cmp.Diff([]byte(outContents), expectedFile))
+
+		By("checking for errors")
+		Expect(hadErrs).To(BeFalse())
```
JoelSpeed: Why is this so late? Surely you want to check this straight after the Run, else the rest of the test doesn't make a lot of sense?

pkg/deepcopy/gen_test.go
```diff
@@ -0,0 +1,138 @@
...
+				}
+			}
+			(*packageHack)(unsafe.Pointer(tt.imports.pkg)).imports = pkgImportsMap
+
+			writeHeader(pkg, &buf, tt.packageName, tt.imports, tt.headerText)
+
+			output := buf.String()
+
+			// Check for empty import block "import ()"
+			if strings.Contains(output, "import (\n)") {
```
JoelSpeed: Would it normally do `import ()` or `import(` with a newline before the closing `)`?

pkg/deepcopy/deepcopy_integration_test.go
ivankatliarchuk: I copied same pattern as in other test https://github.com/kubernetes-sigs/controller-tools/blob/60c448e43ef90c770d53f90300b09f6eacc35f87/pkg/deepcopy/deepcopy_integration_test.go#L105. Moved up

pkg/deepcopy/gen_test.go
ivankatliarchuk: Resolved. As currently it is `import ()`

Image

pkg/deepcopy/deepcopy_integration_test.go
```diff
@@ -104,4 +105,63 @@ var _ = Describe("CRD Generation From Parsing to CustomResourceDefinition", func
...
+						if strings.TrimSpace(lines[j]) != "" {
+							nextNonEmptyLine = lines[j]
+							break
+						}
+					}
+					Expect(nextNonEmptyLine).NotTo(HavePrefix("import"), "should not have import block when no imports are needed")
+				}
+				break
+			}
+		}
```
sbueringer, line 158: Maybe I'm missing something, but why not just check that there is no `import` in the entire file?

Also we already check that the file is identical to `zz_generated.deepcopy.go` so this should be easily enough?

PR #1323, opened 2026-01-15
⚠️ Join multi-line godoc comments into paragraphs for CRD descriptions

ehearne-redhat:
## Problem
  Currently, multi-line godoc comments are preserved line-by-line in CRD field descriptions, causing awkward formatting when viewed with `kubectl explain` or `oc explain`.

 **Example of current behavior:**
  ```go
  // This flag tells the controller to suspend subsequent executions, it does
  // not apply to already started executions.  Defaults to false.
  type Suspend *bool
  ```

  Results in:
  ```
  $ kubectl explain cronjob.spec.suspend
  DESCRIPTION:
      This flag tells the controller to suspend subsequent executions, it does
      not apply to already started executions.  Defaults to false.
  ```
  *(Note the awkward line break mid-sentence)*
  ## Solution
  This PR aligns godoc processing with standard Go documentation behavior: consecutive non-empty lines are joined into paragraphs, with blank lines serving as paragraph breaks.

  **After this change:**
  ```
  $ kubectl explain cronjob.spec.suspend
  DESCRIPTION:
      This flag tells the controller to suspend subsequent executions, it does not apply to already started executions. Defaults to false.
  ```
  ## Implementation
  Modified `extractDoc()` in `pkg/markers/zip.go` to:
  - Join consecutive non-empty lines with spaces
  - Preserve blank lines as paragraph separators

  ## Impact
  **Breaking Change:** Generated CRD YAML files will show different formatting for multi-line descriptions.

  - **Semantic meaning unchanged:** The actual description content is the same, just formatted as continuous paragraphs
  - **Projects will need to regenerate CRDs** when upgrading controller-gen

  **Benefits:**
  - Improved readability in `kubectl explain` / `oc explain`
  ## Testing

  - Updated test fixtures to match new behavior
  - All existing tests pass

  ---

  **Note:** This is a behavior change but arguably fixes incorrect formatting. Happy to discuss if this should be behind a flag or have a migration path.

### DIFF
--- a/pkg/markers/zip_test.go
+++ b/pkg/markers/zip_test.go
@@ -1,0 +1,458 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package markers
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"strings"
+	"testing"
+)
+
+func TestExtractDoc(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name: "single line comment",
+			input: `package test
+// Single line doc
+type Foo struct{}`,
+			expected: "Single line doc",
+		},
+		{
+			name: "multi-line comment joined into paragraph",
+			input: `package test
+// This is a multi-line comment that spans
+// across multiple lines and should be joined
+// into a single paragraph.
+type Foo struct{}`,
+			expected: "This is a multi-line comment that spans across multiple lines and should be joined into a single paragraph.",
+		},
+		{
+			name: "paragraph separation with blank line",
+			input: `package test
+// First paragraph of documentation.
+//
+// Second paragraph after blank line.
+type Foo struct{}`,
+			expected: "First paragraph of documentation.\n\nSecond paragraph after blank line.",
+		},
+		{
+			name: "multiple paragraphs",
+			input: `package test
+// First paragraph line one
+// continues here.
+//
+// Second paragraph line one
+// continues here.
+//
+// Third paragraph.
+type Foo struct{}`,
+			expected: "First paragraph line one continues here.\n\nSecond paragraph line one continues here.\n\nThird paragraph.",
+		},
+		{
+			name: "code block preserved",
+			input: `package test
+// Example usage:
+// ~~~yaml
+// foo:
+//   bar: baz
+// ~~~
+type Foo struct{}`,
+			expected: "Example usage:\n~~~yaml\nfoo:\n  bar: baz\n~~~",
+		},
+		{
+			name: "code block with description before and after (no blank line)",
+			input: `package test
+// This is a description
+// with multiple lines.
+// ~~~yaml
+// foo: bar
+// ~~~
+// More description here.
+type Foo struct{}`,
+			expected: "This is a description with multiple lines.\n~~~yaml\nfoo: bar\n~~~ More description here.",
+		},
+		{
+			name: "code block with blank line after (recommended)",
+			input: `package test
+// This is a description.
+//
+// ~~~yaml
+// foo: bar
+// ~~~
+//
+// More description here.
+type Foo struct{}`,
+			expected: "This is a description.\n\n~~~yaml\nfoo: bar\n~~~\n\nMore description here.",
+		},
+		{
+			name: "triple backticks code block",
+			input: `package test
+// Example:
+// ` + "```" + `go
+// func main() {
+//   fmt.Println("hello")
+// }
+// ` + "```" + `
+type Foo struct{}`,
+			expected: "Example:\n```go\nfunc main() {\n  fmt.Println(\"hello\")\n}\n```",
+		},
+		{
+			name: "TODO comment filtered out",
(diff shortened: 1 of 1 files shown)


pkg/crd/testdata/gen/bar.example.com_foos.yaml
```diff
@@ -19,19 +19,14 @@ spec:
...
         properties:
           apiVersion:
-            description: |-
-              APIVersion defines the versioned schema of this representation of an object.
-              Servers should convert recognized schemas to the latest internal value, and
-              may reject unrecognized values.
-              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
everettraven, line 24: This change looks like it aligns better with the core API openapi descriptions as well. An example: https://github.com/kubernetes/kubernetes/blob/2c677fe0345d510f48e2fe2863deaa502f73d2a6/api/openapi-spec/v3/apis__apps__v1_openapi.json#L8

pkg/markers/zip.go
```diff
@@ -114,8 +115,19 @@ func extractDoc(node ast.Node, decl *ast.GenDecl) string {
 			}
 		}
 
-		outLines = append(outLines, line)
+		if line == "" {
+			flushParagraph()
+			outLines = append(outLines, "")
+		} else if insideCodeBlock {
+			flushParagraph()
+			outLines = append(outLines, line)
```
everettraven: Seems like this could be simplified to:
```suggestion
		if line == "" || insideCodeBlock {
			flushParagraph()
			outLines = append(outLines, line)
		}
```
?

JoelSpeed:
In my experience, there are many many APIs out there where folks use a new line instead of a `.`. For those folks, this would be an undesirable change.

The following today would be correctly rendered as three lines
```
My API does a thing
When set to foo, it does this
When set to bar, it does this
```

IIUC, with your change, it would become a single line and be odd

```
My API does a thing When set to foo, it does this When set to bar, it does this
```

Right?

everettraven:
@JoelSpeed I believe that is correct. Based on context clues within the changed method, I suspect that was the original intention as well having been modeled after https://github.com/kubernetes/apimachinery/blob/8931c298fc6d5097879d7f34f8cf8ac50a05975c/pkg/runtime/swagger_doc_generator.go#L55 . It seems as though the current implementation didn't actually work that way though.

I also see the argument that this is how it has been for many a release at this point and whether intended or not is the expected way that the documentation for fields are generated for CRDs. Fixing it now is probably a breaking change.

What is the general stance from the controller-tools maintainers on making breaking/undesirable changes if it means aligning CRDs and native K8s type generation behaviors?

ehearne-redhat:
@JoelSpeed yes, that's right. That is an important use case that I didn't think about. Would a flag be better instead, so that the current behaviour is the default one, and this could be an optional case for when it makes sense to do so?

ehearne-redhat:
@everettraven thanks for the review. I also see what you're saying too. It could be a problem for many people if they update their controller-tools and discover this odd behavior. I'm happy to implement a flag instead so it would allow for both scenarios.

everettraven:
@ehearne-redhat Another alternative could be to try and detect sentence breaks as well.

In the example that @JoelSpeed highlighted, detecting that the previous line doesn't end with `.` and the next (current?) line begins with a capital letter seems reasonably achievable, but there are probably many edge cases here that might make it more difficult to have a robust "this is a sentence break" determination.

ehearne-redhat:
@everettraven yeah, I was thinking about that when implementing this. I thought the new line approach would be best, because of scenarios where the `.` might not represent a full sentence e.g. `...` . It could be determined from a simple check but as you said there could be more scenarios.

Also, if the comments weren't in English, those languages might use different `.` e.g. `。` in Chinese/Japanese. The logic wouldn't work for these users at all. 

If the comments don't follow the convention above, this change would also be breaking for them too.

I can implement filtering in some way, but I do think this feature, because of what @JoelSpeed said, might be better being an optional flag. I'm happy to proceed in either direction if it suits. :)

ehearne-redhat:
@everettraven I have added unit tests for `ExtractDoc()` . I'm going to move this out of draft PR status for review. Happy to get further feedback on how this can be improved. I think the idea of the optional flag is probably the best, but happy to hear from others on this. :)

JoelSpeed:
I'm hesitant to say that at the moment, we probably just want to leave this as is and not even implement a flag.

The generation at the moment represents fairly well how the CRD author laid out the text, which means they have control over it. Adding clever fixes like this is going to break some people as we've discussed, and the current approach is more flexible. Those who want correct paragraphing can achieve that in their source code.

For example, I certainly wouldn't want to enable this on openshift's APIs. There are too many cases there where I worry it would break. This is a great feature for new APIs if people know about it and review the way the docs are generated, but I suspect most don't, and would expect it to just look like how they have it at source.

So for now I'm currently leaning towards we close this

CC @sbueringer @alvaroaleman in case you have an opposing opinion to mine

jenshu:
I agree that this would be a breaking change for many CRDs and that if we do want to merge it, it should be disabled by default with an opt-in flag. Echoing some of the comments above, many CRD descriptions have explicit spacing (e.g. paragraphs, bulleted lists), as well as yaml snippets that would be garbled if everything was merged into one line.

alvaroaleman:
My take is that we should not have CLI options for things that will not cause controller-tools to rc != 0. Other than the author and _maybe_ three people that read the changelog, no one will ever use it because no one is aware the option exists. And that in turn means the cost of having and thus having to keep it working is not justified.

Other than that I agree with joel, we should not do a breaking change for this. That should be reserved for cases where we know that the new behavior is clearly and obviously better in all or almost all cases, which I don't believe this is.

ehearne-redhat:
OK, fair enough. Thanks so much for taking a look at this everyone. I get your points completely - it is rather niche. Closing now.

PR #1324, opened 2026-01-15
:bug: fix(crd): sort XValidation rules for deterministic output

jaypoulz:
This address a :bug: with inconsistent validation order.

XValidation rules are now sorted by message after each append to ensure consistent CRD generation across runs. Previously, markers processed from a map produced non-deterministic iteration order, causing verify checks to fail intermittently.

Generated-by: Claude Opus 4.5 (Anthropic)

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -17,9 +17,11 @@ limitations under the License.
 package markers
 
 import (
+	"cmp"
 	"encoding/json"
 	"fmt"
 	"math"
+	"slices"
 	"strings"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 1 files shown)


JoelSpeed:
/ok-to-test

Seems reasonable to me, we don't otherwise sort markers apart from by priority which would be the same for all XValidations

/lgtm
/approve

JoelSpeed:
/hold

Thinking about this a second, should we sort on rule? Rule is required, I don't think message actually is?

jaypoulz:
Now with more linter-passing! :partying_face:

JoelSpeed:
/approve cancel
/hold

I'm just reviewing a proof PR for this and considering, is this actually what we want? Looking at the changes, it appears that we previously had all of the rules applied in the order in which they were specified on the types. With this change, this changes that.

This is important as the rules are executed in order, and a failure of an earlier rule prevents execution of a latter rule. Some folks may have written rules that rely on that behaviour

Are you sure you were seeing inconsistent ordering of rules? I haven't seen this on our repos and we have plenty of cases where we get multiple rules on a single field, all coming out consistently in order

JoelSpeed:
Going back through the code, https://github.com/kubernetes-sigs/controller-tools/blob/be3f0dad5e47458ff9007d96bb7b3a6c8b7049cf/pkg/crd/schema.go#L153-L169, it looks to me like we have a list of similar markers which are picked up from the field in the order specified, I don't see anywhere where these would have changed order or need sorting

jaypoulz:
Retiring this since it's fixed by the stable sort update for markers.

Issue #586, opened 2021-07-20
Invalid deepcopy generation with nested type-aliased slices


zx8:
Using https://github.com/kubernetes-sigs/controller-tools/commit/ffa54949a74756bbb4d5c41ae4fe8d6fea759ae7, given the following:

```go
package apis

// +kubebuilder:object:generate=true
type Foo struct {
    Bars *Bars
}

type Bars []*Bar

type Bar struct {
    Name string
}
```

After running:

```sh
$ controller-gen object
```

Then attempting to build:

```
$ go build .
./zz_generated.deepcopy.go:14:8: cannot use new([]*Bar) (type *[]*Bar) as type *Bars in assignment
```

PR #1331, opened 2026-02-05
✨feat: Support nested pointer to type-aliased slices

### LINKED ISSUE #1312: Invalid deepcopy generation for pointers to map type aliases
With types like these:
```go
type MapAlias map[string]string  // any map would work

type Foo struct {
  Bar *MapAlias
}
```

`DeepCopyInto` for `Foo` is generated with a type error:

```go
...
*out = new(map[string]string)  // incompatible types - expected MapAlias, got map[string]string
```
while instead it should generate:
```go
*out = new(MapAlias)
```

Looks like this line is the culprit: https://github.com/kubernetes-sigs/controller-tools/blob/cf302a3e550df64b909e96c8db4936353a7ead3e/pkg/deepcopy/traverse.go#L546

However, I'm not sure if the fix is as simple as replacing `underlyingElem` with `pointerType.Elem()`. I guess there was a reason why this line looks different than the previous if block

Similar to https://github.com/kubernetes-sigs/controller-tools/issues/586

dongjiang1989:
Close #1312 #586 #262


dongjiang1989:
cc @norbertcyran @sbueringer PTAL

norbertcyran:
thanks @dongjiang1989! LGTM (no lgtm rights in this repo, though)

JoelSpeed:
Could we please add a test case to cover the pointer to map case as well? #1312 doesn't look represented in the tests

dongjiang1989:
> Could we please add a test case to cover the pointer to map case as well? #1312 doesn't look represented in the tests

Thanks @JoelSpeed 
Add test case done.
Please re-check it

JoelSpeed:
/approve
/lgtm

Thanks @dongjiang1989

norbertcyran:
@dongjiang1989 Again, thanks a lot! Do you plan a release in the nearest future? If not, could we have a patch release with this fix?

dongjiang1989:
> @dongjiang1989 Again, thanks a lot! Do you plan a release in the nearest future? If not, could we have a patch release with this fix?

@JoelSpeed @sbueringer wdyt? cherry-pick to v0.20.0 or release v0.21.0?

sbueringer:
Let's do a patch release. We usually only create a new minor for new k/k minor releases

PR #1332, opened 2026-02-05
🌱chore: Update golangci-lint version to v2.8.0

dongjiang1989:

### DIFF
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -121,6 +121,10 @@ func (r *Rule) normalize() {
 // removeDupAndSort removes duplicates in strs, sorts the items, and returns a
 // new slice of strings.
 func removeDupAndSort(strs []string) []string {
+	if len(strs) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/rbac/parser.go
```diff
@@ -129,6 +129,9 @@ func removeDupAndSort(strs []string) []string {
 	}
 
 	var result []string
+	if len(set) > 0 {
```
JoelSpeed: Why do you need the if condition? Why not just have the declaration?

dongjiang1989: Thanks @JoelSpeed 
Fixed
Please re-check it

JoelSpeed:
/approve
/lgtm

@dongjiang1989 Would you be able to update controller-runtime to match please?

dongjiang1989:
> /approve
> 
> /lgtm
> 
> 
> 
> @dongjiang1989 Would you be able to update controller-runtime to match please?

Got it
I'll add PR for controller-runtime

PR #1334, opened 2026-02-11
✨ feat: Add optional roleName parameter to RBAC marker

AlirezaPourchali:
## Problem

When using namespace-scoped RBAC markers with different namespaces, controller-gen generates multiple Roles with identical names in different namespaces:

```yaml
# Generated from markers
---
kind: Role
metadata:
  name: manager-role  # ← Same name
  namespace: infrastructure
rules: [...]
---
kind: Role
metadata:
  name: manager-role  # ← Same name
  namespace: users
rules: [...]
```

While this is valid Kubernetes, it causes **kustomize to fail** when applying a global namespace transformation:

```bash
Error: namespace transformation produces ID conflict: 
[{"kind":"Role","metadata":{"name":"manager-role","namespace":"myproject-system"}...}
 {"kind":"Role","metadata":{"name":"manager-role","namespace":"myproject-system"}...}]
```

Both Roles end up in the same namespace with identical names → conflict!

Reported in: kubernetes-sigs/kubebuilder#5148
Follow up from: #1329

## Solution

This PR adds an **optional `roleName` parameter** to the RBAC marker, allowing users to specify custom role names per marker:

```go
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,roleName=infra-manager,resources=deployments,verbs=get;list
// +kubebuilder:rbac:groups="",namespace=users,roleName=user-secrets,resources=secrets,verbs=get
```

**Generated output:**

```yaml
---
kind: Role
metadata:
  name: infra-manager  # ← Unique name
  namespace: infrastructure
rules: [...]
---
kind: Role
metadata:
  name: user-secrets  # ← Unique name
  namespace: users
rules: [...]
```

Now kustomize's namespace transformation succeeds:

```yaml
# After: namespace: myproject-system transform
---
kind: Role
metadata:
  name: infra-manager  # ← Different names
  namespace: myproject-system
---
kind: Role
metadata:
  name: user-secrets  # ← Different names
  namespace: myproject-system
```

✅ No conflicts!

### 1. **Non-breaking (100% backward compatible)**

If `roleName` is not specified, behavior is **identical** to current implementation:

```go
// Existing marker (no roleName specified)
// +kubebuilder:rbac:groups=apps,namespace=prod,resources=deployments,verbs=get

// Generated Role uses default from --roleName flag (same as before)
---
kind: Role
metadata:
  name: manager-role  # ← Uses default
  namespace: prod
```

### 2. **Opt-in when needed**

Only specify `roleName` when you need unique names:

```go
// Use custom name only where you need it
// +kubebuilder:rbac:groups=apps,namespace=infra,roleName=infra-mgr,resources=deployments,verbs=get
// +kubebuilder:rbac:groups=apps,namespace=prod,resources=deployments,verbs=get  // ← Uses default
```

### 3. **Marker merging still works**

Multiple markers with the same `roleName` in the same namespace are merged (as expected):

```go
// These merge into one Role named "infra-manager"
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,roleName=infra-manager,resources=deployments,verbs=get;list
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,roleName=infra-manager,resources=statefulsets,verbs=get;list
```

This PR addresses all the concerns while still solving the original problem.

## Testing

Added comprehensive tests in `pkg/rbac/testdata/`:

1. **Custom roleName**: Roles get specified names
2. **No roleName (backward compat)**: Roles use default name
3. **Marker merging**: Multiple markers with same roleName merge correctly
4. **Multiple namespaces**: Each namespace:roleName combination creates separate Role

All existing tests pass unchanged, confirming backward compatibility.

## Implementation Details

- Modified `Rule` struct to include optional `RoleName string` field
- Updated `GenerateRoles()` to group by `(namespace, roleName)` instead of just `namespace`
- Added helper documentation explaining the use case and conflict scenario
- Zero changes to CLI or default behavior

## Migration Path for Affected Users

Users currently hitting the kustomize conflict can add `roleName` to their markers:

**Before (conflicts):**
```go
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,resources=deployments,verbs=get
// +kubebuilder:rbac:groups="",namespace=users,resources=secrets,verbs=get
```

**After (no conflicts):**
```go
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,roleName=infra-mgr,resources=deployments,verbs=get
// +kubebuilder:rbac:groups="",namespace=users,roleName=user-secrets,resources=secrets,verbs=get
```

Run `make manifests`, commit the updated RBAC YAML, and deploy. No other changes needed.

### DIFF
--- a/pkg/rbac/parser.go
+++ b/pkg/rbac/parser.go
@@ -60,6 +60,19 @@ type Rule struct {
 	// If not set, the Rule belongs to the generated ClusterRole.
 	// If set, the Rule belongs to a Role, whose namespace is specified by this field.
 	Namespace string `marker:",optional"`
+	// RoleName specifies a custom name for the Role or ClusterRole.
+	// If not set, uses the default roleName from the generator.
+	// Useful for avoiding name conflicts when the same roleName is used across multiple namespaces.
+	//
+	// Example: When using namespace-scoped RBAC markers with kustomize's global namespace transformation,
+	// multiple Roles might end up in the same namespace with identical names, causing an "ID conflict" error.
+	// Use roleName to ensure each Role has a unique name:
+	//
+	//   // +kubebuilder:rbac:groups=apps,namespace=infrastructure,roleName=infra-manager,resources=deployments,verbs=get;list
+	//   // +kubebuilder:rbac:groups="",namespace=users,roleName=user-secrets,resources=secrets,verbs=get
+	//
+	// This generates Roles named "infra-manager" and "user-secrets" instead of both being "manager-role".
+	RoleName string `marker:"roleName,optional"`
 }
 
 // ruleKey represents the resources and non-resources a Rule applies.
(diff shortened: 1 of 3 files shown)


camilamacedo86:
@JoelSpeed @sbueringer WDYT?
If we add roleName could we move forward?
have you better name for the marker?
Maybe only name since it is for rbac one?

sbueringer:
General idea makes sense to me. I think roleName is fine. Just name doesn't seem clear enough to me. Let's see what @JoelSpeed thinks.

With this it's clear that this rule belongs to the infra-deployment-manager role:
```
// +kubebuilder:rbac:roleName=infra-deployment-manager,groups=apps,namespace=infrastructure,resources=statefulsets,verbs=get;list
```

With this it's not really clear what the name is referring to
```
// +kubebuilder:rbac:name=infra-deployment-manager,groups=apps,namespace=infrastructure,resources=statefulsets,verbs=get;list
```
(this marker is for a rule, but a rule itself does not have a name)

I realize that namespace is also referring to the Role namespace, but not sure if that's good and we should continue that way


/ok-to-test

sbueringer:
Re: CI failure (https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-tools/1334/pull-controller-tools-test-main/2021884835185823744)

The failure is orthogonal to this PR, looks like our tests are directly downloading envtest/kubebuilder binaries from the old location that is not available anymore. We'll need a PR to migrate this to download from the new location (controller-tools release attachments). I would probably continue to just use curl and not introduce setup-envtest. As part of that we'll also have to download newer binaries (looks like the job is using Kubernetes v1.11, which is not available on the new location)

JoelSpeed:
I agree, roleName works well here, apart from needing to fix CI separately, I think this LGTM

sbueringer:
Thx again!

/lgtm
/approve

(We can run retest once CI is fixed on main)

camilamacedo86:
Thank you a lot @sbueringer, @JoelSpeed  and @AlirezaPourchali 🙏

PR #1335, opened 2026-02-12
:sparkles: feat: add `kubebuilder:externalDoc` marker

pedjak:
Implement the `kubebuilder:externalDocs` marker to allow specifying external documentation (url and description) on fields and types, populating the externalDocs field in the generated OpenAPI schema.

The url field is required and validated to be a well-formed URL.
The description field is optional.

Ref: https://spec.openapis.org/oas/v3.0.0.html#external-documentation-object

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/pkg/crd/testdata/external_docs/types.go
+++ b/pkg/crd/testdata/external_docs/types.go
@@ -1,0 +1,59 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// +groupName=testdata.kubebuilder.io
+// +versionName=v1
+package external_docs
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+
+// ExternalDocSpec defines the desired state of ExternalDoc
+type ExternalDocSpec struct {
+	// This tests that external documentation can be attached to a field with url and description.
+	// +kubebuilder:externalDocs:url="https://example.com/docs",description="external docs description"
+	FieldWithExternalDoc string `json:"fieldWithExternalDoc,omitempty"`
+
+	// This tests that external documentation can be attached with only url.
+	// +kubebuilder:externalDocs:url="https://example.com/docs"
+	FieldWithExternalDocURLOnly string `json:"fieldWithExternalDocURLOnly,omitempty"`
+
+	// This tests that external documentation from a type is propagated.
+	TypeWithExternalDoc TypeWithExternalDoc `json:"typeWithExternalDoc,omitempty"`
+}
+
+// TypeWithExternalDoc is a type with external documentation.
+// +kubebuilder:externalDocs:url="https://example.com/type-docs",description="type-level external docs"
+type TypeWithExternalDoc string
+
+// ExternalDoc is the Schema for the external docs API
+type ExternalDoc struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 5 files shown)


pedjak:
Related to https://github.com/kubernetes/kubernetes/pull/136988 and https://github.com/kubernetes/kubernetes/pull/136977

sttts:
Lgtm.

pedjak:
> The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=controller-tools&pr=1335). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Apedjak). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/controller-tools/issues?q=is:issue+is:open) when you hit one in your PR.

It looks like there is an infra issue with [the failed job](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-tools/1335/pull-controller-tools-test-main/2022025746247061504#1:build-log.txt%3A166-171):

```
 + curl -sL https://storage.googleapis.com/kubebuilder-tools/kubebuilder-tools-1.11.0-linux-amd64.tar.gz -o /tmp/kubebuilder-tools-1.11.0-linux-amd64.tar.gz
+ tar -zvxf /tmp/kubebuilder-tools-1.11.0-linux-amd64.tar.gz -C /tmp/
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
make[1]: *** [Makefile:90: test] Error 2
```

I can even reproduce it locally.

pkg/crd/testdata/external_docs/types.go
```diff
@@ -0,0 +1,87 @@
...
+	// This tests that external documentation can be attached with only url.
+	// +kubebuilder:externalDoc:url="https://example.com/docs"
+	FieldWithExternalDocURLOnly string `json:"fieldWithExternalDocURLOnly,omitempty"`
+
+	// This tests that external documentation from a type is propagated.
+	TypeWithExternalDoc TypeWithExternalDoc `json:"typeWithExternalDoc,omitempty"`
+
+	// This field has a doc comment link definition.
+	//
+	// [Link Def Docs]: https://example.com/link-def
```
JoelSpeed: Why do we need to support this syntax instead of just supporting the explicit markers?

pedjak: I thought it is a nice add-on, given that many projects today put already links in godoc. This would allows us to find them and put them as `externalDocs` structure. If you think it could be surprising to users, we could enable this behavior with a command line flag?

JoelSpeed: I'd prefer we didn't carry this complexity at all if I'm honest. We don't have existing behaviour like this, everything is explicitly part of the markers and I think we should probably stick to that pattern

CC @sbueringer @alvaroaleman keep me honest here

sbueringer:
@pedjak Yes, please see https://github.com/kubernetes-sigs/controller-tools/pull/1334#issuecomment-3889872651 for more details

pkg/crd/testdata/external_docs/types.go
sbueringer: Agree. I also would prefer to only use markers for this

pedjak: > Agree. I also would prefer to only use markers for this

done, ptal.

pedjak: > I'd prefer we didn't carry this complexity at all if I'm honest. We don't have existing behaviour like this, everything is explicitly part of the markers and I think we should probably stick to that pattern

I have removed godoc link parsing, but just got an idea: how about that we control the link source via an additional parameter, e.g.:

```
// +kubebuilder:externalDoc:fromComment
```

or simply just

```
// +kubebuilder:externalDoc
```

and then we look into godoc comment for links? wdyt @JoelSpeed ? This enables users to keep links in the single place, without duplication.

sbueringer:
Thx!

/lgtm
/assign @JoelSpeed

JoelSpeed:
/approve

Thanks @pedjak

Issue #1063, opened 2024-09-20
Panic parsing resources not directly imported


mnencia:
Minimal repro:

Create the needed directories `mkdir repro1 repro2 repro3` and put the files inside

``` go
// +groupName=repro.io
package repro1

import (
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

	"sigs.k8s.io/controller-tools/repro2"

        // Uncomment the following line to fix the panic
	// _ "sigs.k8s.io/controller-tools/repro3"
)

type Repro struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata"`

	Reproducer repro2.MyReproAlias `json:"reproducer"`
}
```

``` go
package repro2

import "sigs.k8s.io/controller-tools/repro3"

type MyReproAlias = repro3.MyReproStruct
```

``` go
package repro3

type MyReproStruct struct {
	ReproString string `json:"repro"`
}
```

Then run controller-gen with:

``` sh
go run ./cmd/controller-gen/ crd paths=./repro
```

The result should be the following:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4d4abb2]

goroutine 1 [running]:
sigs.k8s.io/controller-tools/pkg/crd.(*Parser).NeedPackage(0xc0004040c0, 0x0)
	/Users/mnencia/prj/tmp/controller-tools/pkg/crd/parser.go:237 +0x52
sigs.k8s.io/controller-tools/pkg/crd.(*Parser).NeedSchemaFor(0xc0004040c0, {0x0, {0xc000191710, 0xd}})
	/Users/mnencia/prj/tmp/controller-tools/pkg/crd/parser.go:170 +0x5a
sigs.k8s.io/controller-tools/pkg/crd.(*schemaContext).requestSchema(0xc0014c0540?, {0xc000195d10?, 0x4e31d15?}, {0xc000191710?, 0xd?})
	/Users/mnencia/prj/tmp/controller-tools/pkg/crd/schema.go:108 +0xdb
sigs.k8s.io/controller-tools/pkg/crd.namedToSchema(0xc0014c0540, 0xc0004be1b0)
	/Users/mnencia/prj/tmp/controller-tools/pkg/crd/schema.go:280 +0x206
[...]

PR #1344, opened 2026-02-16
🌱 crd: add integration test for indirect type alias resolution

meliezer:
Adds regression coverage for the scenario reported in #1063.

In controller-tools v0.16.x, CRD generation could panic when a root API type
contained a field defined via a type alias whose target type lived in a package
not directly imported by the root package.

This PR adds integration-level coverage using the existing Ginkgo/Gomega test
pattern in parser_integration_test.go:
- runs from the pkg/crd/testdata module
- uses loader.LoadRoots and parser APIs
- avoids subprocess execution

The test uses a descriptive fixture layout to model the alias chain:
rootpkg → aliaspkg → targetpkg

Also removes the previous subprocess-based test approach and replaces it with
the standard integration test structure used in this package.

The issue no longer reproduces on main (and v0.20.1), likely fixed as a side
effect of #1122. This test ensures the behavior remains covered and prevents
future regressions.

### DIFF
--- a/pkg/crd/parser_integration_test.go
+++ b/pkg/crd/parser_integration_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -233,6 +233,58 @@ var _ = Describe("CRD Generation From Parsing to CustomResourceDefinition", func
 				assertCRDForGroupKind(pkgs[0], schema.GroupKind{Kind: "CronJob"}, "testdata._cronjobs.yaml")
 			})
 		})
+	})
+
+	It("should resolve type aliases to packages not directly imported", func() {
+		By("switching into testdata to appease go modules")
+		cwd, err := os.Getwd()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(os.Chdir("./testdata")).To(Succeed())
+		defer func() { Expect(os.Chdir(cwd)).To(Succeed()) }()
+
+		By("loading the roots")
+		pkgs, err := loader.LoadRoots("./typealiasindirect/rootpkg")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(pkgs).To(HaveLen(1))
+		pkg := pkgs[0]
+
+		By("setting up the parser")
+		reg := &markers.Registry{}
+		Expect(crdmarkers.Register(reg)).To(Succeed())
+		parser := &crd.Parser{
+			Collector: &markers.Collector{Registry: reg},
+			Checker:   &loader.TypeChecker{},
+		}
+		crd.AddKnownTypes(parser)
+
+		By("requesting that the package be parsed")
+		parser.NeedPackage(pkg)
+
+		By("checking that no non-type errors occurred along the way")
+		Expect(packageErrors(pkg, packages.TypeError)).NotTo(HaveOccurred())
+
+		By("requesting that the CRD be generated")
+		groupKind := schema.GroupKind{Kind: "Repro", Group: "repro.io"}
+		parser.NeedCRDFor(groupKind, nil)
+
+		By("fixing top level ObjectMeta on the CRD")
+		crd.FixTopLevelMetadata(parser.CustomResourceDefinitions[groupKind])
+
+		By("checking that the CRD is present")
+		generated, found := parser.CustomResourceDefinitions[groupKind]
+		Expect(found).To(BeTrue())
+
+		By("verifying schema includes the field from the aliased package")
+		versions := generated.Spec.Versions
+		Expect(versions).NotTo(BeEmpty())
+		schemaProps := versions[0].Schema.OpenAPIV3Schema.Properties
+		specSchema, ok := schemaProps["spec"]
+		Expect(ok).To(BeTrue())
+		reproducerSchema, ok := specSchema.Properties["reproducer"]
+		Expect(ok).To(BeTrue())
+		Expect(reproducerSchema.Type).To(Equal("object"))
+		_, ok = reproducerSchema.Properties["repro"]
+		Expect(ok).To(BeTrue())
 	})
 
 	It("should generate plural words for Kind correctly", func() {
--- a/pkg/crd/testdata/typealiasindirect/aliaspkg/alias.go
+++ b/pkg/crd/testdata/typealiasindirect/aliaspkg/alias.go
@@ -1,0 +1,21 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aliaspkg
+
+import "testdata.kubebuilder.io/cronjob/typealiasindirect/targetpkg"
+
+type MyAlias = targetpkg.TargetStruct
--- a/pkg/crd/testdata/typealiasindirect/rootpkg/doc.go
+++ b/pkg/crd/testdata/typealiasindirect/rootpkg/doc.go
@@ -1,0 +1,20 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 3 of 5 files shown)


pkg/crd/issue1063_alias_not_directly_imported_test.go
```diff
@@ -0,0 +1,91 @@
+package crd_test
```
sbueringer: Sorry not sure I follow why this test cannot be added like all our other tests

meliezer: Thanks for taking a look! The only reason this test runs controller-gen from pkg/crd/testdata is that pkg/crd/testdata is its own Go module (module testdata.kubebuilder.io/cronjob in pkg/crd/testdata/go.mod).

The repro packages live under that module (testdata.kubebuilder.io/cronjob/issue1063/...). If the test runs controller-gen with cmd.Dir set to the repo root (main module), go/packages ends up trying to resolve those imports from the main module and fails with “no required module provides package …/issue1063/repro2”. Running from the testdata module root makes the imports resolve exactly as intended and matches how this repository already structures other fixtures under pkg/crd/testdata.

I can add a short comment in the test explaining this (pkg/crd/testdata is a separate module; run controller-gen from there so package resolution uses that go.mod) if that helps readability.

If you prefer a different pattern that keeps cmd.Dir at repo root (e.g. setting GOMOD/-C), I’m happy to adjust, but I wasn’t able to make that work reliably with go/packages without executing inside the testdata module.

sbueringer: I was just naively wondering what makes this test different from the ones in parser_integration_test.go

meliezer: Thanks for the question, and for your patience. 
The existing tests in parser_integration_test.go are close, but they exercise the parser through the internal API (loader.LoadRoots + parser.NeedPackage/NeedCRDFor) after chdir-ing into pkg/crd/testdata.

This test is intentionally a bit different: it runs controller-gen end-to-end as a subprocess, from the existing pkg/crd/testdata module (testdata.kubebuilder.io/cronjob). That matters for the #1063 scenario because the historical failure was tied to package/module resolution for the alias chain repro1 -> repro2 -> repro3, where the target type lived in a package not directly imported by the root package.

So the overlap is real, but this adds regression coverage for the CLI/module-resolution path rather than only the parser API path.

If you’d prefer, I can also rework it into parser_integration_test.go, but I wanted to keep the test exercising the exact module-resolution setup that originally made this scenario fail.

pkg/crd/testdata/issue1063/repro1/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package repro1 contains API Schema definitions.
```
camilamacedo86: Instead of issue number and repro1 and etc 
Could we add meaningful names?

pkg/crd/issue1063_alias_not_directly_imported_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+// repoRoot walks up from this test file until it finds the repository go.mod.
+func repoRoot(t *testing.T) string {
```
camilamacedo86: The tests should follow the project standards that means BBD style and use;

```
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
```

Could you please check the other integration and tests in the project?

pkg/crd/issue1063_alias_not_directly_imported_test.go
camilamacedo86: I think,   I understand the concern, but this is already solved by the existing test pattern. Look at parser_integration_test.go lines:
                                                                                
```
  By("switching into testdata to appease go modules")       
  prevCwd, err = os.Getwd()                                                     
  Expect(prevCwd).To(BeADirectory())                                            
  Expect(os.Chdir("./testdata")).To(Succeed())  // ← Here we have the magic         
``` 
                                                                                
All existing tests:                                                           
  1. Change directory to testdata (separate module)                             
  2. Call loader.LoadRoots()                                                    
  3. The loader uses the testdata/go.mod for resolution
                                                                                
This is exactly the same as running controller-gen from the testdata directory, but using internal APIs instead of subprocess.

pkg/crd/testdata/issue1063/repro1/doc.go
camilamacedo86: Something like

```
  pkg/crd/testdata/                                         
  └── typealiasindirect/
      ├── rootpkg/      
      │   ├── doc.go      // Package doc with kubebuilder markers               
      │   └── types.go    // CRD with field using alias from aliaspkg
      ├── aliaspkg/                                                             
      │   └── alias.go    // type MyAlias = targetpkg.TargetStruct
      └── targetpkg/                                                            
          └── types.go    // The actual struct definition   
```

Also, it seems that would fit better to pkg/crd/parser_integration_test.go or create pkg/crd/type_alias_integration_test.go

camilamacedo86 (changes requested):
Thanks for adding regression coverage for this. The scenario is valid and worth testing. 🎉 

I do think the test should be refactored before merge so it matches the existing patterns in this package.

The main changes I’d ask for are:

- Please rewrite it in the existing Ginkgo/Gomega style instead of `testing.T`
- Please avoid the subprocess approach here and use the internal parser APIs instead
- Please rename the testdata packages to something descriptive, so the scenario is easier to understand when reading the fixture layout

I think the internal API path should work for this case too. The existing integration tests already change into `testdata` to use that module’s `go.mod`, then call `loader.LoadRoots(...)`. That seems like the right fit here as well, and it keeps this test consistent with the rest of the suite.

I’d also suggest renaming the fixture packages from `repro1/repro2/repro3` to names that describe their role in the alias chain. That will make the test much easier to follow later.

**So overall: good regression test, but It think it need to be rewritten to follow the established integration-test structure before merging.**

meliezer:
Thanks for the detailed feedback — that makes sense.

You’re right that the existing integration-test pattern already handles the separate testdata module correctly by switching into pkg/crd/testdata before calling loader.LoadRoots(...), so there’s no real need for the subprocess approach here.

I’ll rework this to follow the existing Ginkgo/Gomega integration style, use the internal parser APIs, and rename the fixture packages to something more descriptive for the alias chain scenario. I also agree it fits better alongside the existing parser integration coverage.

Thanks again — I’ll update the PR accordingly.

meliezer:
Thanks for the guidance — I’ve updated the PR to follow the existing integration-test pattern.

The test now:
- uses the internal parser APIs (loader.LoadRoots + parser)
- runs from the existing pkg/crd/testdata module
- avoids subprocess execution
- uses more descriptive fixture names for the alias chain

I’ve also updated the PR description to reflect the current approach.

camilamacedo86:
Hi @meliezer 

Thank you for the contribution 🎉 
Could you please fix the PR title?

camilamacedo86:
Hi @meliezer 

Could you please check it is not passing in the tests: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-tools/1344/pull-controller-tools-test-main/2049894933682196480/build-log.txt

Also, because this one is to add a test only we should not use the emoji 🐛 
It does not change the final result for end users, in this case the right emoji is 🌱

camilamacedo86:
If you can squash all commits will be great to ensure a nice git history.
Thank you

meliezer:
`+ go generate ./...
/home/prow/go/src/sigs.k8s.io/controller-tools/pkg/crd/testdata/wrong_annotation_format/cronjob_types.go:505:6: annotation api-approved.kubernetes.io:https://github.com/kubernetes-sigs/controller-tools is not in 'xxx=xxx' format
Error: not all generators ran successfully`
@camilamacedo86: 

I re-ran `go generate ./...` locally from `pkg/crd/testdata` and it fails on existing testdata packages before reaching anything related to this PR.

In particular, existing `gen/*` fixtures fail with `apiVersion "apiextensions.k8s.io/v1beta1" is not supported`, and `wrong_annotation_format` is also part of the existing negative marker testdata. I do not see an interaction with the new alias-resolution fixture.

The PR-specific parser integration test failure was addressed by aligning the error handling with the existing tests.

sbueringer:
@meliezer You'll have to fix the copyright header in the files listed in the failing job: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-tools/1344/pull-controller-tools-test-main/2084381664741429248

(this is the correct one: https://github.com/kubernetes-sigs/controller-tools/blob/863b80b2745ffe99beaf10c4c5c5ce6fda4fa348/hack/boilerplate/boilerplate.go.txt)

meliezer:
> @meliezer You'll have to fix the copyright header in the files listed in the failing job: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_controller-tools/1344/pull-controller-tools-test-main/2084381664741429248
> 
> (this is the correct one: https://github.com/kubernetes-sigs/controller-tools/blob/863b80b2745ffe99beaf10c4c5c5ce6fda4fa348/hack/boilerplate/boilerplate.go.txt)

Thank you, yes, I know, but I had to log off yesterday. Working on it.

PR #1346, opened 2026-02-17
:sparkles: Applyconfigs: Generate extract functions

alvaroaleman:
This change makes us generate the extract functions for applyconfigurations. It re-used the existing openapi generation with some adjustments and then feeds that into the applyconfig generator.

Contrary to CRD generation we do not flatten the schema because we can and because it will likely often reduce the overall size.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -119,12 +119,6 @@ if ! git diff --quiet --exit-code; then
   exit 1
 fi
 
-if ! git diff --quiet --exit-code; then
-  header_text "go generate produced changes"
-  git diff
-  exit 1
-fi
-
 header_text "running golangci-lint"
 make lint
 


pkg/applyconfiguration/openapi.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+	}
+
+	p.NeedPackage(root)
+
+	// Collect root CRD type names and trigger schema generation for all
+	// transitive types. NeedSchemaFor (not NeedFlattenedSchemaFor) preserves
+	// $ref references in the schemas.
+	crdTypeSet := make(map[string]bool)
+	var genErr error
```
sbueringer: genErr never seems to be written?

pkg/applyconfiguration/testdata/cronjob/api/v1/applyconfiguration/api/v1/cronjob.go
```diff
@@ -30,6 +33,47 @@ func CronJob(name, namespace string) *CronJobApplyConfiguration {
 	return b
 }
```
sbueringer, line 35: As we now have everything in place. Do you think it would make sense to have a small unit test that actually uses this generated code to go through an extract/modify-in-place/apply workflow?

pkg/applyconfiguration/openapi.go
```diff
@@ -0,0 +1,301 @@
...
+			if !found {
+				continue
+			}
+			resolved := refSchema.DeepCopy()
+			// Recurse into the resolved schema to handle nested embeddings.
+			resolveAllOfRefs(resolved, pkg, p, pkgByPath)
+			schema.AllOf[i] = *resolved
+		} else {
+			// Recurse into non-ref AllOf entries.
+			resolveAllOfRefs(entry, contextPkg, p, pkgByPath)
```
sbueringer: If I see correctly this code path is not executed by the integration test, let's extend the test type accordingly?

pkg/applyconfiguration/openapi.go
```diff
@@ -0,0 +1,301 @@
...
+			schema.Ref = &newRef
+		}
+	}
+
+	for k, v := range schema.Properties {
+		convertRefs(&v, contextPkg)
+		schema.Properties[k] = v
+	}
+	for i := range schema.AllOf {
+		convertRefs(&schema.AllOf[i], contextPkg)
```
sbueringer, line 260: This one is also not executed, let's extend the test type

(same for l.295-298)

pkg/applyconfiguration/openapi.go
```diff
@@ -0,0 +1,301 @@
...
+// resolveAllOfRefs walks the schema and resolves $ref entries inside AllOf slices
+// by replacing them with the referenced type's schema (deep-copied). This preserves
+// $ref in other locations (Properties, Items, etc.) while making AllOf entries ready
+// for flattening by FlattenEmbedded.
+func resolveAllOfRefs(schema *apiextensionsv1.JSONSchemaProps, contextPkg *loader.Package, p *crd.Parser, pkgByPath map[string]*loader.Package) {
+	if schema == nil {
+		return
+	}
+
+	for i := range schema.AllOf {
```
sbueringer, line 183: Is this only relevant for AllOf? What about similar fields like AnyOf, OneOf, ..

EDIT: Just noticed that we are dropping nullable, anyOf, oneOf, not below. So I assume this is the reason why we only handle AllOf?

Let's maybe add some more godocs around here (same for a few of my other comments below)

Q: What does this mean for Extract funcs if anyOf / oneOf etc are not in this schema? Are they negatively affected somehow?

pkg/applyconfiguration/openapi.go
```diff
@@ -0,0 +1,301 @@
...
+		entry := &schema.AllOf[i]
+		if entry.Ref != nil && len(*entry.Ref) > 0 {
+			typeName, pkgPath, err := crd.RefParts(*entry.Ref)
+			if err != nil {
+				continue
```
sbueringer: Is it fine to just continue here (and in similar cases below) or should these cases be at least logged?

pkg/applyconfiguration/openapi.go
```diff
@@ -0,0 +1,301 @@
...
+			resolveAllOfRefs(resolved, pkg, p, pkgByPath)
+			schema.AllOf[i] = *resolved
+		} else {
+			// Recurse into non-ref AllOf entries.
+			resolveAllOfRefs(entry, contextPkg, p, pkgByPath)
+		}
+	}
+
+	// Recurse into other schema locations that may contain nested AllOf refs.
+	for k, v := range schema.Properties {
```
sbueringer, line 217: I would assume at least here we also have to recurse for other cases like AnyOf, OneOf, Not, ...?

Same for l.245++

pkg/applyconfiguration/openapi.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	if schema.AdditionalProperties != nil && schema.AdditionalProperties.Schema != nil {
+		resolveAllOfRefs(schema.AdditionalProperties.Schema, contextPkg, p, pkgByPath)
+	}
+}
+
+// convertRefs walks the schema and converts internal $ref links from the
+// controller-tools format (#/definitions/pkg~1path~0TypeName) to swagger
+// definition keys (#/definitions/io.k8s.pkg.path.TypeName).
+func convertRefs(schema *apiextensionsv1.JSONSchemaProps, contextPkg *loader.Package) {
```
sbueringer, line 239: Would it make sense to have unit tests for this func? (also fine if we transitively test this in other ways, e.g. by having a unit test that uses the generated types that relies on all of this working correctly)

sbueringer:
/hold
Just to ensure PR doesn't get accidentally merged too early :)

pkg/applyconfiguration/testdata/cronjob/api/v1/applyconfiguration/api/v1/cronjob.go
alvaroaleman: I am guessing by unit test you mean envtest?

sbueringer: Didn't think about it, but yes

pkg/applyconfiguration/openapi.go
alvaroaleman: > EDIT: Just noticed that we are dropping nullable, anyOf, oneOf, not below. So I assume this is the reason why we only handle AllOf?

Yes

> Q: What does this mean for Extract funcs if anyOf / oneOf etc are not in this schema? Are they negatively affected somehow?

I can only speculate but I'd assume that the assumption is that the server does any validation and we don't need to do it here anymore. The important part I believe it the `elementRelationship` as that controls for example if you will take ownership of the entire list when changing an element or just the element

pkg/applyconfiguration/openapi.go
alvaroaleman: I don't follow, why would we have to recurse in anyOf, oneOf etc, we just drop them later?

alvaroaleman:
Temporarily based this on top of https://github.com/kubernetes-sigs/controller-tools/pull/1348 so I can see if it passes CI

pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go
```diff
@@ -293,8 +294,12 @@ type CronJobSpec struct {
 	// +kubebuilder:validation:MultipleOf=2
 	Int32WithValidations int32 `json:"int32WithValidations"`
 
-	// This tests that unexported fields are skipped in the schema generation
-	unexportedField string
+	// This tests that unexported fields are skipped in the schema generation.
+	// Commented out because structured-merge-diff's reflectcache includes
+	// unexported fields without JSON tags (a bug in lookupJsonTags which
+	// doesn't check reflect.StructField.PkgPath), causing ExtractInto to
+	// fail with "field not declared in schema".
```
alvaroaleman, line 301: https://github.com/kubernetes-sigs/structured-merge-diff/pull/312

pkg/applyconfiguration/openapi.go
alvaroaleman: I ended up writing and envtest test which tests the correctness of this

pkg/applyconfiguration/openapi.go
alvaroaleman: fixed (why did the linter not catch this?)

sbueringer: Seems claude is smarter hiding something like this than our linters are finding it 😃

pkg/applyconfiguration/openapi.go
sbueringer: I realized that only when reading further down and forget to add this here as well 

> EDIT: Just noticed that we are dropping nullable, anyOf, oneOf, not below. So I assume this is the reason why we only handle AllOf?

pkg/applyconfiguration/openapi.go
alvaroaleman: They are all "should never happen" cases but you are right, I have added an error return in case they do happen, silently doing the wrong thing should never happen

sbueringer:
/lgtm

Keeping hold for now in case Joel also wants to take a look

/assign @JoelSpeed

PR #1347, opened 2026-02-18
🐛 Fix applyconfiguration generator for cluster-scoped resources

joelanford:
Detect `+kubebuilder:resource:scope=Cluster` and inject `+genclient:nonNamespaced` so the generated ApplyConfiguration constructor omits the namespace parameter.


pkg/applyconfiguration/testdata/cronjob/api/v1/applyconfiguration/api/v1/clusterscopedresource.go
```diff
@@ -0,0 +1,220 @@
...
+// ClusterScopedResource is a cluster-scoped resource for testing applyconfiguration generation
+type ClusterScopedResourceApplyConfiguration struct {
+	metav1.TypeMetaApplyConfiguration    `json:",inline"`
+	*metav1.ObjectMetaApplyConfiguration `json:"metadata,omitempty"`
+	Spec                                 *apiv1.ClusterScopedResourceSpec `json:"spec,omitempty"`
+}
+
+// ClusterScopedResource constructs a declarative configuration of the ClusterScopedResource type for use with
+// apply.
+func ClusterScopedResource(name string) *ClusterScopedResourceApplyConfiguration {
```
joelanford, line 24: This is the important bit. The constructor now correctly accepts just the name. On `main`, the constructor accepts both name and namespace as a string, which makes the field owner own the `metadata.namespace` field. But of course the apiserver throws that away and reconcilers continuously notice this diff and keep trying to set `metadata.namespace` to `""`

alvaroaleman (approved):

PR #1348, opened 2026-02-19
:seedling: Fix go generate validation

alvaroaleman:
* We only ever run it for the root go module, so invalid or otherwise broken go genreate directives in other go modules went unnoticed
* We didn't validate that it doesn't produce changes

### DIFF
--- a/pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go
+++ b/pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go
@@ -16,7 +16,7 @@ limitations under the License.
 // TODO(directxman12): test this across both versions (right now we're just
 // trusting k/k conversion, which is probably fine though)
 
-//go:generate ../../../../../../.run-controller-gen.sh crd:ignoreUnexportedFields=true,allowDangerousTypes=true paths=./;./deprecated;./unserved;./job/... output:dir=.
+//go:generate ../../../../../../.run-controller-gen.sh crd:ignoreUnexportedFields=true,allowDangerousTypes=true paths=. output:dir=.
 
 // +groupName=testdata.kubebuilder.io
 // +versionName=v1
--- a/pkg/crd/testdata/gen/zoo/zoo_types.go
+++ b/pkg/crd/testdata/gen/zoo/zoo_types.go
@@ -13,8 +13,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-//go:generate ../../../../../.run-controller-gen.sh crd:crdVersions=v1beta1 paths=. output:dir=.
-//go:generate mv bar.example.com_zoos.yaml bar.example.com_zoos.v1beta1.yaml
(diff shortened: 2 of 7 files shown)


pkg/crd/testdata/gen/zoo/zoo_types.go
```diff
@@ -13,8 +13,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-//go:generate ../../../../../.run-controller-gen.sh crd:crdVersions=v1beta1 paths=. output:dir=.
```
alvaroaleman, line 16: We don't support v1beta1 CRDs

pkg/crd/testdata/wrong_annotation_format/cronjob_types.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 // TODO(directxman12): test this across both versions (right now we're just
 // trusting k/k conversion, which is probably fine though)
 
-//go:generate ../../../.run-controller-gen.sh crd:ignoreUnexportedFields=true,allowDangerousTypes=true paths=./;./deprecated;./unserved;./job/... output:dir=.
+//go:generate bash -c "! ../../../../.run-controller-gen.sh crd:ignoreUnexportedFields=true,allowDangerousTypes=true paths=. output:none"
```
alvaroaleman, line 19: This is expected to fail, so assert that

pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 // TODO(directxman12): test this across both versions (right now we're just
 // trusting k/k conversion, which is probably fine though)
 
-//go:generate ../../../../../../.run-controller-gen.sh crd:ignoreUnexportedFields=true,allowDangerousTypes=true paths=./;./deprecated;./unserved;./job/... output:dir=.
+//go:generate ../../../../../../.run-controller-gen.sh crd:ignoreUnexportedFields=true,allowDangerousTypes=true paths=. output:dir=.
 
 // +groupName=testdata.kubebuilder.io
```
alvaroaleman, line 21: The generated CRD is almost identical with the generated CRD under pkg/crd/testdata/ but the issue is in the almost - We removed some fields because the AC generator doesn't support them and in order to write an envtest test in https://github.com/kubernetes-sigs/controller-tools/pull/1346 I need a CRD with a schema

JoelSpeed: Do you know which fields were removed, can we comment with that list? Do we want to keep this in ~sync in the future?

alvaroaleman: We generally just comment them out, e.G. here: https://github.com/kubernetes-sigs/controller-tools/pull/1346/changes#diff-d8169aa02594b49946ef0329f828f88afae342f4f508f7b8bad1fab9d66ffd0aR302

sbueringer:
Thank you!

/lgtm

/hold
for @JoelSpeed's review

PR #1349, opened 2026-02-19
:seedling: Validate tidyness of all go modules

alvaroaleman:
We only checked the root module


sbueringer:
/lgtm
/approve 

Thx!

PR #1352, opened 2026-02-21
:sparkles: Add support for k8s:enum markers

alvaroaleman:
This change adds support for the k8s:enum marker as described [in the kep](https://github.com/kubernetes/enhancements/tree/c68dfb941894fc8859a951fe47a60b2161300b88/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen).

It is a rebase of https://github.com/kubernetes-sigs/controller-tools/pull/1250 plus a test

Closes https://github.com/kubernetes-sigs/controller-tools/pull/1250

/assign @JoelSpeed @lalitc375

### DIFF
--- a/pkg/crd/markers/schema_context.go
+++ b/pkg/crd/markers/schema_context.go
@@ -1,0 +1,30 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package markers
+
+import (
+	"sigs.k8s.io/controller-tools/pkg/loader"
+	"sigs.k8s.io/controller-tools/pkg/markers"
+)
+
+// SchemaContext carries contextual information passed to
+// SchemaMarker.ApplyToSchema implementations that need more than the
+// schema itself (for example, to inspect package AST or type info).
+type SchemaContext struct {
+	Package  *loader.Package
+	TypeInfo *markers.TypeInfo
+}
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -19,7 +19,11 @@ package markers
 import (
 	"encoding/json"
 	"fmt"
+	"go/ast"
+	"go/token"
+	"go/types"
 	"math"
+	"slices"
 	"strings"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -93,7 +97,7 @@ var TypeOnlyMarkers = []*definitionWithHelp{
 		WithHelp(markers.SimpleHelp("CRD validation", "specifies a list of field names that must conform to the ExactlyOneOf constraint.")),
 	must(markers.MakeDefinition(ValidationAtLeastOneOfPrefix, markers.DescribesType, AtLeastOneOf(nil))).
 		WithHelp(markers.SimpleHelp("CRD validation", "specifies a list of field names that must conform to the AtLeastOneOf constraint.")),
-	must(markers.MakeDefinition(K8sEnumTag, markers.DescribesType, struct{}{})).
+	must(markers.MakeDefinition(K8sEnumTag, markers.DescribesType, K8sEnum{})).
 		WithHelp(markers.SimpleHelp("CRD", "indicates that the given type is an enum; all const values of this type are considered values in the enum")),
 }
 
@@ -659,7 +663,7 @@ type ExactlyOneOf []string
 // +controllertools:marker:generateHelp:category="CRD validation"
 type AtLeastOneOf []string
 
-func (m Maximum) ApplyToSchema(schema *apiextensionsv1.JSONSchemaProps) error {
+func (m Maximum) ApplyToSchema(ctx *SchemaContext, schema *apiextensionsv1.JSONSchemaProps) error {
 	if !hasNumericType(schema) {
 		return fmt.Errorf("must apply maximum to a numeric value, found %s", schema.Type)
 	}
@@ -673,7 +677,7 @@ func (m Maximum) ApplyToSchema(schema *apiextensionsv1.JSONSchemaProps) error {
 	return nil
 }
 
-func (m Minimum) ApplyToSchema(schema *apiextensionsv1.JSONSchemaProps) error {
+func (m Minimum) ApplyToSchema(ctx *SchemaContext, schema *apiextensionsv1.JSONSchemaProps) error {
 	if !hasNumericType(schema) {
 		return fmt.Errorf("must apply minimum to a numeric value, found %s", schema.Type)
 	}
@@ -687,7 +691,7 @@ func (m Minimum) ApplyToSchema(schema *apiextensionsv1.JSONSchemaProps) error {
 	return nil
 }
 
-func (m ExclusiveMaximum) ApplyToSchema(schema *apiextensionsv1.JSONSchemaProps) error {
+func (m ExclusiveMaximum) ApplyToSchema(ctx *SchemaContext, schema *apiextensionsv1.JSONSchemaProps) error {
 	if !hasNumericType(schema) {
 		return fmt.Errorf("must apply exclusivemaximum to a numeric value, found %s", schema.Type)
 	}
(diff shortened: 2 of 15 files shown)


pkg/crd/markers/validation.go
```diff
@@ -35,6 +35,9 @@ const (
 	ValidationExactlyOneOfPrefix = validationPrefix + "ExactlyOneOf"
 	ValidationAtMostOneOfPrefix  = validationPrefix + "AtMostOneOf"
 	ValidationAtLeastOneOfPrefix = validationPrefix + "AtLeastOneOf"
+
+	// K8sEnumTag indicates that the given type is an enum; all const values of this type are considered values in the enum
```
JoelSpeed, line 44: @lalitc375 Do you know in DV if there would be a way to gate a new value within an enum based on some combination of `+k8s:ifEnabled`?

pkg/crd/testdata/cronjob_types.go
```diff
@@ -61,6 +61,14 @@ type CronJobSpec struct {
 	// +optional
 	ConcurrencyPolicy ConcurrencyPolicy `json:"concurrencyPolicy,omitempty"`
 
+	// Specifies how to treat concurrent executions of a Job.
+	// Valid values are:
+	// - "Allow" (default): allows CronJobs to run concurrently;
+	// - "Forbid": forbids concurrent runs, skipping next run if previous run hasn't finished yet;
+	// - "Replace": cancels currently running job and replaces it with a new one
+	// +optional
+	K8sConcurrencyPolicy K8sConcurrencyPolicy `json:"k8sConcurrencyPolicy,omitempty"`
```
JoelSpeed, line 70: Are folks likely to put the tag on a field, does the generator error in that case?

pkg/crd/testdata/suite_test.go
```diff
@@ -37,7 +37,11 @@ func TestCRD(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	testEnv = &envtest.Environment{}
+	testEnv = &envtest.Environment{
+		CRDInstallOptions: envtest.CRDInstallOptions{
+			Paths: []string{"testdata.kubebuilder.io_enums.yaml"},
```
JoelSpeed, line 42: What do we do for other test suites? They must also install schemas right?

pkg/crd/schema.go
```diff
@@ -114,6 +114,9 @@ func (c *schemaContext) requestSchema(pkgPath, typeName string) {
 
 // infoToSchema creates a schema for the type in the given set of type information.
 func infoToSchema(ctx *schemaContext) *apiextensionsv1.JSONSchemaProps {
+	if ctx.info.Markers.Get(crdmarkers.K8sEnumTag) != nil {
```
JoelSpeed: This is surprisingly early in the flow of the types to schema conversion

Does this make sense here vs in somewhere like `typeToSchema`? What's the flow for the existing enum marker?

pkg/crd/testdata/cronjob_types.go
alvaroaleman: I've added a field-level marker that will error out

pkg/crd/testdata/suite_test.go
alvaroaleman: No, because they all use unstructured.Unstructured

pkg/crd/schema.go
alvaroaleman: The reason for that was that this marker, contrary to all other ones, requires type info which the markers `ApplyToSchema` didn't have. I've changed it, together with all existing markers, to take it as a new argument

pkg/crd/testdata/enum/api.go
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
camilamacedo86, line 2: The year here should be 2026

pkg/crd/markers/schema_context.go
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright The Kubernetes Authors.
```
camilamacedo86, line 2: Missing the Year

pkg/crd/markers/validation.go
```diff
@@ -1009,3 +1019,71 @@ func fieldsToOneOfCelRuleStr(fields []string) string {
...
+	pkg := ctx.Package
+	info := ctx.TypeInfo
+	typeDef := pkg.TypesInfo.Defs[info.RawSpec.Name]
+	if typeDef == nil {
+		return fmt.Errorf("unknown enum type %s", info.Name)
+	}
+	typeInfo := typeDef.Type()
+	basicInfo, isBasic := typeInfo.Underlying().(*types.Basic)
+	if !isBasic || basicInfo.Info()&types.IsString == 0 {
+		return fmt.Errorf("enum type must be a string, not %s", typeInfo.String())
```
camilamacedo86, line 1049: Do we have any reason for this limitation?
OpenAPI and k8s support all types right?

pkg/crd/markers/validation.go
```diff
@@ -1009,3 +1019,71 @@ func fieldsToOneOfCelRuleStr(fields []string) string {
...
+						continue
+					}
+					// trim quotes
+					value := basicLit.Value[1 : len(basicLit.Value)-1]
+					enumValues = append(enumValues, apiextensionsv1.JSON{Raw: []byte(`"` + value + `"`)})
+				}
+			}
+		}
+	}
+
```
camilamacedo86, line 1081:
```suggestion
  if len(enumValues) == 0 {                                                     
      return fmt.Errorf("no const values found for enum type %s; ensure const   
  declarations match the type exactly", info.Name)                              
  }         
```

It can be empty `schema.Enum = enumValues` right?

pkg/crd/markers/validation.go
```diff
@@ -1009,3 +1019,71 @@ func fieldsToOneOfCelRuleStr(fields []string) string {
...
+					if obj == nil || obj.Type() != typeInfo {
+						continue
+					}
+					val := valueSpec.Values[i]
+					basicLit, ok := val.(*ast.BasicLit)
+					if !ok || basicLit.Kind != token.STRING {
+						continue
+					}
+					// trim quotes
+					value := basicLit.Value[1 : len(basicLit.Value)-1]
```
camilamacedo86, line 1075: It seems that it can break in some use cases 
I think we would need something like

```
  // To work with both "..." and `...`
  value, err := strconv.Unquote(basicLit.Value)                       
  if err != nil {                                                               
      continue  // Skip malformed literals                                      
  }                                                                             

  // To scape JSON nicely here                    
  quotedValue, err := json.Marshal(value)                  
  if err != nil {                                                               
      continue                                                                  
  }                                                                             
  enumValues = append(enumValues, apiextensionsv1.JSON{Raw: quotedValue})
```

pkg/crd/markers/schema_context.go
alvaroaleman: years should not be used anymore going forward, ref https://github.com/kubernetes/steering/issues/299#issuecomment-3434029839

pkg/crd/testdata/enum/api.go
alvaroaleman: I've removed the year, ref https://github.com/kubernetes-sigs/controller-tools/pull/1352#discussion_r3142433485

pkg/crd/markers/validation.go
alvaroaleman: I guess in theory you could use other scalar types but I am not sure if it makes sense to have a non-string enum?

pkg/crd/markers/validation.go
alvaroaleman: added a zero length validation

pkg/crd/markers/validation.go
alvaroaleman: Updated it to use strconv.Unquote and json.Marshal

pkg/crd/markers/validation.go
camilamacedo86: I think the best would be follow OpenAPI and k8s always,
But we can improve in a follow up as well and move incrementally.

camilamacedo86 (approved):
It seems good to get merged.
Just the enum, I think we should follow k8s/openAPI schema types but we can add the scope in a follow up and enhance this one 

So

/lgtm

pkg/crd/markers/validation.go
JoelSpeed: `+kubebuilder:validation:Enum` does not have this restriction, at least, it supports integer enums. For example I know of an API that uses a numeric restriction on the `replicas` field, to limit which values are acceptable. I'm not sure if the `+k8s:enum` supports this upstream in DV though

sbueringer:
@JoelSpeed I think this PR was merged without giving you another chance for a review. Are you fine with the PR as is? (asking because of the upcoming minor release)

JoelSpeed:
My only issue here is that it's currently limited to strings, but the other enum implementation isn't. Need to confirm if DV intended the allow integer enums for example.

We could allow this to ship now and relax the restriction to strings later though

sbueringer:
This seems somewhat related: https://github.com/kubernetes-sigs/controller-tools/pull/1370

PR #1353, opened 2026-02-21
:seedling: Validate we use the same k8s.io/* version in all modules

alvaroaleman:
This change validates that we use the same version of the versioned k8s.io deps in all go modules


sbueringer:
Thx!

/lgtm
/approve

PR #1354, opened 2026-02-22
:sparkles: Add support for k8s:immutable

alvaroaleman:
This change adds support for `k8s:immutable` as described in the [KEP][0].

It replaces https://github.com/kubernetes-sigs/controller-tools/pull/1216 and follows @jpbetz [suggestion](https://github.com/kubernetes-sigs/controller-tools/pull/1216#issuecomment-3001177470) for the CEL expression. It also adds integration tests.

Closes #1216

[0]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -134,6 +134,9 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition(SchemalessName, markers.DescribesField, Schemaless{})).
 		WithHelp(Schemaless{}.Help()),
+
+	must(markers.MakeDefinition("k8s:immutable", markers.DescribesField, Immutable{})).
+		WithHelp(Immutable{}.Help()),
 }
 
 // ValidationIshMarkers are field-and-type markers that don't fall under the
@@ -578,6 +581,33 @@ type XIntOrString struct{}
 //
 // +controllertools:marker:generateHelp:category="CRD validation"
 type Schemaless struct{}
+
+// Immutable marks a field as immutable. Once set, the value cannot be changed.
+// For optional fields, a single transition from unset to set is allowed.
+//
+// Note that immutable fields that are nested below optional fields can still be
+// updated by unsetting the optional parent field and re-setting it again.
+//
+// Examples:
+//
+//	// +k8s:immutable
+//	// +required
+//	Port intstr.IntOrString
+//
+//	// +k8s:immutable
+//	// +optional
+//	TargetPort intstr.IntOrString
+//
+// +controllertools:marker:generateHelp:category="CRD validation"
+type Immutable struct{}
+
+func (m Immutable) ApplyToSchema(_ *SchemaContext, schema *apiextensionsv1.JSONSchemaProps) error {
+	schema.XValidations = append(schema.XValidations, apiextensionsv1.ValidationRule{
+		Rule:    "self == oldSelf",
+		Message: "field is immutable",
+	})
+	return nil
+}
 
 func hasNumericType(schema *apiextensionsv1.JSONSchemaProps) bool {
 	return schema.Type == string(Integer) || schema.Type == string(Number)
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -464,6 +464,8 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiextensio
 		return props
 	}
 
+	var immutableFields []string
+
 	for _, field := range ctx.info.Fields {
 		// Skip if the field is not an inline field, ignoreUnexportedFields is true, and the field is not exported
 		if field.Name != "" && ctx.ignoreUnexportedFields && !ast.IsExported(field.Name) {
@@ -552,11 +554,28 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiextensio
 			continue
 		}
 
+		if field.Markers.Get("k8s:immutable") != nil {
+			immutableFields = append(immutableFields, fieldName)
+		}
+
 		props.Properties[fieldName] = *propSchema
 	}
 
 	// Ensure the required fields are always listed alphabetically.
 	slices.Sort(props.Required)
+
+	// For optional immutable fields, add a parent-level validation rule to prevent
+	// clearing the field once set. The field-level rule prevents value changes, but
+	// when an optional field is removed, the field-level rule doesn't execute.
+	for _, fieldName := range immutableFields {
+		if slices.Contains(props.Required, fieldName) {
+			continue
+		}
+		props.XValidations = append(props.XValidations, apiextensionsv1.ValidationRule{
+			Rule:    fmt.Sprintf("!has(oldSelf.%s) || has(self.%s)", fieldName, fieldName),
+			Message: fmt.Sprintf("field %s is immutable once set", fieldName),
+		})
+	}
 
 	return props
 }
--- a/pkg/crd/testdata/immutable/types.go
+++ b/pkg/crd/testdata/immutable/types.go
@@ -1,0 +1,83 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// +groupName=testdata.kubebuilder.io
+// +versionName=v1
+package immutable
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+type ImmutableType struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ImmutableSpec `json:"spec"`
+}
+
+// +kubebuilder:object:root=true
+type ImmutableTypeList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ImmutableType `json:"items"`
+}
+
+type ImmutableSpec struct {
+	// +k8s:immutable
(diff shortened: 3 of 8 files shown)


pkg/crd/markers/validation.go
```diff
@@ -568,6 +571,35 @@ type XIntOrString struct{}
...
+//	// +optional
+//	TargetPort intstr.IntOrString
+//
+// +controllertools:marker:generateHelp:category="CRD validation"
+type Immutable struct{}
+
+func (m Immutable) ApplyToSchema(schema *apiextensionsv1.JSONSchemaProps) error {
+	optionalOldSelf := true
+	schema.XValidations = append(schema.XValidations, apiextensionsv1.ValidationRule{
+		Rule:            "!oldSelf.hasValue() || self == oldSelf.value()",
```
JoelSpeed: Does this achieve the same thing if we use `self == oldSelf` and drop the `OptionalOldSelf`?

pkg/crd/markers/validation.go
```diff
@@ -568,6 +571,35 @@ type XIntOrString struct{}
 // +controllertools:marker:generateHelp:category="CRD validation"
 type Schemaless struct{}
 
+// Immutable marks a field as immutable. Once set, the value cannot be changed.
+// For optional fields, a single transition from unset to set is allowed.
+//
+// Note that immutable fields that are nested below optional fields can still be
+// updated by unsetting the optional parent field and re-setting it again.
```
JoelSpeed, line 589: Is there anything we can do to handle the "no clear" element of the KEP? 

The usual pattern is to find the nearest required pattern and add a validation to check that if the oldSelf has the immutable field set, that it cannot be unset

pkg/crd/testdata/immutable/types.go
```diff
@@ -0,0 +1,75 @@
...
+	Items           []ImmutableType `json:"items"`
+}
+
+type ImmutableSpec struct {
+	// +k8s:immutable
+	RequiredString string `json:"requiredString"`
+
+	// +k8s:immutable
+	// +optional
+	OptionalString *string `json:"optionalString,omitempty"`
```
JoelSpeed, line 46: Do we need to test a version of this where string isn't a pointer?

pkg/crd/testdata/immutable/types.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// +k8s:immutable
+	RequiredStruct NestedStruct `json:"requiredStruct"`
+
+	// +k8s:immutable
+	// +optional
+	OptionalStruct *NestedStruct `json:"optionalStruct,omitempty"`
```
JoelSpeed, line 57: A version without a pointer, but with omitzero to test that combination?

pkg/crd/testdata/immutable_test.go
```diff
@@ -0,0 +1,192 @@
...
+	DescribeTable("should reject changing an immutable required field",
+		func(ctx SpecContext, name string, mutate, clear func(*unstructured.Unstructured)) {
+			obj := newObj(name)
+			Expect(k8sClient.Create(ctx, obj)).To(Succeed())
+
+			got := obj.DeepCopy()
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(got), got)).To(Succeed())
+
+			mutate(got)
+			Expect(k8sClient.Update(ctx, got)).NotTo(Succeed())
```
JoelSpeed: Can we check the error contains the immutable substring to make sure this is doing what we think it is?

pkg/crd/testdata/immutable_test.go
```diff
@@ -0,0 +1,192 @@
...
+			setInitial(got)
+			Expect(k8sClient.Update(ctx, got)).To(Succeed())
+
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(got), got)).To(Succeed())
+			setChanged(got)
+			Expect(k8sClient.Update(ctx, got)).NotTo(Succeed())
+
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(got), got)).To(Succeed())
+			clear(got)
+			Expect(k8sClient.Update(ctx, got)).NotTo(Succeed())
```
JoelSpeed: I thought the doc said we weren't expecting clear to be rejected?

pkg/crd/schema.go
```diff
@@ -534,12 +536,29 @@ func structToSchema(ctx *schemaContext, structType *ast.StructType) *apiextensio
...
+
 		props.Properties[fieldName] = *propSchema
 	}
 
 	// Ensure the required fields are always listed alphabetically.
 	slices.Sort(props.Required)
 
+	// For optional immutable fields, add a parent-level validation rule to prevent
+	// clearing the field once set. The field-level rule prevents value changes, but
+	// when an optional field is removed, the field-level rule doesn't execute.
```
JoelSpeed, line 569: If this is an optional parent, nothing stops them from removing the optional parent, you need to find the nearest required parent and add it there

JoelSpeed:
/hold

Since this has approve, lets avoid an accidental merge with a fly by lgtm

pkg/crd/schema.go
alvaroaleman: No, the kep explicitly says that its valid to remove an optional parent and that that starts a new cycle: https://github.com/kubernetes/enhancements/blob/c68dfb941894fc8859a951fe47a60b2161300b88/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md#L1143

pkg/crd/markers/validation.go
alvaroaleman: This is handled by the code in schema.go that adds a CEL rule to disallow unsetting once set

pkg/crd/schema.go
JoelSpeed: Fair, I had missed that part of the EP

pkg/crd/testdata/immutable/types.go
alvaroaleman: Sure but at that point we are mostly testing the go serialization?

JoelSpeed: Are we actually using these structs in the tests? If not don't worry

I was just thinking we could example all of the combinations CRD authors might use

pkg/crd/schema.go
alvaroaleman: This surprised me initially as well which is why I left the comment on the marker. For the purpose of implementing the markers, I would just go by the assumption that the relevant discussions if that does or doesn't make sense were already had when the KEP was written/approved

pkg/crd/testdata/immutable_test.go
alvaroaleman: The docs said we do expect clear to be rejected: https://github.com/kubernetes/enhancements/blob/c68dfb941894fc8859a951fe47a60b2161300b88/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md#k8simmutable

Or what do you mean?

pkg/crd/testdata/immutable/types.go
alvaroaleman: Added

pkg/crd/testdata/immutable/types.go
alvaroaleman: Added

pkg/crd/markers/validation.go
alvaroaleman: Yeah, good catch, updated it to do that instead

pkg/crd/testdata/immutable_test.go
alvaroaleman: Done

JoelSpeed:
/lgtm

@alvaroaleman having re-read the EP, I think this is good to go

Would be good if @lalitc375 could ack the behaviour as well, fields are immutable, and cannot be unset unless their parent is cleared

lalitc375:
>Would be good if @lalitc375 could ack the behaviour as well, fields are immutable, and cannot be unset unless their parent is cleared

I am out of office right now. @aaron-prindle Can you please take a look?

lalitc375:
> I am out of office right now. @aaron-prindle Can you please take a look?

I am back, Will review it now.

pkg/crd/markers/validation.go
```diff
@@ -568,6 +571,33 @@ type XIntOrString struct{}
...
+//	// +k8s:immutable
+//	// +optional
+//	TargetPort intstr.IntOrString
+//
+// +controllertools:marker:generateHelp:category="CRD validation"
+type Immutable struct{}
+
+func (m Immutable) ApplyToSchema(schema *apiextensionsv1.JSONSchemaProps) error {
+	schema.XValidations = append(schema.XValidations, apiextensionsv1.ValidationRule{
+		Rule:    "self == oldSelf",
```
lalitc375, line 606: This is not enough. The rule will not be invoked if the field is not set. The rule needs to be set at the field's parent.

alvaroaleman: Yes, that is why this code exists that prevents removal of optional immutable fields: https://github.com/kubernetes-sigs/controller-tools/pull/1354/changes#diff-b0e5a8039953a27d7b1ee61665799f2a1f8681fdcb490f431848bb98c1b1245dR576

This is also tested in `should allow setting an optional field for the first time but reject further changes`. Does that cover the cases you were thinking of?

lalitc375 (commented):
/lgtm

Thank you for implementing it and handling all corner cases.

pkg/crd/markers/validation.go
lalitc375: yes, this handles it. Sorry I missed it earlier.

pkg/crd/markers/validation.go
```diff
@@ -579,6 +582,33 @@ type XIntOrString struct{}
...
+//	// +optional
+//	TargetPort intstr.IntOrString
+//
+// +controllertools:marker:generateHelp:category="CRD validation"
+type Immutable struct{}
+
+func (m Immutable) ApplyToSchema(_ *SchemaContext, schema *apiextensionsv1.JSONSchemaProps) error {
+	schema.XValidations = append(schema.XValidations, apiextensionsv1.ValidationRule{
+		Rule:    "self == oldSelf",
+		Message: "field is immutable",
```
JoelSpeed, line 607: Do we have access to the field name here? Would be preferable for the error message to include the field name IMO

alvaroaleman: The apiserver does this for us already, a sample error message when trying to mutate an immutable field from the tests is
```
ImmutableType.testdata.kubebuilder.io "test-opt-str" is invalid: spec.optionalString: Invalid value: "second-value": field is immutable
```

PR #1376, opened 2026-04-09
🐛Fix "literal not terminated" error for markers with unpaired backticks in Pattern   

### LINKED ISSUE #1084: Generating manifest with backtick in pattern
I have a struct with this annotation. Note the backtick in the pattern

```
// +kubebuilder:validation:items:Pattern=^[`a-zA-Z]+$
```

When generating the manifests I get this error `literal not terminated (at <input>:1:3)`

dongjiang1989:
<!-- please add a icon to the title of this PR (see VERSIONING.md), and delete this line and similar ones -->
<!-- the icon will be either ⚠️ (:warning:, major or breaking changes), ✨ (:sparkles:, feature additions), 🐛 (:bug:, patch and bugfixes), 📖 (:book:, documentation or proposals), or 🌱 (:seedling:, minor or other) -->

<!-- What does this do, and why do we need it? -->

## Summary

Fixes the scanner error "literal not terminated" when parsing markers that contain an unpaired backtick in their pattern values, such as: 
```go                                                                                                                                                    
  // +kubebuilder:validation:items:Pattern=^[`a-zA-Z]+$                                                                                                      
```       

Closes: #1084                                                                                                                      
                                                                                                                                                             
## Changes                                                                                                                                                      
   
  1. `pkg/markers/parse.go` - Modified parserScanner function to:                                                                                                
    - `parserScanner` — replaced the backtick-counting heuristic with a simple rawStrings bool parameter                                                                       
    - `Parse + parse` function — extracted parsing logic into parse(rawMarker, rawStrings) helper; Parse now tries with rawStrings=true first, then retries with false if "literal not  
  terminated" error occurs                                                                                                                                                  
    - hasTerminatedErr — new helper to detect the specific error that triggers fallback                                                                                    
    - guessType — updated two internal parserScanner calls to pass true (preserves original behavior)                                                                                                  
                                                                          
  2. `pkg/markers/parse_internal_test.go` ( add new file) - Added test cases covering:                                                                                
    - Updated all parserScanner calls to include the new rawStrings parameter                                                                                                
    - Fixed type mismatch in assertions (uint vs int)                                                                                                                      
    - Added test case for two literal backticks inside regex character class (addresses camilamacedo86's concern)                                                            
    - Removed misleading "balanced" terminology from comments

### DIFF
--- a/pkg/markers/parse.go
+++ b/pkg/markers/parse.go
@@ -812,7 +812,20 @@ func (d *Definition) loadFields() error {
 func parserScanner(raw string, err func(*sc.Scanner, string)) *sc.Scanner {
 	scanner := &sc.Scanner{}
 	scanner.Init(bytes.NewBufferString(raw))
-	scanner.Mode = sc.ScanIdents | sc.ScanInts | sc.ScanFloats | sc.ScanStrings | sc.ScanRawStrings | sc.SkipComments
+
+	// Base scanning modes we always want enabled.
+	mode := sc.ScanIdents | sc.ScanInts | sc.ScanFloats | sc.ScanStrings | sc.SkipComments
+
+	// ScanRawStrings causes the scanner to treat backticks (`) as raw string delimiters.
+	// If the input contains an unbalanced (odd) number of backticks the scanner will
+	// report "literal not terminated". To avoid that error for annotations that may
+	// include unpaired backticks, only enable ScanRawStrings when the number of
+	// backticks is even (i.e. balanced).
+	if strings.Count(raw, "`")%2 == 0 {
+		mode |= sc.ScanRawStrings
+	}
+
+	scanner.Mode = uint(mode)
 	scanner.Error = err
 
 	return scanner
--- a/pkg/markers/parse.go
+++ b/pkg/markers/parse.go
@@ -847,6 +847,10 @@ type markerParser interface {
 // Parse uses the type information in this Definition to parse the given
 // raw marker in the form `+a:b:c=arg,d=arg` into an output object of the
 // type specified in the definition.
+//
+// Parse attempts to support both raw strings (backtick-delimited) and literal
+// backtick characters by retrying without raw string support if parsing fails
+// with "literal not terminated" error.
 func (d *Definition) Parse(rawMarker string) (any, error) {
 	// Try with ScanRawStrings enabled first (supports raw strings like `pattern`).
 	result, errs := d.parse(rawMarker, true)
@@ -854,7 +858,11 @@ func (d *Definition) Parse(rawMarker string) (any, error) {
 	// may be literal characters (e.g. inside a regex character class), not raw string
 	// delimiters.
 	if hasTerminatedErr(errs) {
-		result, errs = d.parse(rawMarker, false)
+		result, retryErr := d.parse(rawMarker, false)
+		if retryErr != nil {
+			return result, fmt.Errorf("parse failed (tried with and without raw strings): %w", retryErr)
+		}
+		return result, nil
 	}
 	return result, errs
 }
@@ -863,7 +871,10 @@ func hasTerminatedErr(err error) bool {
 	if err == nil {
 		return false
 	}
-	return strings.Contains(err.Error(), "literal not terminated")
+	// Check for the specific scanner error, not just string matching
+	errStr := err.Error()
+	return strings.Contains(errStr, "literal not terminated") ||
+		strings.Contains(errStr, "string not terminated")
 }
 
 func (d *Definition) parse(rawMarker string, rawStrings bool) (any, error) {
--- a/pkg/markers/parse.go
+++ b/pkg/markers/parse.go
@@ -265,7 +265,7 @@ func guessType(scanner *sc.Scanner, raw string, allowSlice bool) *Argument {
 		maybeItem := guessType(scanner, raw, false)
 
 		subRaw := raw[scanner.Pos().Offset:]
-		subScanner := parserScanner(subRaw, scanner.Error, true)
+		subScanner := parserScanner(subRaw, func(*sc.Scanner, string) {}, true)
 
 		var tok rune
 		for {
--- a/pkg/markers/parse.go
+++ b/pkg/markers/parse.go
@@ -830,7 +830,20 @@ func (d *Definition) loadFields() error {
 func parserScanner(raw string, err func(*sc.Scanner, string)) *sc.Scanner {
 	scanner := &sc.Scanner{}
 	scanner.Init(bytes.NewBufferString(raw))
-	scanner.Mode = sc.ScanIdents | sc.ScanInts | sc.ScanFloats | sc.ScanStrings | sc.ScanRawStrings | sc.SkipComments
+
+	// Base scanning modes we always want enabled.
+	mode := sc.ScanIdents | sc.ScanInts | sc.ScanFloats | sc.ScanStrings | sc.SkipComments
+
+	// ScanRawStrings causes the scanner to treat backticks (`) as raw string delimiters.
+	// If the input contains an unbalanced (odd) number of backticks the scanner will
+	// report "literal not terminated". To avoid that error for annotations that may
+	// include unpaired backticks, only enable ScanRawStrings when the number of
+	// backticks is even (i.e. balanced).
+	if strings.Count(raw, "`")%2 == 0 {
+		mode |= sc.ScanRawStrings
+	}
+
+	scanner.Mode = uint(mode)
 	scanner.Error = err
 
 	return scanner
--- a/pkg/markers/parse.go
+++ b/pkg/markers/parse.go
@@ -265,7 +265,7 @@ func guessType(scanner *sc.Scanner, raw string, allowSlice bool) *Argument {
 		maybeItem := guessType(scanner, raw, false)
 
 		subRaw := raw[scanner.Pos().Offset:]
-		subScanner := parserScanner(subRaw, func(*sc.Scanner, string) {})
+		subScanner := parserScanner(subRaw, scanner.Error, true)
 
 		var tok rune
 		for {
@@ -292,7 +292,7 @@ func guessType(scanner *sc.Scanner, raw string, allowSlice bool) *Argument {
 	// go to use this -- Go doesn't like scanners that can be rewound).
 	subRaw := raw[scanner.Pos().Offset:]
 	var hadScanError bool
-	subScanner := parserScanner(subRaw, func(*sc.Scanner, string) { hadScanError = true })
+	subScanner := parserScanner(subRaw, func(*sc.Scanner, string) { hadScanError = true }, true)
 
 	// skip whitespace
 	hint := peekNoSpace(subScanner)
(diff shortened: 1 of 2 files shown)


pkg/markers/parse.go
```diff
@@ -812,7 +812,20 @@ func (d *Definition) loadFields() error {
...
+
+	// Base scanning modes we always want enabled.
+	mode := sc.ScanIdents | sc.ScanInts | sc.ScanFloats | sc.ScanStrings | sc.SkipComments
+
+	// ScanRawStrings causes the scanner to treat backticks (`) as raw string delimiters.
+	// If the input contains an unbalanced (odd) number of backticks the scanner will
+	// report "literal not terminated". To avoid that error for annotations that may
+	// include unpaired backticks, only enable ScanRawStrings when the number of
+	// backticks is even (i.e. balanced).
+	if strings.Count(raw, "`")%2 == 0 {
```
alvaroaleman, line 824: But doesn't that mean we will always allow unpaired backticks, even in situations where that is incorrect?

dongjiang1989: Thanks @alvaroaleman 

This is how the current PR is implemented.
Should we support such unpaired backticks, or throw an error directly?
Personally, I think scenarios with unpaired backticks do exist. WDYT?

pkg/markers/parse.go
```diff
@@ -812,7 +812,20 @@ func (d *Definition) loadFields() error {
...
+		mode |= sc.ScanRawStrings
+	}
+
+	scanner.Mode = uint(mode)
```
camilamacedo86:
```suggestion
	scanner.Mode = mode
```

We can simplify it here.

pkg/markers/parse_internal_test.go
```diff
@@ -0,0 +1,107 @@
...
+				break
+			}
+			tokens = append(tokens, tok)
+		}
+		Expect(tokens).NotTo(BeEmpty(), "should have scanned some tokens")
+	})
+
+	It("should handle kubebuilder validation items pattern with single backtick", func() {
+		// Full marker as it would appear in source code
+		raw := "+kubebuilder:validation:items:Pattern=^[`a-zA-Z]+$"
```
camilamacedo86:
```suggestion
     It("should handle fields string with single backtick", func() {         
                // Just the fields portion (after splitMarker processes the full
                // This is what parserScanner actually receives                 
                raw := "^[`a-zA-Z]+$

```

It seems that the test is incorrect here. Shows like it should test just the pattern  value, not the full marker string.

pkg/markers/parse_internal_test.go
```diff
@@ -0,0 +1,107 @@
...
+			Fail(fmt.Sprintf("scanner error: %s (at %s)", msg, s.Position))
+		}
+
+		scanner := parserScanner(raw, errFunc)
+
+		// With 1 backtick (odd), ScanRawStrings should be disabled
+		Expect(scanner.Mode&sc.ScanRawStrings).To(BeZero(), "ScanRawStrings should be disabled with 1 backtick")
+	})
+
+	It("should handle pattern with two backticks (balanced)", func() {
```
camilamacedo86: Nit: It seems for me that the word “balanced” is misleading here. In the regex `^[a-zA-Z]+$`, the two backticks are not actually paired in any meaningful way for a raw string — they’re just literal characters inside the character class. What about use like `(even but unpaired)`?

Because the heuristic only checks that the number of backticks is even, it will enable ScanRawStrings, but  I think that doesn’t guarantee the string is valid and could still lead to errors.

It might valid either doc the limitation

pkg/markers/parse_internal_test.go
```diff
@@ -0,0 +1,107 @@
...
+		// Pattern with two backticks (balanced)
+		raw := "^[`a-zA-Z`]+$"
+		errFunc := func(s *sc.Scanner, msg string) {
+			Fail(fmt.Sprintf("scanner error: %s (at %s)", msg, s.Position))
+		}
+
+		scanner := parserScanner(raw, errFunc)
+
+		// With 2 backticks (even), ScanRawStrings should be enabled
+		Expect(scanner.Mode&uint(sc.ScanRawStrings)).To(Equal(uint(sc.ScanRawStrings)), "ScanRawStrings should be enabled with even backticks")
```
camilamacedo86:
```suggestion
		Expect(scanner.Mode&sc.ScanRawStrings).To(Equal(sc.ScanRawString
  "ScanRawStrings should be enabled with even backticks")
```

I think we can simplify here as well.

pkg/markers/parse_internal_test.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+
```
camilamacedo86, line 97: I think it is missing a scenario, Are we not only testing the edge case of two unpaired backticks?

pkg/markers/parse.go
```diff
@@ -812,7 +812,20 @@ func (d *Definition) loadFields() error {
 func parserScanner(raw string, err func(*sc.Scanner, string)) *sc.Scanner {
```
camilamacedo86: **I have some concerns about this approach, see:** 

This has two backticks (even), so the heuristic enables `ScanRawStrings`, but these are just literal characters inside the regex — not a real raw string. This can lead to incorrect parsing or different errors. The main issue is that the heuristic can't distinguish between:
                                                                                
  * **Actual raw strings**: `^[a-z]+$` (backticks as delimiters)                
  * **Literal backticks**: ^[`a-z`]+$ (backticks as regex characters)

Example:

```
  ```go                                                                         
  // Pattern: ^[`a-zA-Z`]+$                                 
  // Count: 2 backticks (even) the ScanRawStrings enabled                         
  // Scanner tries to parse: `a-zA-Z` as a raw string 
```
So, It seems that we are not solving the issue right?

**(Proposed Solution)** Therefore, WDYT:

Instead of guessing intent by counting, we can use error recovery to let the
  scanner tell us what works:

1. Try parsing with `ScanRawStrings` enabled
2. If we get a “literal not terminated” error, retry without it

Something like

```
```go
  func (d *Definition) Parse(rawMarker string) (any, error) {
      name, anonName, fields := splitMarker(rawMarker)                          
                                                                                
      // Try with ScanRawStrings enabled (supports raw strings like `pattern`)  
      result, err := d.parseWithScanner(name, anonName, fields, true)           
                                                                                
      // Retry without ScanRawStrings if we get "literal not terminated"
      if err != nil && strings.Contains(err.Error(), "literal not terminated") {
          result, err = d.parseWithScanner(name, anonName, fields, false)       
      }
                                                                                
      return result, err                                    
  }
```


This way we rely on the scanner itself and then I think we have here all addressed:

* Handles all cases correctly (paired, unpaired, quoted, etc.)
* It might make easier to understand how it works ( improve readability )
* I think it will be probably more robust than a heuristic

camilamacedo86 (changes requested):
Thanks for working on this !! Really cool 🎉 

I have a concern about the backtick counting approach. I don’t think it’s very reliable, because counting even/odd backticks doesn’t actually tell us whether they are properly paired.

And I added some comments I think we can refine this one a little better to ensure that we properly address the scenario.

pkg/markers/parse.go
dongjiang1989: Thanks. updated

pkg/markers/parse_internal_test.go
dongjiang1989: Updated

pkg/markers/parse.go
dongjiang1989: Thanks @camilamacedo86 

Please re-check it

Change:                                                                                                                                                                                                                                                                                                   
  1. parserScanner — replaced the backtick-counting heuristic with a simple rawStrings bool parameter                                                                       
  2. Parse + parse — extracted parsing logic into parse(rawMarker, rawStrings) helper; Parse now tries with rawStrings=true first, then retries with false if "literal not  
  terminated" error occurs                                                                                                                                                  
  3. hasTerminatedErr — new helper to detect the specific error that triggers fallback                                                                                    
  4. guessType — updated two internal parserScanner calls to pass true (preserves original behavior)

pkg/markers/parse_internal_test.go
dongjiang1989: Update docs

dongjiang1989:
Thanks @camilamacedo86  Please re-check it

pkg/markers/parse_internal_test.go
```diff
@@ -0,0 +1,150 @@
+/*
```
camilamacedo86, line 1: I think we could add a test like 

```
It("should handle full marker parsing with backtick retry", func() {
      // Test the complete Parse() flow, not just parserScanner                                 
      registry := &Registry{}                                                                   
      def := Must(MakeDefinition("kubebuilder:validation:Pattern", DescribesField, ""))         
      registry.Register(def)                                                                    
                                                                                                
      // This should trigger the retry mechanism internally                                     
      result, err := def.Parse("+kubebuilder:validation:Pattern=^[`a-zA-Z]+$")                  
      Expect(result).NotTo(BeNil())                                                             
  })              
```

That would check both 
Could you please supplement?

pkg/markers/parse.go
```diff
@@ -826,6 +830,25 @@ type markerParser interface {
 // raw marker in the form `+a:b:c=arg,d=arg` into an output object of the
 // type specified in the definition.
 func (d *Definition) Parse(rawMarker string) (any, error) {
```
camilamacedo86, line 854: I think we need improve the description of it now, such as:

```
  // Parse uses the type information in this Definition to parse the given
  // raw marker in the form `+a:b:c=arg,d=arg` into an output object of the                     
  // type specified in the definition.                                                          
  //                                                                                            
  // Parse attempts to support both raw strings (backtick-delimited) and literal                
  // backtick characters by retrying without raw string support if parsing fails                
  // with "literal not terminated" error.
```

pkg/markers/parse.go
```diff
@@ -826,6 +830,25 @@ type markerParser interface {
...
+	// Try with ScanRawStrings enabled first (supports raw strings like `pattern`).
+	result, errs := d.parse(rawMarker, true)
+	// Retry without ScanRawStrings if we get "literal not terminated" — the backticks
+	// may be literal characters (e.g. inside a regex character class), not raw string
+	// delimiters.
+	if hasTerminatedErr(errs) {
+		result, errs = d.parse(rawMarker, false)
```
camilamacedo86: What about we return the error here when fails, such as 

```
  if hasTerminatedErr(errs) {                               
      result, retryErr := d.parse(rawMarker, false)                                             
      if retryErr != nil {                                                                      
          return result, fmt.Errorf("parse failed (tried with and without raw strings): %w",    
  retryErr)                                                                                     
      }                                                     
      return result, nil                                                                        
  } 
```

pkg/markers/parse.go
```diff
@@ -826,6 +830,25 @@ type markerParser interface {
...
+	}
+	return result, errs
+}
+
+func hasTerminatedErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "literal not terminated")
```
camilamacedo86: I think we can improve here such as

```
// Check for the specific scanner error, not just string matching
      errStr := err.Error()                                                                     
      return strings.Contains(errStr, "literal not terminated") ||
             strings.Contains(errStr, "string not terminated") 
```

camilamacedo86 (commented):
That is amazing 🎉 
Great job.

I just add a few nit comments 
After those I am happy to LGTM / approve

dongjiang1989:
Thanks @camilamacedo86 all nits fixed. Please re-check it, thanks

camilamacedo86 (approved):
/lgtm

I am OK with this one. 

We will need a second reviewer
to approve 

Thank you for looking on that 🎉

JoelSpeed:
Left a [comment](https://github.com/kubernetes-sigs/controller-tools/issues/1084#issuecomment-4832222282) last week on the issue. Is escaping a viable fix here?

PR #544, opened 2021-02-11
✨ Add header and year options for yaml generator

nak3:
This patch adds headerFile and year options to CRD, RBAC and Webhook generator

Currently headerFile and year options are available on only generate option.
This patch adds the option for YAMLs.

e.g. generate the crd with headerfile
```
 go run sigs.k8s.io/controller-tools/cmd/controller-gen \
   crd:crdVersions=v1,year=2021,headerFile="${PATH_TO_YOUR_BOILDER_PLATE}" \
   output:crd:artifacts:config=config/crd/bases \
   paths=./...
```

Fix https://github.com/kubernetes-sigs/controller-tools/issues/464

PR #1379, opened 2026-04-18
🐛 (applyconfiguration): applyconfiguration: add a `year` option to pin the copyright year in generated file headers, matching the object generator and avoiding annual diffs on regeneration.

### LINKED ISSUE #1371: Inconsistent handling of YEAR substitution across generators
### Background

https://github.com/kubernetes-sigs/controller-tools/pull/544 added a `year` option to generator commands (`object`, `crd`, `rbac`, and `webhook`), allowing callers to substitute a `YEAR` placeholder in boilerplate header files with a specific value.

### Problem

The `applyconfiguration` generator is missing the `year` option and unconditionally replaces `YEAR` using `time.Now().UTC().Year()`, leading to inconsistent behaviour.

The full call chain is:

1. [`controller-tools/pkg/applyconfiguration/gen.go#L215`](https://github.com/kubernetes-sigs/controller-tools/blob/fbbff17babfacd1ebbbb7a7842bb5378179dadf5/pkg/applyconfiguration/gen.go#L215)
2. [`controller-tools/pkg/applyconfiguration/gen.go#L293`](https://github.com/kubernetes-sigs/controller-tools/blob/fbbff17babfacd1ebbbb7a7842bb5378179dadf5/pkg/applyconfiguration/gen.go#L293)
3. [`code-generator/cmd/applyconfiguration-gen/generators/targets.go#L60`](https://github.com/kubernetes/code-generator/blob/v0.35.0/cmd/applyconfiguration-gen/generators/targets.go#L60)
4. [`gengo/v2/execute.go#L59`](https://github.com/kubernetes/gengo/blob/5ee0d033ba5bcb073f0e69c32057520b82d75ccb/v2/execute.go#L59)

### Proposed Solution

Add a optional `year` option to the `applyconfiguration` generator, which when set, will take precedence. This change would be backwards compatible, as if the option is not specified, the current logic stays as it currently is.

Additionally, the other generators could have same logic, of defaulting to the current year as a subsitution for the string "YEAR" if not overridden by the option.

camilamacedo86:
## What

Add a `year=<value>` option to the `applyconfiguration` generator, matching the existing `object` generator.

## Why

Today, the applyconfiguration generator passes the boilerplate header directly to the `applyconfiguration-gen`, which replaces `YEAR` with the current calendar year at generation time.

For example:

```text
# boilerplate.go.txt
Copyright YEAR The Kubernetes Authors.
```

Without a `year=` option:

- Jan 2024: `make generate` → `Copyright 2024`
- Jan 2025: `make generate` again (even with no API changes) → `Copyright 2025`

This results in a large, unnecessary diff every January.

The `object` generator already supports `year=`, producing deterministic output without annual copyright churn. This change brings the applyconfiguration generator to the same behavior.

Example: 

- **Dec 31** — your generated files say `Copyright 2024`. The working tree is clean.
- **Jan 1** — a teammate (or CI) runs `make generate`. Nothing in the API has changed, but the generator now stamps `Copyright 2025` into every generated applyconfiguration file.
- **Result** — **every generated applyconfiguration file changes, producing a large, meaningless diff caused solely by the calendar rolling over. That is the problem solved here.** 

## How

Before invoking the upstream generator, this change:

1. Reads the boilerplate header.
2. Replaces `YEAR` with the configured `year=` value.
3. Writes the result to a temporary file.
4. Passes the temporary file to the upstream generator.

Because `YEAR` has already been replaced, the upstream generator has nothing left
to substitute with the current calendar year, so the configured year remains
unchanged.

For example, with `year=2024`:

- Jan 2024: `make generate` → `Copyright 2024`
**- Jan 2025: `make generate` → still `Copyright 2024`**

This pins a single year for the entire generation run, matching the behavior of the `object` generator. It does **not** preserve the existing copyright year of individual generated files.

## References

- Discussion: https://github.com/kubernetes-sigs/controller-tools/pull/1379
- Existing `year=` support for `object`: https://github.com/kubernetes-sigs/controller-tools/pull/544
- `gengo` performs the `YEAR` substitution: https://github.com/kubernetes/gengo/blob/5ee0d033ba5bcb073f0e69c32057520b82d75ccb/v2/execute.go#L59
- Kubernetes removed the year to avoid annual churn: https://github.com/kubernetes/kubernetes/pull/59172
- Issue in Kubebuilder https://github.com/kubernetes-sigs/kubebuilder/issues/5553 

Fixes: https://github.com/kubernetes-sigs/controller-tools/issues/1371

### DIFF
--- a/pkg/applyconfiguration/gen.go
+++ b/pkg/applyconfiguration/gen.go
@@ -62,6 +62,9 @@ type Generator struct {
 	// HeaderFile specifies the header text (e.g. license) to prepend to generated files.
 	HeaderFile string `marker:",optional"`
 
+	// Year specifies the year to substitute for " YEAR" in the header file.
+	Year string `marker:",optional"`
+
 	// ExternalApplyConfigurations provides mappings between external types and their applyconfiguration packages.
 	//
 	// Use this to reference apply configuration types for external types referenced
@@ -144,12 +147,25 @@ func isCRD(info *markers.TypeInfo) bool {
 func (d Generator) Generate(ctx *genall.GenerationContext) error {
 	headerFilePath := d.HeaderFile
 
-	if headerFilePath == "" {
+	if d.HeaderFile != "" {
+		headerBytes, err := ctx.ReadFile(d.HeaderFile)
+		if err != nil {
+			return err
+		}
+		headerText := string(headerBytes)
+		headerText = strings.ReplaceAll(headerText, " YEAR", " "+d.Year)
+
 		tmpFile, err := os.CreateTemp("", "applyconfig-header-*.txt")
 		if err != nil {
 			return fmt.Errorf("failed to create temporary file: %w", err)
 		}
+		if _, err := tmpFile.WriteString(headerText); err != nil {
+			tmpFile.Close()
+			os.Remove(tmpFile.Name())
+			return fmt.Errorf("failed to write to temporary file: %w", err)
+		}
 		if err := tmpFile.Close(); err != nil {
+			os.Remove(tmpFile.Name())
 			return fmt.Errorf("failed to close temporary file: %w", err)
 		}
 
--- a/pkg/applyconfiguration/gen.go
+++ b/pkg/applyconfiguration/gen.go
@@ -62,6 +62,9 @@ type Generator struct {
 	// HeaderFile specifies the header text (e.g. license) to prepend to generated files.
 	HeaderFile string `marker:",optional"`
 
+	// Year specifies the year to substitute for " YEAR" in the header file.
+	Year string `marker:",optional"`
+
 	// ExternalApplyConfigurations provides mappings between external types and their applyconfiguration packages.
 	//
 	// Use this to reference apply configuration types for external types referenced
@@ -154,6 +157,30 @@ func (d Generator) Generate(ctx *genall.GenerationContext) error {
 		}
 
 		defer os.Remove(tmpFile.Name())
+
+		headerFilePath = tmpFile.Name()
+	} else {
+		// Substitute the year into the header and write the result to a temp
+		// file, since the generator reads the header from a file.
+		headerBytes, err := ctx.ReadFile(d.HeaderFile)
+		if err != nil {
+			return err
+		}
+		headerText := strings.ReplaceAll(string(headerBytes), " YEAR", " "+d.Year)
+
+		tmpFile, err := os.CreateTemp("", "applyconfig-header-*.txt")
+		if err != nil {
+			return fmt.Errorf("failed to create temporary file: %w", err)
+		}
+		defer os.Remove(tmpFile.Name())
+
+		if _, err := tmpFile.WriteString(headerText); err != nil {
+			_ = tmpFile.Close()
+			return fmt.Errorf("failed to write to temporary file: %w", err)
+		}
+		if err := tmpFile.Close(); err != nil {
+			return fmt.Errorf("failed to close temporary file: %w", err)
+		}
 
 		headerFilePath = tmpFile.Name()
 	}
--- a/pkg/applyconfiguration/year_test.go
+++ b/pkg/applyconfiguration/year_test.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package applyconfiguration
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestYearSubstitution(t *testing.T) {
+	tests := []struct {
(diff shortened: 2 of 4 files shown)


camilamacedo86:
@JoelSpeed I think this one is your baby :-)

alvaroaleman (commented):
/hold

For the discussion in https://github.com/kubernetes-sigs/controller-tools/issues/1371

camilamacedo86:
Hi @alvaroaleman 

But it is problematic, I added in the header and in the issue why it is problematic
The problem is in a project we can have both scenarios, with SSA and not then it does not work well.

Consistent behaviour is a key and expected. 
So, I do see only 2 options here:
- a) Fix and ensure consistency 
- b) No longer support pass the YEAR in any.  (but that would be breaking change and besides k8s no longer care about the YEAR anymore, many end users and projects care so I would not go in this direction)

Vaishnav88sk:
I see this has a `do-not-merge/hold` on it, is there an open question this is blocked on, or just waiting for review? Happy to take a closer look once the hold is resolved.

camilamacedo86:
Hi @Vaishnav88sk 

> I see this has a do-not-merge/hold on it, is there an open question this is blocked on, or just waiting for review? Happy to take a closer look once the hold is resolved.

@alvaroaleman blocked it for discussion: https://github.com/kubernetes-sigs/controller-tools/pull/1379#pullrequestreview-4134954959

As I described in https://github.com/kubernetes-sigs/controller-tools/pull/1379#issuecomment-4275289848 IMO it is a bug and must be done or we should not support the YEAR at all.

pkg/applyconfiguration/gen.go
```diff
@@ -144,12 +147,25 @@ func isCRD(info *markers.TypeInfo) bool {
...
+		headerText := string(headerBytes)
+		headerText = strings.ReplaceAll(headerText, " YEAR", " "+d.Year)
+
 		tmpFile, err := os.CreateTemp("", "applyconfig-header-*.txt")
 		if err != nil {
 			return fmt.Errorf("failed to create temporary file: %w", err)
 		}
+		if _, err := tmpFile.WriteString(headerText); err != nil {
+			tmpFile.Close()
+			os.Remove(tmpFile.Name())
```
JoelSpeed: What if these error?

pkg/applyconfiguration/year_test.go
```diff
@@ -0,0 +1,81 @@
...
+			headerContent:  "Copyright YEAR- YEAR The Kubernetes Authors.",
+			year:           "2024",
+			expectedOutput: "Copyright 2024- 2024 The Kubernetes Authors.",
+		},
+		{
+			name:           "YEAR without space not replaced",
+			headerContent:  "CopyrightYEAR YEAR",
+			year:           "2024",
+			expectedOutput: "CopyrightYEAR 2024",
+		},
```
JoelSpeed: Does this match what DeepCopy generator would do in the upstream code-generator repo?

camilamacedo86: Yes, that is the intention. 
I changed the tests for use gingo.
But we are covering here the scenarios to ensure that it works like object.

Both should work at the same way.

camilamacedo86:
Hi @alvaroaleman @JoelSpeed @sbueringer @felix-kaestner @liggitt,

I've updated the PR description to better explain why the `YEAR` placeholder is needed.

`controller-tools` is consumed by many projects, and the `object` generator already supports `year=`. **This PR simply brings `applyconfiguration` in line with that existing behavior.**

Supporting `year=` avoids the January 1 regeneration issue while remaining fully backwards compatible.

If we decide to remove the `YEAR` placeholder in the future, I think that should be done consistently across all generators and templates, **as a separate breaking change.** This PR only fixes the inconsistency in `applyconfiguration`.

JoelSpeed:
@camilamacedo86 I'm not following how the code enables the files to retain their previous year. As far as I can see the generator has a static year which it dumps into the header file, and then the generator runs and substitutes that one file for all assets it generates

What am I missing?

camilamacedo86:
Hi @JoelSpeed 

Thank you for taking the time to look into this and all your support !!

I've updated both the PR title and desc to make the intent clearer. Hopefully they're easier to follow now. AI also helped me here to wording better. 

So "retain the previous year" was the wrong wording. What this PR actually does is pin a single configured year so regeneration no longer rewrites it to the current calendar year. It does **not** read existing generated files and preserve each file's copyright year. **It make the behaviour consistent accross the project (object and apply works at the same way) and solve the Jan 1 issue.**

Today, the applyconfiguration generator passes the boilerplate header directly to the `applyconfiguration-gen`, which replaces `YEAR` with the current calendar year at generation time.

For example:

```text
# boilerplate.go.txt
Copyright YEAR The Kubernetes Authors.
```

Without the `year=` option, here's the Jan 1 problem concretely:

The `applyconfiguration-gen` replaces `YEAR` with the current calendar year every time it runs. That means the copyright year in generated files is tied to the wall clock, not to when the code actually changed.

- **Dec 31** — your generated files say `Copyright 2024`. The working tree is clean.
- **Jan 1** — a teammate (or CI) runs `make generate`. Nothing in the API has changed, but the generator now stamps `Copyright 2025` into every generated applyconfiguration file.
- **Result** — **every generated applyconfiguration file changes, producing a large, meaningless diff caused solely by the calendar rolling over. That is the problem solved here.** 

For example, with `year=2024`:

- Jan 2024: `make generate` → `Copyright 2024`
- Jan 2025: `make generate` → still `Copyright 2024`

This pins a single year for the entire generation run, matching the behavior of the `object` generator. It does **not** preserve the existing copyright year of individual generated files.

**The goal is simply to achieve parity with the object generator and produce deterministic output.**

PR #1384, opened 2026-04-20
:sparkles: (crd): Add +kubebuilder:skip:description markers to omit package and spec descriptions

### LINKED ISSUE #441: Ability to skip descriptions per field or per package

camilamacedo86:
This adds two new markers to reduce CRD size when types have extensive documentation:        
                                                                                

-   **Spec-level: Skip description for specific specs**                             

```go
  type MyCRD struct {                                                           
      // This description will appear in the CRD                                
      ImportantField string `json:"importantField"`                             
                                                   
      // This long description won't appear in the CRD                          
      // +kubebuilder:skip:description                                     
      LargeField ComplexType `json:"largeField"`                                
  }
```                                                                   
                                                                                
- **Package-level: Skip all descriptions in a package**

```go    
  // +kubebuilder:skip:description                                              
  package v1alpha1                                          
```                                                                                

Useful when CRDs exceed size limits due to extensive godoc comments on deeply nested types.                                                                 
                   
Closes: https://github.com/kubernetes-sigs/controller-tools/issues/441

### DIFF
--- a/pkg/crd/gen_integration_test.go
+++ b/pkg/crd/gen_integration_test.go
@@ -169,6 +169,156 @@ var _ = Describe("CRD Generation proper defaulting", func() {
 		By("comparing the two")
 		Expect(out.buf.String()).To(Equal(string(expectedFile)), cmp.Diff(out.buf.String(), string(expectedFile)))
 	})
+
+	It("should skip spec desc when marker is found", func() {
+		By("switching to test directory")
+		cwd, err := os.Getwd()
+		Expect(err).NotTo(HaveOccurred())
+		defer func() { Expect(os.Chdir(cwd)).To(Succeed()) }()
+
+		skipDescDir := filepath.Join(cwd, "testdata", "skip_desc")
+		Expect(os.Chdir(skipDescDir)).To(Succeed())
+
+		By("loading test packages")
+		pkgs, err := loader.LoadRoots(".")
+		Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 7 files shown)


Vaishnav88sk:
Should this also work at the field level, or is it intentionally package/spec-scoped only? Worth documenting that scope explicitly in the marker definition so users don't go hunting.

camilamacedo86:
Hi @Vaishnav88sk 

> Should this also work at the field level, or is it intentionally package/spec-scoped only? Worth documenting that scope explicitly in the marker definition so users don't go hunting.

But yes, it is the field level as well. 
You can check the description examples

Please, feel free to let me know if the PR description does not clarifies this one.

Vaishnav88sk:
> Hi @Vaishnav88sk
> 
> > Should this also work at the field level, or is it intentionally package/spec-scoped only? Worth documenting that scope explicitly in the marker definition so users don't go hunting.
> 
> Thank you for looking on this one. But yes, it is the field level as well. You can check the description examples
> 
> Please, feel free to let me know if the PR description does not clarifies this one.

Got it, thanks for clarifying! Checked the examples in the description - the field-level scope is clear from there. Makes sense as-is then. 👍

Vaishnav88sk:
LGTM 👍🏻

camilamacedo86:
Hi @sbueringer @JoelSpeed @alvaroaleman 

This one has LGTM from community members, wdyt?
Lets get merged?

PR #1389, opened 2026-04-21
🐛 fix(controller-gen): correctly handle individual file paths (e.g.,paths=./apis/v1/types.go) instead of only accepting directories or recursive patterns.

### LINKED ISSUE #837: controller-gen fails to handle individual file in paths

camilamacedo86:
### What was broken                                                           
                                                            
The loader treated all filesystem paths as directories, causing errors when individual files were specified:
                                                                                
  ```bash                                                   
  # This failed with "not a directory" error
  controller-gen object paths=./apis/v1beta1/some_types.go
                                                                                
  Error: load packages in root ".../some_types.go": err: fork/exec
  /usr/local/go/bin/go: not a directory                                         
  ```
                                                            
### What this fixes                                                               
                                                            
Added file detection logic to distinguish between files and directories. Now individual file paths work correctly:
                                                                                
```                                                                 
  controller-gen object paths=./apis/v1beta1/some_types.go                      
  controller-gen object paths=./apis/v1          # still works                  
  controller-gen object paths=./apis/...         # still works    
```

PS: We are increasing the coverage as well to ensure that it works well with all scenarios. 

Closes: https://github.com/kubernetes-sigs/controller-tools/issues/837

### DIFF
--- a/pkg/loader/loader.go
+++ b/pkg/loader/loader.go
@@ -567,33 +567,58 @@ func LoadRootsWithConfig(cfg *packages.Config, roots ...string) ([]*Package, err
 	//    2. update the loader config's Dir property to be the directory from
 	//       step one
 	//
-	//    3. determine whether the root passed to the loader should be "./."
-	//       or "./..."
+	//    3. determine whether the root passed to the loader should be "./.",
+	//       "./...", or an individual file
 	//
 	//    4. execute the loader with the value from step three
 	for _, r := range fspRoots {
 		b, d := filepath.Base(r), filepath.Dir(r)
 
-		// we want the base part of the path to be either "..." or ".", except
-		// Go's filepath utilities clean paths during manipulation, removing the
-		// ".". thus, if not "...", let's update the path components so that:
-		//
-		//   d = r
-		//   b = "."
-		if b != "..." {
-			d = r
-			b = "."
+		// we want the base part of the path to be either "...", ".", or a filename.
+		// handle "..." first since it's not a real path that can be stat'd
+		if b == "..." {
+			// update the loader configuration's Dir field to the directory part of
+			// the root
+			l.cfg.Dir = d
+
+			// update the root to be "./..."
+			// (with OS-specific filepath separator). please note filepath.Join
+			// would clean up the trailing "." character that we want preserved,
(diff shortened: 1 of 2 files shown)


sbueringer:
Please remove the controller-gen binary from this PR

camilamacedo86:
> Please remove the controller-gen binary from this PR

Good catcher. We need to add it to the .gitignore
I will do that in a follow-up.
Done, it can get merged now.

camilamacedo86:
@sbueringer could we move with this one now?
Any reason for we do not get this one merged?

pkg/loader/loader.go
```diff
@@ -567,33 +567,53 @@ func LoadRootsWithConfig(cfg *packages.Config, roots ...string) ([]*Package, err
...
+			if err != nil {
+				return nil, fmt.Errorf("unable to stat path %q: %w", r, err)
+			}
 
-		// update the root to be "./..." or "./."
-		// (with OS-specific filepath separator). please note filepath.Join
-		// would clean up the trailing "." character that we want preserved,
-		// hence the more manual path concatenation logic
-		r = fmt.Sprintf(".%s%s", string(filepath.Separator), b)
+			if !fi.IsDir() {
```
JoelSpeed: Is everything that stat returns either a file or a dir?

controller-gen
JoelSpeed, line 1: This file should not be checked in

camilamacedo86: It should not. I will remove I added changes in the gitignore to avoid that . I will rebase.

pkg/loader/loader.go
camilamacedo86: Good catcher. I do not think so. We have a gap here. I am working on that.

camilamacedo86: Done 👍

PR #1391, opened 2026-04-26
:sparkles: Upgrade to Kubernetes 1.36.0 and Go 1.26.0

camilamacedo86:
- Update Go version to 1.26.0 (from 1.25.0)
- Update Kubernetes dependencies to v0.36.0 (from v0.35.0)

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -27,7 +27,7 @@ jobs:
         id: vars
         run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
       - name: Set up Go
-        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # tag=v6.3.0
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # tag=v6.4.0
         with:
           go-version: ${{ steps.vars.outputs.go_version }}
       - name: golangci-lint
--- a/.github/workflows/pr-dependabot.yaml
+++ b/.github/workflows/pr-dependabot.yaml
@@ -24,7 +24,7 @@ jobs:
       id: vars
       run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
     - name: Set up Go
-      uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # tag=v6.3.0
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # tag=v6.4.0
       with:
         go-version: ${{ steps.vars.outputs.go_version }}
     - name: Update all modules
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,14 +22,14 @@ jobs:
       id: vars
       run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
     - name: Set up Go
-      uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # tag=v6.3.0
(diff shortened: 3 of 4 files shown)


pkg/applyconfiguration/testdata/cronjob/go.mod
```diff
@@ -1,23 +1,23 @@
...
 require (
 	github.com/onsi/ginkgo/v2 v2.28.1
 	github.com/onsi/gomega v1.39.1
-	k8s.io/api v0.35.0
-	k8s.io/apimachinery v0.35.0
-	k8s.io/client-go v0.35.0
+	k8s.io/api v0.36.0
+	k8s.io/apimachinery v0.36.0
+	k8s.io/client-go v0.36.0
 	sigs.k8s.io/controller-runtime v0.23.1
```
JoelSpeed, line 11: Is there an updated CR release we should be picking up?

pkg/crd/testdata/go.mod
```diff
@@ -1,14 +1,14 @@
...
-	k8s.io/api v0.35.0
-	k8s.io/apiextensions-apiserver v0.35.0
-	k8s.io/apimachinery v0.35.0
-	k8s.io/client-go v0.35.0
+	k8s.io/api v0.36.0
+	k8s.io/apiextensions-apiserver v0.36.0
+	k8s.io/apimachinery v0.36.0
+	k8s.io/client-go v0.36.0
 	sigs.k8s.io/controller-runtime v0.23.1
 	sigs.k8s.io/yaml v1.6.0
```
JoelSpeed, line 13: New release of this to match the 0.36 deps?

pkg/deepcopy/testdata/go.mod
```diff
@@ -1,10 +1,10 @@
 module testdata.kubebuilder.io/cronjob
 
-go 1.25.0
+go 1.26.0
```
JoelSpeed: Do we absolutely need to bump this?

camilamacedo86: We often bump into go and K8S right
Go version should be the same as that used by k8s`
Do you want to do an separated PR?

pkg/applyconfiguration/testdata/cronjob/go.mod
camilamacedo86: We will need to do that in a follow up when we have a CR release with

pkg/crd/testdata/go.mod
camilamacedo86: Yes, we have a validation. See: https://github.com/kubernetes-sigs/controller-tools/pull/1353

pkg/deepcopy/testdata/go.mod
camilamacedo86: Is that really needed to be split??

pkg/applyconfiguration/testdata/cronjob/go.mod
sbueringer: If possible I would not block the CT on the CR release (I think there should be no prod code dependencies from CT to CR?)

pkg/deepcopy/testdata/go.mod
sbueringer: Yes this should be absolutely needed, see: https://github.com/kubernetes/api/blob/v0.36.0/go.mod#L5

camilamacedo86: We must upgrade to be able to upgrade k8s
If you want to split the changes we have: https://github.com/kubernetes-sigs/controller-tools/pull/1402

Just please help us get that merged, then we can rebase this one.

camilamacedo86:
GH close that due a bug it should not be closed. :-(

sbueringer:
You'll have to open a new PR (I also don't have a button to reopen)

Issue #622, opened 2021-09-21
+kubebuilder:default={} does not use nested defaults


astoycos:
So I believe I exposed a bug in kubebuilder... 

when I define a +kubebuilder:default={} as a pointer to another type with +kubebuilder:default values I assume those values should propagate to the parent object. 

For example 

```
// policyAuditConfig is the configuration for network policy audit events. If unset,
// reported defaults are used.
// +kubebuilder:default={}
// +optional
PolicyAuditConfig *PolicyAuditConfig `json:"policyAuditConfig,omitempty"`
```

Where the PolicyAuditConfig is defined in the same package
```
type PolicyAuditConfig struct {
	// rateLimit is the approximate maximum number of messages to generate per-second per-node. If
	// unset the default of 20 msg/sec is used.
	// +kubebuilder:default=20
	// +kubebuilder:validation:Minimum=1
	// +optional
	RateLimit *uint32 `json:"rateLimit,omitempty"`

	// maxFilesSize is the max size an ACL_audit log file is allowed to reach before rotation occurs
	// Units are in MB and the Default is 50MB
	// +kubebuilder:default=50
	// +kubebuilder:validation:Minimum=1
	// +optional
	MaxFileSize *uint32 `json:"maxFileSize,omitempty"`

	// destination is the location for policy log messages.
	// Regardless of this config, persistent logs will always be dumped to the host
	// at /var/log/ovn/ however
	// Additionally syslog output may be configured as follows.
	// Valid values are:
	// - "libc" -> to use the libc syslog() function of the host node's journdald process
	// - "udp:host:port" -> for sending syslog over UDP
	// - "unix:file" -> for using the UNIX domain socket directly
	// - "null" -> to discard all messages logged to syslog
	// The default is "null"
	// +kubebuilder:default=null
	// +kubebuilder:pattern='^libc$|^null$|^udp:(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]):([0-9]){0,5}$|^unix:(\/[^\/ ]*)+([^\/\s])$'
	// +optional
	Destination string `json:"destination,omitempty"`

	// syslogFacility the RFC5424 facility for generated messages, e.g. "kern". Default is "local0"
	// +kubebuilder:default=local0
[...]

PR #1392, opened 2026-04-27
🌱Add envtest tests for +kubebuilder:default={} with nested field defaults

camilamacedo86:
Adds test coverage for the scenarios in #622 (+kubebuilder:default={} on a pointer field whose nested fields have their own defaults)

### DIFF
--- a/pkg/crd/parser_integration_test.go
+++ b/pkg/crd/parser_integration_test.go
@@ -196,6 +196,17 @@ var _ = Describe("CRD Generation From Parsing to CustomResourceDefinition", func
 			})
 		})
 
+		Context("EmptyObjectDefault API", func() {
+			BeforeEach(func() {
+				pkgPaths = []string{"./emptyobjectdefault/..."}
+				expPkgLen = 1
+			})
+			It("should generate default: {} for pointer fields defaulting to an empty object", func() {
+				assertCRD(pkgs[0], "EmptyObjectDefault", "testdata.kubebuilder.io_emptyobjectdefaults.yaml")
+				assertCRD(pkgs[0], "RequiredChildDefault", "testdata.kubebuilder.io_requiredchilddefaults.yaml")
+			})
+		})
+
 		Context("Enum API with marker on field", func() {
 			BeforeEach(func() {
 				pkgPaths = []string{"./enum_error/..."}
--- a/pkg/crd/testdata/cronjob_types.go
+++ b/pkg/crd/testdata/cronjob_types.go
@@ -17,7 +17,7 @@ limitations under the License.
 // TODO(directxman12): test this across both versions (right now we're just
 // trusting k/k conversion, which is probably fine though)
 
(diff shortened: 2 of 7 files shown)


sbueringer:
This seems like a breaking change to me.

Additional I think the CRDs we generate already behave this way as the apiserver recursively defaults top down

camilamacedo86:
Hi @sbueringer,

Sorry, I took a closer look at this, and you're absolutely right to push back.

I also read your comment on the issue, and I agree.

We could add tests to verify the current behavior, but I don't think this warrants keeping the issue open. I'll leave a comment on the issue and suggest we close it.

JoelSpeed:
This doesn't really test our behaviour so much as it tests the API server behaviour, but I guess if that's beneficial to make sure that if the upstream behaviour changes, that we can handle it, then I'm ~ok with merging this

/lgtm

@sbueringer do you agree?

camilamacedo86:
Hi @sbueringer,

Could you please help with this one?

After taking a deeper look and revisiting your comment, I completely agree with your feedback. We're only adding tests here now, as discussed: https://github.com/kubernetes-sigs/controller-tools/issues/622#issuecomment-4827414412

Thank you so much for your help!

sbueringer:
> This doesn't really test our behaviour so much as it tests the API server behaviour,

Agree. controller-gen just takes the default values from the markers and sets them in the corresponding fields in the CRD. I assume we already have sufficient test coverage for this. Accordingly this PR  just adds regression test coverage for the defaulting code in the kube-apiserver. I would assume and hope that this test coverage already exists in the k/k repository and they don't intentionally make breaking changes to the v1 CRD API.

>  but I guess if that's beneficial to make sure that if the upstream behaviour changes, that we can handle it, then I'm ~ok with merging this
> @sbueringer do you agree?

Independent of that, as Joel's okay with it and I don't have a strong objection against it, I"m fine with adding this test coverage.

But can we please change this test to use envtest so it tests actual kube-apiserver behavior? This would not be guaranteed if we directly call the same functions that the apiserver is *currently* calling internally. Additionally the maintenance effort for the test should be lower if we don't have to keep the test code in sync with the kube-apiserver code.

Please also change the PR title to refer to what we are testing vs. just pointing to an issue.

camilamacedo86:
Hi @sbueringer 

Thank you for looking this one.

> But can we please change this test to use envtest so it tests actual kube-apiserver behavior? This would not be guaranteed if we directly call the same functions that the apiserver is currently calling internally. Additionally the maintenance effort for the test should be lower if we don't have to keep the test code in sync with the kube-apiserver code.

That was indeed a good idea and suggestion done !
Rebased with main as well. 
So we are probably good to fly 🪰 here.

sbueringer:
Thx

/lgtm

/assign @JoelSpeed 
(in case you want to take another look)

camilamacedo86:
@JoelSpeed could you give a help here for we get this one merged?

c/c @sbueringer

sbueringer:
I think it's okay, let's merge

/lgtm
/approve

PR #1339, opened 2026-02-15
✨ feat(webhook): Add patch field for webhook customization

camilamacedo86:
Enables filtering webhooks by namespace and object labels to solve the webhook bootstrap problem and support namespace-scoped operators. This markers will allow Kubebuilder abstracts the complexities. 

Closes:  https://github.com/kubernetes-sigs/controller-tools/issues/553 and https://github.com/kubernetes-sigs/controller-tools/issues/1338 and https://github.com/kubernetes-sigs/kubebuilder/issues/5758

PR #1396, opened 2026-04-27
🌱 Migrate to Ginkgo v2 and ensure test temp directories are cleaned up on failure

camilamacedo86:
Motivates by the @alvaroaleman comment: https://github.com/kubernetes-sigs/controller-tools/pull/1339#discussion_r3105659060

We either ensure standards across the project


pkg/applyconfiguration/applyconfiguration_integration_test.go
```diff
@@ -94,6 +93,11 @@ var _ = Describe("ApplyConfiguration generation from API types", func() {
 	AfterEach(func() {
 		// Reset the working directory
 		Expect(os.Chdir(originalCWD)).To(Succeed())
+
+		// Clean up the temporary directory
+		if tmpDir != "" {
+			os.RemoveAll(tmpDir)
+		}
```
JoelSpeed: Have you considered using `DeferCleanup` directly at the point where `tmpDir` is created (L73), that seems to be a neater, more modern pattern for scheduling cleanup of test artifacts

camilamacedo86: You raised a very good point that made me look deeper. DeferCleanup only exists in Ginkgo v2 — and on v1, GinkgoT().TempDir() is a silent no-op, so the cleanup here wasn't working at all. I migrated the module to Ginkgo v2, used DeferCleanup as you suggested, and removed the v1 workarounds. Let me know wdyt.

PR #1398, opened 2026-04-27
🐛 fix(crd): Fix duplicate MinProperties/MaxProperties

### LINKED ISSUE #1234: controller-gen does not deduplicate/flatten MinProperties with embedded struct
controller-gen currently doesn't deduplciate/flatten MinProperties if a struct is embedded in a struct and both have MinProperties set.

Example

```go
// +kubebuilder:validation:MinProperties=1
type MachineHealthCheckTopology struct {
	// +optional
	Enable *bool `json:"enable,omitempty"`

	MachineHealthCheckClass `json:",inline"`
}

// +kubebuilder:validation:MinProperties=1
type MachineHealthCheckClass struct {
	UnhealthyNodeConditions []UnhealthyNodeCondition `json:"unhealthyNodeConditions,omitempty"`
...
}
```

Resulting CRD YAML:
```yaml
                      machineHealthCheck:
                        allOf:
                        - minProperties: 1
                        - minProperties: 1
```

(The resulting  CRD YAML is valid and works, would be just better if we would deduplicate like in other cases)

camilamacedo86:
When a struct with `+kubebuilder:validation:MinProperties=1` embeds another struct with the same setting, the CRD generates duplicates.

Example:

```yaml
allOf:
  - minProperties: 1
  - minProperties: 1   # duplicate
```

This PR fixes this scenario and ensures that the duplication is removed.

Fixes: https://github.com/kubernetes-sigs/controller-tools/issues/1234

Assisted-by: Claude

### DIFF
--- a/pkg/crd/flatten.go
+++ b/pkg/crd/flatten.go
@@ -47,6 +47,89 @@ func isOrNil(val reflect.Value, valInt any, zeroInt any) bool {
 	}
 }
 
+// resolveFieldConflict handles conflicts when both src and dst have values for a field
+func resolveFieldConflict(fieldName string, srcField, dstField reflect.Value, srcInt, dstInt any, errRec ErrorRecorder, srcRemVal, dstRemVal reflect.Value, fieldIndex int) bool {
+	zeroVal := reflect.Zero(srcField.Type())
+
+	switch fieldName {
+	case "Properties":
+		// merge if possible, use all of otherwise
+		srcMap := srcInt.(map[string]apiextensionsv1.JSONSchemaProps)
+		dstMap := dstInt.(map[string]apiextensionsv1.JSONSchemaProps)
+
+		for k, v := range srcMap {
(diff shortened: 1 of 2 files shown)


pkg/crd/flatten.go
```diff
@@ -47,6 +47,89 @@ func isOrNil(val reflect.Value, valInt any, zeroInt any) bool {
 	}
 }
 
+// resolveFieldConflict handles conflicts when both src and dst have values for a field
+func resolveFieldConflict(fieldName string, srcField, dstField reflect.Value, srcInt, dstInt any, errRec ErrorRecorder, srcRemVal, dstRemVal reflect.Value, fieldIndex int) bool {
```
camilamacedo86, line 51: The only change here is where has `// remove duplicate if values are identical`
The rest is == but we need to add all to its func to fix the lint issue (reduces the complexity)

PR #1035, opened 2024-08-14
🐛 Fix duplicate default value when generating CRDs with corev1.Protocol

sbueringer:
The protocol schema was hard-coded, because of that now that the "default" marker works we are ending up with all of.

Simple example:

```go
	// Protocol for port. Must be UDP, TCP, or SCTP.
	// Defaults to "TCP".
	// +optional
	// +default="TCP"
	Protocol corev1.Protocol `json:"protocol,omitempty" protobuf:"bytes,4,opt,name=protocol,casttype=Protocol"`
```

leads to

```yaml
              protocol:
                allOf:
                - default: TCP
                - default: TCP
                description: |-
                  Protocol for port. Must be UDP, TCP, or SCTP.
                  Defaults to "TCP".
                type: string
```

This PR contains 3 commits:
* bump testdata to 1.31.0 (necessary to have the current Protocol type)
  * this one is relatively hard to review. It's easier after dropping jobTemplate (which is just an embedded type) + a bit of re-ordering
* add test coverage for protocol
* drop the embeded schema for protocol

Fixes #1027

Issue #1027, opened 2024-08-09
Error applying a CRD generated by controller-tools that contains the ContainerPort struct


MenD32:
I have the following types.go file:

```
// types.go

type ExampleSpec struct {
	apiv1.ContainerPort `json:",inline"`
}

// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
// +kubebuilder:subresource:status
// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[0].message"
// +kubebuilder:printcolumn:name="Synced",type="string",JSONPath=".status.conditions[0].status"
// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
// +genclient

type Example struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec ExampleSpec	`json:"spec"`
}
```

after generating a CRD from this code with controller-gen like so:
`controller-gen crd:generateEmbeddedObjectMeta=true paths="./pkg/apis/..." output:stdout | tail -n +2 > manifests/crd.yaml` 

and trying to apply with `kubectl create` I get the following error(s):
```
* spec.validation.openAPIV3Schema.properties[spec].properties[protocol].allOf[0].default: Forbidden: must be undefined to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[protocol].allOf[1].default: Forbidden: must be undefined to be structural
```

after some messing around i found that removing the `// +default="TCP"` comment from ContainerPort.Protocol in k8s.io/api, fixes the issue while maintaining the default value of TCP in the CRD itself.

versions:
- Kubernetes v1.30.0
- Controller-gen built from master, hash is b98a19a9c576353e1093ef2bbe4a38701029d381
- k8s.io/api@v0.30.3

aerfio:
It affects https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.16.0-beta.0, works with v0.15.0 in my case

sbueringer:
Another case was reported here: https://github.com/kubernetes-sigs/controller-tools/issues/1034

sbueringer:
I assume the relevant change since v0.15.0 is this one: https://github.com/kubernetes-sigs/controller-tools/pull/938

sbueringer:
Not sure what's going on, clearly looks like a bug:

```go
[...]

PR #1401, opened 2026-04-28
🐛 Dedupe equal Default values in flattenAllOfInto

ravisastryk:
Adds an explicit `case "Default":` so two *apiextensionsv1.JSON pointers with equal raw bytes are deduped instead of hoisted into a fresh allOf, and adds regression tests for the equal-bytes and differing-bytes cases.

Follow-up to #1027 / #1035.

## What does this do, and why do we need it?

Closes the `// TODO(directxman12): Default -- use field?` in `pkg/crd/flatten.go` by giving `Default` an explicit merge case in `flattenAllOfInto`.

Today the `Default` field of `JSONSchemaProps` is `*apiextensionsv1.JSON`. That type is `Comparable` in `reflect`'s sense, but `==` on it compares pointer addresses, not pointee bytes. The equality check at `flatten.go:107`:

```go
if fldTyp.Comparable() && srcInt == dstInt {
    continue // same value
}
```

returns `false` for two distinct allocations that both wrap `"TCP"`. With no explicit case, the field falls into the `default:` branch which hoists both values into a fresh `allOf`, producing structural-schema-violating output:

```yaml
protocol:
  allOf:
  - default: TCP
  - default: TCP
  type: string
```

The API server rejects this with:

```
properties[protocol].allOf[0].default: Forbidden: must be undefined to be structural
properties[protocol].allOf[1].default: Forbidden: must be undefined to be structural
```

## How this relates to #1035

#1035 fixed the only known producer of this shape by removing the hard-coded `corev1.Protocol` schema from `pkg/crd/known_types.go` once `k8s.io/api` started shipping the `+default="TCP"` marker upstream. That fix is preserved here — including the regression test on `CronJobSpec.Protocol` in `pkg/crd/testdata/cronjob_types.go`, which now also serves as a higher-level integration check for this PR.

This PR closes the **merge-logic** half of the problem so the same YAML can't recur from a future `KnownPackages` override, a `+default=`-carrying alias type, or any other code path that puts a default on a referenced type schema.

## Behavior change

`flattenAllOfInto` now handles `Default` explicitly:

- Equal raw bytes → dedupe, keep one, no `allOf`.
- Differing bytes → keep `dst` (parent / field-level value), drop `src`. This matches how `XPreserveUnknownFields` and `XMapType` are already merged: the field-level value wins over the type-level value, since users add `+default=` on a field precisely to override anything the type provides. No error is recorded — silently honoring the override is the less surprising behavior. If reviewers prefer an error here, swapping the final `continue` for `errRec.AddError(...)` is a one-line change.

No public API changes. No marker changes. No testdata bumps. Behavior change is limited to schemas that previously produced `default: ...` inside `allOf`, which were already broken at apply-time, so this can only fix CRDs, not break them.

## Test plan

```
go test ./pkg/crd/...
```

Two new specs in `pkg/crd/flatten_all_of_test.go`:

1. Two `*JSON` defaults with equal raw bytes (the #1027 shape) — expects no `allOf` and a single preserved default.
2. Two `*JSON` defaults with different raw bytes — expects the dst (field-level) value to win and no `allOf`.

The existing `It("should leave properties not in an AllOf branch ... alone")` spec still passes, confirming we haven't regressed the "no merge needed" path.

The `Protocol corev1.Protocol` field added to the cronjob testdata in #1035 continues to flatten to a single `default: TCP` (verified against `pkg/crd/testdata/testdata.kubebuilder.io_cronjobs.yaml`).

## What issue does this fix

Fixes #65

Follow-up to / completes #1027 (originally fixed by #1035).

### DIFF
--- a/pkg/crd/flatten.go
+++ b/pkg/crd/flatten.go
@@ -48,6 +48,96 @@ func isOrNil(val reflect.Value, valInt any, zeroInt any) bool {
 	}
 }
 
+// resolveFieldConflict handles per-field merge logic when both src and dst
+// have non-zero values for fieldName. It returns true if the values were
+// hoisted into the allOf remainders (caller must set hoisted = true).
+func resolveFieldConflict(fieldName string, srcField, dstField reflect.Value, srcInt, dstInt any, errRec ErrorRecorder, srcRemVal, dstRemVal reflect.Value, fieldIndex int) bool {
+	switch fieldName {
+	case "Properties":
+		// merge if possible, use all of otherwise
+		srcMap := srcInt.(map[string]apiextensionsv1.JSONSchemaProps)
+		dstMap := dstInt.(map[string]apiextensionsv1.JSONSchemaProps)
+
+		for k, v := range srcMap {
+			dstProp, exists := dstMap[k]
+			if !exists {
+				dstMap[k] = v
+				continue
+			}
+			flattenAllOfInto(&dstProp, v, errRec)
+			dstMap[k] = dstProp
+		}
+		return false
+	case "Required":
+		// merge
+		dstField.Set(reflect.AppendSlice(dstField, srcField))
+		return false
+	case "Type":
+		if srcInt != dstInt {
+			// TODO(directxman12): figure out how to attach this back to a useful point in the Go source or in the schema
+			errRec.AddError(fmt.Errorf("conflicting types in allOf branches in schema: %s vs %s", dstInt, srcInt))
+		}
+		// keep the destination value, for now
+		return false
+	case "Default":
+		// Default is *apiextensionsv1.JSON, which is comparable by pointer
+		// address rather than by the raw bytes it wraps. The generic
+		// equality short-circuit in the caller therefore cannot detect that
+		// two distinct allocations carry the same JSON value, so without
+		// this case equal defaults arriving from different sources (a
+		// type-level schema with a baked-in default plus a field-level
+		// +default= marker, for example) would be hoisted into a fresh
+		// allOf and produce a structural-schema-violating CRD that the API
+		// server rejects at apply time.
+		//
+		// When the raw bytes match, dedupe and keep one. When they differ,
+		// keep dst (the parent / field-level value) and drop src, matching
+		// how XPreserveUnknownFields and XMapType are merged below.
+		srcDef := srcInt.(*apiextensionsv1.JSON)
+		dstDef := dstInt.(*apiextensionsv1.JSON)
+		if srcDef != nil && dstDef != nil && bytes.Equal(srcDef.Raw, dstDef.Raw) {
+			return false
+		}
+		// values differ: keep dst, drop src.
+		return false
+	// TODO(directxman12):
+	// - Dependencies: if field x is present, then either schema validates or all props are present
+	// - AdditionalItems: like AdditionalProperties
+	// - Definitions: common named validation sets that can be references (merge, bail if duplicate)
+	case "AdditionalProperties":
+		// as of the time of writing, `allows: false` is not allowed, so we don't have to handle it
+		srcProps := srcInt.(*apiextensionsv1.JSONSchemaPropsOrBool)
+		if srcProps.Schema == nil {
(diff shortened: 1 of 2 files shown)


ravisastryk:
@sbueringer @alvaroaleman Please review when you get a chance

pkg/crd/flatten_all_of_test.go
```diff
@@ -396,4 +396,59 @@ var _ = Describe("AllOf Flattening", func() {
 			AllOf:     []apiextensionsv1.JSONSchemaProps{{Pattern: "^[abc]$"}, {Pattern: "^[abcdef]$"}},
 		}))
 	})
+
+	Context("when the same default appears on both the type schema and the field (the #1027 pattern)", func() {
```
camilamacedo86:
```suggestion
	Context("when the same default appears on both the type schema and the field ", func() {
```

I do not think we should add the issue here.
Could we make teh docs on it more clear?

pkg/crd/flatten_all_of_test.go
```diff
@@ -396,4 +396,59 @@ var _ = Describe("AllOf Flattening", func() {
...
+		It("should dedupe equal defaults instead of hoisting them into allOf", func() {
```
camilamacedo86: I think we could simplify, something like

should not allow duplications when .. 

Could we try it out?

pkg/crd/flatten.go
```diff
@@ -134,7 +135,25 @@ func flattenAllOfInto(dst *apiextensionsv1.JSONSchemaProps, src apiextensionsv1.
...
 			// keep the destination value, for now
-		// TODO(directxman12): Default -- use field?
+		case "Default":
+			// Default is *apiextensionsv1.JSON. Two pointers to equal-valued
+			// JSON are "comparable" in reflect's sense (any pointer is) but
+			// compare by address, not by the bytes they wrap. Without this
+			// case, two equal-valued defaults arriving from different sources
+			// (e.g. a type-level schema with a baked-in default plus a
+			// field-level +default= marker) get hoisted into allOf, producing
+			// the structural-schema-violating shape reported in #1027.
```
camilamacedo86: Do we need add the issue?
Can we not doc what/why/when here?

camilamacedo86 (commented):
I have minor suggestions. This PR sort out:

**Example**   

Before PR    
```
  protocol:                                                                                     
    allOf:                                                                                      
    - default: "TCP"                                                                            
    - default: "TCP"  # duplicate!                                                              
    type: string                                            
```

Very similar to the issue fixed in the PR as well: https://github.com/kubernetes-sigs/controller-tools/pull/1398
Also, see that we have lint issues to sort out, that is why I reduced the complexity by creating its own func in:  https://github.com/kubernetes-sigs/controller-tools/pull/1398

It might be easier if we get merged:  https://github.com/kubernetes-sigs/controller-tools/pull/1398, then you do that on top. 


Beyond the nits, it seems great !!!.
Could you please address those, then I am happy to LGTM.

pkg/crd/flatten.go
```diff
@@ -47,6 +48,96 @@ func isOrNil(val reflect.Value, valInt any, zeroInt any) bool {
...
+		// two distinct allocations carry the same JSON value, so without
+		// this case equal defaults arriving from different sources (a
+		// type-level schema with a baked-in default plus a field-level
+		// +default= marker, for example) would be hoisted into a fresh
+		// allOf and produce a structural-schema-violating CRD that the API
+		// server rejects at apply time.
+		//
+		// When the raw bytes match, dedupe and keep one. When they differ,
+		// keep dst (the parent / field-level value) and drop src, matching
+		// how XPreserveUnknownFields and XMapType are merged below.
```
JoelSpeed, line 95: So what does this achieve? It means that if there is a default defined on the type, I can override it at the type's call site (field level) and only the override validation would end up in the schema?

And we are arguing that this is a safe change because today, if you were to have multiple default values that conflict, these are unstructural and the API server won't admit it? Ergo no one could possibly have an valid generation that would change under this change?

PR #1403, opened 2026-04-29
🌱 Add OS-specific files and build artifacts to .gitignore

camilamacedo86:
Improve .gitignore to avoid we commit files that we should not.


v47:
LGTM.
There's a small nit: the description is missing.

pratikmota (approved):
LGTM

PR #1415, opened 2026-05-18
🐛Remove gopkg.in/yaml.v2 and migrate usage to gopkg.in/yaml.v3

chimanjain:
Remove `gopkg.in/yaml.v2` and migrate usage to `gopkg.in/yaml.v3`

Currently we are using both `gopkg.in/yaml.v2`  and `gopkg.in/yaml.v3`, which may lead to inconsistency in it's behavior. This will help bring closure to the above issue.


pkg/genall/genall.go
```diff
@@ -23,7 +23,7 @@ import (
 	"os"
 
 	"golang.org/x/tools/go/packages"
-	rawyaml "gopkg.in/yaml.v2"
+	rawyaml "gopkg.in/yaml.v3"
```
sbueringer, line 27: Will this lead to any changes in the output?

chimanjain: Yes, there is a minor formatting difference in the generated YAML files.

Since the codebase is migrating from yaml.v2 to yaml.v3, the style of sequences (arrays) inside maps is formatted slightly differently by the underlying library:

yaml.v2 (Previous output): List items (prefixed by -) were not indented relative to their parent key.

```yaml
versions:
- name: foo
  schema:
    openAPIV3Schema:
```

yaml.v3 (New output): List items (prefixed by -) are now indented by spaces relative to their parent key.

```yaml
versions:
  - name: foo
    schema:
      openAPIV3Schema:
```
Apart from this list-indentation styling difference, there are no semantic or structural changes to the generated manifests (all keys, values, annotations, and schemas remain exactly the same).

I'll update the integration tests to take in these changes

sbueringer: Makes sense, fine for me once the tests are green

sbueringer:
Please rebase onto main instead of merging main into the PR

chimanjain:
Done, rebased the PR, let me fix the failing ut's, thanks.

chimanjain:
Fixed the ut's some yaml files needs to be regenerated.

sbueringer:
Hm, this produces an annoying amount of churn (not worried about the controller-tools repo but everyone using controller-gen). Is there a way to use the v3 library that avoids that churn?

chimanjain:
> Hm, this produces an annoying amount of churn (not worried about the controller-tools repo but everyone using controller-gen). Is there a way to use the v3 library that avoids that churn?

Yes, you are correct but it won't be an issue:
`gopkg.in/yaml.v3`: Offers cleaner support and alignment with the YAML 1.2 specification (especially regarding boolean representation, null handling, anchors, etc.).
`gopkg.in/yaml.v2`: Largely conforms only to the older YAML 1.1 specification.

Most of the modern systems are supporting the newer specification. And we should make sure we update the gen before they drop the support for older specs. However I don't think there's a solution for that ask. One more observation, we are already using `gopkg.in/yaml.v3` specs in the repository. 
Please guide if we should upgrade the package at remaining remaining places or let it be for now.

sbueringer:
> Yes, you are correct but it won't be an issue:

The issue is that when folks bump controller-gen they will get a huge diff in their YAML files so they'll have a hard time to figure out if something else changed.


@alvaroaleman @JoelSpeed Do you have a take on this?

alvaroaleman:
> @alvaroaleman @JoelSpeed Do you have a take on this?

I don't really see why we would change this if that doesn't solve a concrete problem and I don't see a concrete problem here. Depending on both versions may not be ideal, but its not an actual problem either IMHO

camilamacedo86:
/hold

I think it needs discussion

K8s generally uses `sigs.k8s.io/yaml`.

In `controller-tools`, we can find usage of all three packages:

- `sigs.k8s.io/yaml`: https://github.com/search?q=repo%3Akubernetes-sigs%2Fcontroller-tools+%22sigs.k8s.io%2Fyaml%22&type=code
- `gopkg.in/yaml.v2`: https://github.com/search?q=repo%3Akubernetes-sigs%2Fcontroller-tools+%22gopkg.in%2Fyaml.v2%22&type=code
- `gopkg.in/yaml.v3`: https://github.com/search?q=repo%3Akubernetes-sigs%2Fcontroller-tools+%22gopkg.in%2Fyaml.v3%22&type=code

That said, I would prefer that we stick with `sigs.k8s.io/yaml`.

Following the Kubernetes convention keeps our code consistent with the wider ecosystem and helps avoid unexpected differences in behavior.

We could also enforce this with `golangci-lint` + `depguard`, allowing direct usage of `sigs.k8s.io/yaml` and preventing direct imports of `gopkg.in/yaml.v2` and `gopkg.in/yaml.v3`.

sbueringer:
Would sigs.k8s.io/yaml produce the same output as we produce today? Or a similar amount of churn as gopkg.in/yaml.v3?

camilamacedo86:
Hi @sbueringer @chimanjain @alvaroaleman 

See that both v2 and v3 are no longer maintained.
The project is no longer maintained
We should no longer use any of them.

PR prosing we move out with coverage to ensure that nothing broke with those changes: https://github.com/kubernetes-sigs/controller-tools/pull/1440

chimanjain:
Thanks @camilamacedo86 for opening the new PR, it would be better to move to the maintained repository. If it's ok, I can close this PR and we can review #1440 .

camilamacedo86:
Hi @chimanjain 

> Thanks @camilamacedo86 for opening the new PR, it would be better to move to the maintained repository. If it's ok, I can close this PR and we can review https://github.com/kubernetes-sigs/controller-tools/pull/1440 .

Thank you for raise this one and bring to our attention it.
So, I think we can close this one. 

/close

PR #1416, opened 2026-05-22
✨feat: add non-fatal warning system for code generation diagnostics

### LINKED ISSUE #434: Warn about things that we strongly recommend avoiding
#317 introduces support for a pattern we generally don't recommend: pointers as map values.  This is mainly for a few legacy use cases that folks want to support, but I think we generally want to warn people that this isn't a pattern that we/the kube API conventions generally suggest/support.

We probably want something that looks a lot like our error reporting system, but works adds a "non-fatal" warning instead of a semi-fatal error.  I'd expect something like:

```go
pkg.AddWarning(loader.ErrorFromNode(...))
```

and then print out warnings at the end.

#317 is the only case I know of off the top of my head, but having this infra would prob be useful in the future.

/help
/good-first-issue
/kind feature
/priority important-soon

mattsu2020:
## Summary

This PR implements the warning system requested in #434 by adding a non-fatal warning infrastructure that mirrors the existing error reporting system.

- Add `Package.Warnings` field and `AddWarning()` method to `pkg/loader`, following the same error type dispatch pattern as `AddError()`
- Add `PrintWarnings()` and `VisitPackages()` helper functions to `pkg/loader`
- Add `Runtime.PrintWarnings()` method to `pkg/genall` and call it after successful generation in `controller-gen`
- Emit a warning when pointers are used as map values in CRD schemas (`map[string]*string`), which is supported but not recommended per Kubernetes API conventions
- Add tests verifying that pointer map values produce a warning and non-pointer map values do not

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/loader/...` — 11/11 passed
- [x] `go test ./pkg/crd/...` — all passed (including new `Test_Schema_PointerMapValue_Warning` and `Test_Schema_NonPointerMapValue_NoWarning`)
- [ ] Manual test: run `controller-gen` against a type with `map[string]*string` and verify warning output on stderr

Closes: #434

### DIFF
--- a/cmd/controller-gen/main.go
+++ b/cmd/controller-gen/main.go
@@ -188,6 +188,8 @@ func main() {
 				// don't obscure the actual error with a bunch of usage
 				return noUsageError{fmt.Errorf("not all generators ran successfully")}
 			}
+
+			rt.PrintWarnings()
 			return nil
 		},
 		SilenceUsage: true, // silence the usage, then print it out ourselves if it wasn't suppressed
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -414,6 +414,7 @@ func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *apiextensionsv1.JSON
 	case *ast.ArrayType:
 		valSchema = arrayToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
 	case *ast.StarExpr:
+		ctx.pkg.AddWarning(loader.ErrFromNode(fmt.Errorf("using pointers as map values is not recommended per Kubernetes API conventions"), mapType.Value))
 		valSchema = typeToSchema(ctx.ForInfo(&markers.TypeInfo{}), val)
 	case *ast.MapType:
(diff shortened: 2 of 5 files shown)


sbueringer:
I wonder if producing findings like this should be rather implemented in kube-api-linter

@JoelSpeed WDYT?

mattsu2020:
Thanks

I looked into kube-api-linter and I agree this check fits better there than in controller-tools. There is already a `nomaps` linter that inspects map types and enforces Kubernetes API design guidance around maps, so I think the closest implementation path would be to extend that linter with an option to report map types whose value type is a pointer, e.g. `map[string]*string`.

I can move the actual API-convention check to kube-api-linter, likely as a `nomaps` configuration such as `disallowPointerValues`, and keep this controller-tools PR closed unless maintainers still want a generic non-fatal warning mechanism here.

JoelSpeed:
Agreed, this looks better to be a KAL style change. Lets look at extnending KAL instead of controller tools

@mattsu2020 can we close this?

mattsu2020:
@JoelSpeed  It's fine to close it.
Create a PR on the KAL side.

PR #1419, opened 2026-06-03
:bug: Accept `encoding.TextMarshaler` types as CRD map keys

### LINKED ISSUE #1420: controller-gen: types implementing `encoding.TextMarshaler` are accepted as fields but rejected as map keys
## Situation

A type that implements `encoding.TextMarshaler` is correctly emitted as `type: string` when used as a struct field, but the identical type is rejected with a fatal error when used as a map key:
```
map keys must be strings, not struct{...}
```

Example of an affected struct:
```go
// Addr embeds a netip.Addr, which implements MarshalText/UnmarshalText.
// +kubebuilder:validation:Type=string
type Addr struct{ netip.Addr }

type Spec struct {
    AsField Addr                 `json:"asField"`   // emits type: string
    AsKey   map[Addr]ValueType   `json:"asKey"`     // errors with "map keys must be strings"
}
```

This is an inconsistency, because `controller-gen` already treats any `TextMarshaler` type as a string for fields (in `infoToSchema` [here](https://github.com/kubernetes-sigs/controller-tools/blob/1a246df581e0043d386d82e571bdeeb3d3122ae3/pkg/crd/schema.go#L130)). The only place this rule isn't applied is in `mapToSchema`, which inspects the raw key kind and rejects anything that doesn't resolve to a Go string basic kind.

The emitted CRD schema never represents the key type. A map becomes `type: object + additionalProperties: <valueSchema>`. `controller-gen` emits no `propertyNames` or key schema. So the key's Go type is invisible in the output, the validation gate in `mapToSchema` is the only place the key type matters.

## Proposed fix

In `mapToSchema`, before erroring on a non-string key, check `implements(keyType, textMarshaler)` (the same helper/interface the field path uses) and accept the key if `true`. This is output-neutral, the generated YAML is byte-identical to the string-keyed case, so no existing CRD is affected.

I have a draft PR (#1419) implementing this proposed patch.

HadrienPatte:
`controller-gen` already emits `type: string` for any type implementing `encoding.TextMarshaler` when it is used as a struct field (`infoToSchema`), but `mapToSchema` rejected the identical type as a map key with a fatal `"map keys must be strings"` error. This is an internal inconsistency: the key type is never represented in the generated schema (a map is always `type: object` + `additionalProperties`, with no `propertyNames`), so the validation gate in mapToSchema was the only place the key type mattered.

This PR applies the same `implements(keyType, textMarshaler)` check used by the field path before erroring on a non-string key. The original named/struct key type is tested, not the unwrapped underlying, since the method set lives there. Non-string keys that do not implement `TextMarshaler` still error.

The change is output-neutral: generated YAML is byte-identical to the string-keyed case. It is also runtime-safe, since CRDs are stored as JSON and both `encoding/json` and `sigs.k8s.io/json` marshal map keys through `encoding.Text{Marshaler,Unmarshaler}`.

Closes #1420

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -386,20 +386,31 @@ func arrayToSchema(ctx *schemaContext, array *ast.ArrayType) *apiextensionsv1.JS
 // mapToSchema creates a schema for items of the given map.  Key types must eventually resolve
 // to string (other types aren't allowed by JSON, and thus the kubernetes API standards).
 func mapToSchema(ctx *schemaContext, mapType *ast.MapType) *apiextensionsv1.JSONSchemaProps {
-	keyInfo := ctx.pkg.TypesInfo.TypeOf(mapType.Key)
-	// check that we've got a type that actually corresponds to a string
+	keyType := ctx.pkg.TypesInfo.TypeOf(mapType.Key)
+	// check that we've got a type that actually corresponds to a string, or that
+	// implements encoding.TextMarshaler (in which case it serializes to a string,
+	// just like text-marshaler-implementing field types do).
+	keyInfo := keyType
 	for keyInfo != nil {
 		switch typedKey := keyInfo.(type) {
 		case *types.Basic:
 			if typedKey.Info()&types.IsString == 0 {
-				ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map keys must be strings, not %s", keyInfo.String()), mapType.Key))
+				if implements(keyType, textMarshaler) {
+					keyInfo = nil // stop iterating
+					break
+				}
+				ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map keys must be strings or implement encoding.TextMarshaler, not %s", keyInfo.String()), mapType.Key))
 				return &apiextensionsv1.JSONSchemaProps{}
 			}
 			keyInfo = nil // stop iterating
 		case *types.Named:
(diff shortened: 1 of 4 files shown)


sbueringer:
Thank you very much!

/lgtm

/assign @alvaroaleman @JoelSpeed

pkg/crd/schema.go
```diff
@@ -386,19 +386,30 @@ func arrayToSchema(ctx *schemaContext, array *ast.ArrayType) *apiextensionsv1.JS
...
+	keyInfo := keyType
 	for keyInfo != nil {
 		switch typedKey := keyInfo.(type) {
 		case *types.Basic:
 			if typedKey.Info()&types.IsString == 0 {
+				if implements(keyType, textMarshaler) {
+					keyInfo = nil // stop iterating
+					break
+				}
 				ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("map keys must be strings, not %s", keyInfo.String()), mapType.Key))
```
alvaroaleman: Could we extend this error and the error in the default branch to say "or implement TextMarshaler"?

alvaroaleman (commented):
One small comment, lgtm otherwise

pkg/crd/schema.go
HadrienPatte: 👍 updated both error messages

alvaroaleman (approved):
thanks!

PR #1427, opened 2026-06-10
:sparkles: Add --skip-unchanged flag to controller-gen

### LINKED ISSUE #1425: controller-gen: add --skip-unchanged flag to avoid overwriting files when content has not meaningfully changed
### Motivation

When `controller-gen` runs, it unconditionally overwrites every output file, even when the generated content is identical or nearly identical to what already exists on disk. This causes two problems:

* **Unnecessary file timestamp changes.** Even when the output is byte-identical, the file modification time is updated. Build systems and editors that watch for file modifications (e.g., Make, Bazel, IDE file watchers) are triggered unnecessarily, causing redundant rebuilds and reloads.

* **Version-only diffs across controller-gen upgrades.** CRD manifests include a `controller-gen.kubebuilder.io/version` annotation that reflects the controller-gen build version. Generated Go files include a `Code generated by` header that depends on the controller-gen version. When a team upgrades controller-gen, every generated file shows a diff even though the actual schema, RBAC rules, or deepcopy logic is unchanged. This creates large, noisy pull requests that obscure real changes and complicate code review.

### Proposal

Add an opt-in `--skip-unchanged` flag to `controller-gen`. When enabled:

- Generated output is buffered and compared against existing files before writing.
- Files with byte-identical content are left untouched, preserving their timestamps.
- A line-level diff identifies whether all changes are confined to known auto-generated markers (the version annotation, the `Code generated by` header, or the build tag). Files whose content has not meaningfully changed are left untouched.
- The flag applies to all generators: CRD, deepcopy, RBAC, webhook, and applyconfiguration.

pedjak:
### What does this do, and why do we need it?

When `controller-gen` runs, it unconditionally overwrites every output file, even when the generated content is identical or nearly identical to what already exists on disk. This causes unnecessary timestamp churn (triggering build systems and file watchers) and noisy version-only diffs when upgrading controller-gen.

This PR adds an opt-in `--skip-unchanged` flag. When enabled:

- Generated output is buffered and compared against existing files before writing.
- Files with byte-identical content are left untouched, preserving their timestamps.
- A line-level diff (using `go-difflib`) identifies whether all changes are confined to known auto-generated markers (the `controller-gen.kubebuilder.io/version` annotation, the `Code generated by` header, or the build tag). Files whose content has not meaningfully changed are left untouched.
- The flag applies to all generators: CRD, deepcopy, RBAC, webhook, and applyconfiguration.

Fixes  #1425


camilamacedo86 (commented):
So the go here would NOT update the generated files with the version of controller-tools used when has no changes in the files itself? is that? And it only occurs if the flag is is informed.

pedjak:
> So the go here would NOT update the generated files with the version of controller-tools used when has no changes in the files itself? is that? And it only occurs if the flag is is informed.

Correct.

sbueringer:
/hold

For discussion on the issue

camilamacedo86:
Can we close this one?
I think you will open a new one with a flag like --no-version-annotations,
WDYT?

pedjak:
Closing this PR. Will open a new one that focuses on making version annotations optional rather than skipping unchanged files.

Issue #1431, opened 2026-06-14
controller-gen: add a flag to ignore top level object and type metadata to crd generator


zak905:
Having the required object and type meta as part of the CRD declaration is  optional (and even obsolete) because those fields are automatically assumed for any CRD, when not present. When declaring a golang type for a CRD, we need to add the `metav1.TypeMeta` and `metav1.ObjectMeta` embedded structs:
```
import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

type Test struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
    Spec any `json:"spec"`
        //other fields
}
```

Those fields are useful programatically. However, they are not necessary for the generated CRD definition:

for example:
```
schema:
      openAPIV3Schema:
        properties:
          apiVersion:
            description: the description
            type: string
          kind:
            description: the description
            type: string
          metadata:
            type: object
          spec:
            # spec fields
```
is equivalent to:

```
schema:
      openAPIV3Schema:
        properties:
          spec:
            # spec fields
```

I propose having an additional option for the crd generator that can ignore those object and type meta fields when toggled. For example, the option can be named `IgnoreTopLevelObjectAndTypeMeta`: `controller-gen crd:ignoreTopLevelObjectAndTypeMeta=true paths="./..."`

sbueringer:
I wonder if we should just use one of the `k8s:` markers to express this instead. Maybe `+k8s:opaqueType`?

@JoelSpeed WDYT?

alvaroaleman:
Can you explain what the value/motivation of such a change would be? I'd prefer it if we don't needlessly add options.

zak905:
> Can you explain what the value/motivation of such a change would be? I'd prefer it if we don't needlessly add options.

The value that I see from this is to avoid bloating the CRD definition with some fields that are not necessary. We know by convention that every K8 object has those fields, as mentioned in the api conventions: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#resources 

Additionally, this would alleviate the burden of having to  "fix" the top level metadata: https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/crd/gen.go#L186, which looks likes a hack at first sight. 

> I wonder if we should just use one of the `k8s:` markers to express this instead. Maybe `+k8s:opaqueType`? 

[...]

PR #1432, opened 2026-06-14
:sparkles: add ignore top level object and type metadata flag

zak905:
# What does this do, and why do we need it? 

as mentioned in #1431, this changeset adds an additional option/flag for the crd generator that can ignore those object and type meta fields when toggled. Reason: object and type meta are optional for the CRD definition because those fields are automatically assumed for any CRD, when not present. Please refer to the [issue](#1431) for more details and examples.


sbueringer:
/hold 

for discussion on the issue

sbueringer:
/close
for now until there is a clear way forward identified via the discussion on the issue

PR #1437, opened 2026-06-17
:sparkles: Add option to omit controller-gen version in generated output

### LINKED ISSUE #1436: Make controller-gen version in generated output optional
### Motivation

When `controller-gen` runs, it embeds its version into generated files:

- CRD manifests include a `controller-gen.kubebuilder.io/version` annotation
- ApplyConfiguration Go files include the binary name (which may contain a version suffix) in the `Code generated by` comment header

When a team upgrades controller-gen, every generated file shows a diff even though the actual schema, RBAC rules, or deepcopy logic is unchanged. This creates large, noisy pull requests that obscure real changes and complicate code review.

### Proposal

Add per-generator options that allow users to opt out of including version information in generated output. Both CRD and ApplyConfiguration generators should support this. Default behavior (version included) should remain unchanged.

Ref: #1425, #1427

pedjak:
### What does this do, and why do we need it?

When `controller-gen` runs, it embeds its version into generated files:

- CRD manifests include a `controller-gen.kubebuilder.io/version` annotation
- ApplyConfiguration Go files include the binary name (which may contain a version suffix) in the `Code generated by` comment header

When a team upgrades controller-gen, every generated file shows a diff even though the actual schema, RBAC rules, or deepcopy logic is unchanged. This creates large, noisy pull requests that obscure real changes and complicate code review.

This PR adds per-generator options to opt out of including version information:

- `crd:omitToolVersion=true` — omits the `controller-gen.kubebuilder.io/version` annotation from CRD manifests
- `applyconfiguration:omitToolVersion=true` — normalizes the generator name to `controller-gen` in the `Code generated by` comment, preventing version-suffixed binary names from leaking into output

Both options default to `false`, preserving current behavior.

Fixes #1436


pkg/crd/gen.go
```diff
@@ -98,6 +98,12 @@ type Generator struct {
 	// See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning
 	// for more information about field pruning and v1beta1 resources compatibility.
 	DeprecatedV1beta1CompatibilityPreserveUnknownFields *bool `marker:",optional"`
+
+	// AddToolVersionAnnotation specifies if the controller-gen version annotation
+	// (controller-gen.kubebuilder.io/version) should be added to the CRD.
+	//
+	// Left unspecified, the default is true.
+	AddToolVersionAnnotation *bool `marker:",optional"`
```
camilamacedo86: The current behaviour is to have that by default so should not be skip something like?
I do not think we would like to change the default behaviour.

pedjak: @camilamacedo86 Agreed, the default behavior stays unchanged — the annotation is always added unless you explicitly opt out.

Would you prefer a naming like `SkipToolVersionAnnotation` (default `false`, set to `true` to opt out)? Or do you have another suggestion in mind?

pedjak: Done — renamed to `SkipToolVersionAnnotation` (default `false`, set to `true` to opt out). Also renamed the applyconfiguration field to `SkipToolVersion` for consistency.

pkg/crd/gen.go
```diff
@@ -98,6 +98,12 @@ type Generator struct {
 	// See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning
 	// for more information about field pruning and v1beta1 resources compatibility.
 	DeprecatedV1beta1CompatibilityPreserveUnknownFields *bool `marker:",optional"`
+
+	// SkipToolVersionAnnotation specifies if the controller-gen version annotation
+	// (controller-gen.kubebuilder.io/version) should be skipped when generating the CRD.
+	//
+	// Left unspecified, the default is false.
+	SkipToolVersionAnnotation *bool `marker:",optional"`
```
camilamacedo86: Why not keep the same name for both places ?
Why not `SkipToolVersion` here as well?

WDYT?

pkg/applyconfiguration/gen.go
```diff
@@ -71,6 +71,13 @@ type Generator struct {
 	// For example, to reference the apply configuration for corev1.LocalObjectReference:
 	//   k8s.io/api/core/v1.LocalObjectReference@k8s.io/client-go/applyconfigurations/core/v1
 	ExternalApplyConfigurations []string `marker:",optional"`
+
+	// SkipToolVersion specifies if the tool version should be excluded from the
+	// "Code generated by" comment header. When true, the generator name is
```
camilamacedo86: `helpgen` only uses the first physical line of a field comment as the summary. Since the first sentence wraps onto a second line, the generated summary gets cut off:

```go
Summary: "specifies if the tool version should be excluded from the"
```

That is what `controller-gen applyconfiguration -w` shows, so the help text looks truncated.

Could we make the first line a complete sentence and move the extra explanation into the details in both one? For example:


For CRD

```go
// OmitToolVersion omits the controller-gen version annotation from the generated CRD.
//
// When set, the controller-gen.kubebuilder.io/version annotation is not written
// to the CRD's metadata. This only affects that annotation; the generated schema
// is unchanged. Left unspecified, the default is false.
```

And For Here:

```go
// OmitToolVersion omits the tool version from the "Code generated by" file header.
//
// When set, the generator name in the header is normalized to "controller-gen"
// regardless of the binary name, instead of any version-bearing name. This only
// affects the header comment; no annotations are involved. Left unspecified, the
// default is false.
```

Also, it makes clear what it does.

pkg/applyconfiguration/applyconfiguration_integration_test.go
```diff
@@ -237,6 +237,61 @@ var _ = Describe("ApplyConfiguration generation from API types", func() {
...
+		Expect(hadErrs).To(BeFalse())
+
+		By("Checking the generated file header")
+		generatedFile := filepath.Join("api/v1", applyConfigurationDir, "api/v1/cronjob.go")
+		content, err := os.ReadFile(generatedFile)
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(string(content)).To(HavePrefix("// Code generated by controller-gen. DO NOT EDIT."))
+	})
+})
```
camilamacedo86, line 321: I think this test should also cover the default case.

Right now it only checks that the generated header is:

```go
// Code generated by controller-gen. DO NOT EDIT.
```

That passes because the marker changes the generator name to `controller-gen`. But the test does not prove that the marker caused the change.

Could we add a second case that runs without `skipToolVersion=true` and checks that the header is not `controller-gen`?

```go
It("should leave the generator name unchanged when SkipToolVersion is unset", func() {
    // run without skipToolVersion=true

    Expect(string(content)).NotTo(HavePrefix(
        "// Code generated by controller-gen. DO NOT EDIT.",
    ))
})
```

Then the existing test can keep checking the marker case:

```go
It("should normalize the generator name when SkipToolVersion is true", func() {
    // run with skipToolVersion=true

    Expect(string(content)).To(HavePrefix(
        "// Code generated by controller-gen. DO NOT EDIT.",
    ))
})
```

Together, these prove that `SkipToolVersion` is what changes the generated header, instead of the test passing because of the test binary name.

pkg/applyconfiguration/gen.go
```diff
@@ -169,6 +176,15 @@ func (d Generator) Generate(ctx *genall.GenerationContext) error {
 		externalACs[typeName] = parts[1]
 	}
 
+	// When SkipToolVersion is true, override os.Args[0] so that gengo's
+	// GoBoilerplate (which reads filepath.Base(os.Args[0])) emits a
+	// version-free "Code generated by controller-gen" header.
+	if d.SkipToolVersion != nil && *d.SkipToolVersion {
```
camilamacedo86: WDYT about OmitToolVersion?
Omit seems to be a bette verb, wdyt?

pkg/crd/gen.go
pedjak: Done — renamed to `OmitToolVersion` in both generators for consistency.

pkg/applyconfiguration/gen.go
pedjak: Good catch — restructured the godoc so the first line is a complete sentence in both generators. The generated help summaries now read cleanly.

pkg/applyconfiguration/applyconfiguration_integration_test.go
pedjak: Added a default-case test that verifies the header preserves the original generator name when `OmitToolVersion` is unset. Also added `os.RemoveAll(tmpDir)` to `AfterEach` to clean up the temp directory.

pkg/applyconfiguration/gen.go
pedjak: Agreed — renamed to `OmitToolVersion` everywhere. Much clearer.

camilamacedo86 (commented):
/ok-to-test

pkg/crd/gen_integration_test.go
```diff
@@ -153,6 +153,19 @@ var _ = Describe("CRD Generation proper defaulting", func() {
...
+		yes := true
+		gen := &crd.Generator{
+			CRDVersions:     []string{"v1"},
+			OmitToolVersion: &yes,
+		}
+		Expect(gen.Generate(ctx)).NotTo(HaveOccurred())
+
+		By("searching for version annotation")
+		Expect(out.buf.String()).NotTo(ContainSubstring("controller-gen.kubebuilder.io/version"))
+	})
```
camilamacedo86, line 178: Could we have the other way around here? 
Should add the version annotation by default when omittoolversion is not informed at all.

camilamacedo86:
Hi @pedjak,

> ./hack/verify-boilerplate.sh
> Boilerplate header is wrong for: /home/prow/go/src/sigs.k8s.io/controller-tools/pkg/applyconfiguration/applyconfiguration_integration_test.go

The current error message is not clear. The actual fix is to remove the existing year from the header of applyconfiguration_integration_test.go.

This failure is caused by the new Python check introduced by @sbueringer and @alvaroaleman here:
https://github.com/kubernetes-sigs/controller-tools/pull/1451

@sbueringer @alvaroaleman, honestly, the Python implementation seems more complex than necessary for this requirement, and it does not appear to correctly ignore files that existed before the check was introduced.

The most straightforward fix would be to remove the year from all existing license headers. Otherwise, we will continue to see inconsistent failures in files that were not changed.

The check should also clearly explain that the year must be removed. The current error message does not identify the actual problem.

sbueringer:
> The most straightforward fix would be to remove the year from all existing license headers. Otherwise, we will continue to see inconsistent failures in files that were not changed.

No, we should not modify existing copyright headers to remove the year, see:

> Their guidance is not to modify existing copyright lines in existing files, but to start using the new line for all new file contributions.
> This is because attribution notices shouldn't be removed or replaced, rather just added to, without a very specific and legal reviewed reason.
> https://github.com/kubernetes/steering/issues/299


The issue here is not that the boilerplate verification somehow didn't check this file before. It always verifies all files.

The reason why this check fails now is because we intentionally have different rules for generated and not generated files:
* For generated files the year should be not there
* For not generated files we tolerate the year until 2026, afterwards the year should not be there

The problem here in this PR is that we are introducing the "DO NOT EDIT" string into this file. Because of that the file is now detected as a generated file and accordingly we are now getting a hint that we should remove the year.

I see two options on how to resolve this:
* avoid using "DO NOT EDIT" in the unit test
* add the test file to this list in boilerplate.py (requires a rebase onto main first)
  ```py
  # list all the files contain 'DO NOT EDIT', but are not generated
  skipped_ungenerated_files = ['hack/boilerplate/boilerplate.py']
  ```

pkg/crd/gen_integration_test.go
pedjak: Done.

pedjak:
@camilamacedo86 @sbueringer — rebased onto main and went with option 2.

PR #1440, opened 2026-06-19
🌱 Replace gopkg.in/yaml with sigs.k8s.io/yaml and go.yaml.in/yaml/v3

camilamacedo86:
- Replace gopkg.in/yaml.v2 with sigs.k8s.io/yaml in pkg/genall/genall.go
- Replace gopkg.in/yaml.v3 with go.yaml.in/yaml/v3 in pkg/schemapatcher and its internal yaml package
- Fix `map[any]any` to `map[string]any` type assertions in pkg/crd/gen.go and pkg/genall/genall.go to match JSON-based output
- Add depguard rules to block direct imports of `gopkg.in/yaml.v2` and `gopkg.in/yaml.v3`, with links to the unmaintained upstream and the correct replacements
- Add unit tests for yamlMarshal and TransformRemoveCreationTimestamp covering number types, booleans, nil, arrays, nested maps, and transform chains

Motivation

Both are no longer maintained.
See: https://github.com/go-yaml/yaml


camilamacedo86:
Hi @sbueringer,

This one is pretty straightforward. Could we get it merged?

Could you please take a look and help us out with a review? Thanks!

pkg/genall/genall.go
```diff
@@ -201,8 +198,11 @@ func yamlJSONToYAMLWithFilter(j []byte, options ...*WriteYAMLOptions) ([]byte, e
 		}
 	}
 
-	// Marshal this object into YAML.
-	return rawyaml.Marshal(jsonObj)
+	out, err := json.Marshal(jsonObj)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling into JSON: %w", err)
+	}
+	return sigsyaml.JSONToYAML(out)
```
JoelSpeed: Can you not use the sigsyaml to marshal straight to yaml?

pkg/genall/genall_test.go
```diff
@@ -0,0 +1,205 @@
...
+		out, err := yamlMarshal(obj)
+		g.Expect(err).NotTo(g.HaveOccurred())
+
+		y := string(out)
+		g.Expect(y).To(g.ContainSubstring("name: hello-world"))
+		g.Expect(y).To(g.Or(
+			g.ContainSubstring("value: with colon"),
+			g.ContainSubstring("'value: with colon'"),
+			g.ContainSubstring(`"value: with colon"`),
+		))
```
JoelSpeed: Shouldn't this be predictable?

pkg/genall/genall_test.go
```diff
@@ -0,0 +1,205 @@
...
+		obj := map[string]any{
+			"items": []any{"a", "b", "c"},
+		}
+		out, err := yamlMarshal(obj)
+		g.Expect(err).NotTo(g.HaveOccurred())
+
+		y := string(out)
+		g.Expect(y).To(g.ContainSubstring("- a"))
+		g.Expect(y).To(g.ContainSubstring("- b"))
+		g.Expect(y).To(g.ContainSubstring("- c"))
```
JoelSpeed: Can we not compare to a full single yaml document instead of doing substring matches?

pkg/genall/genall_test.go
```diff
@@ -0,0 +1,205 @@
...
+						"labels": map[string]any{
+							"app": "myapp",
+						},
+					},
+				},
+			},
+		}
+		out, err := yamlMarshal(obj)
+		g.Expect(err).NotTo(g.HaveOccurred())
+		g.Expect(string(out)).To(g.ContainSubstring("app: myapp"))
```
JoelSpeed: Does this truly validate that it marshalled correctly?

pkg/genall/genall_test.go
JoelSpeed, line 1: Could we update this file to use full documents not substring matches, I don't trust the substrings are sufficient

pkg/genall/genall.go
```diff
@@ -23,9 +23,9 @@ import (
 	"os"
 
 	"golang.org/x/tools/go/packages"
-	rawyaml "gopkg.in/yaml.v2"
 	"sigs.k8s.io/controller-tools/pkg/loader"
 	"sigs.k8s.io/controller-tools/pkg/markers"
+	sigsyaml "sigs.k8s.io/yaml"
```
sbueringer: What is the reason for the import alias here?

pkg/genall/genall.go
```diff
@@ -182,14 +185,8 @@ func yamlMarshal(o any, options ...*WriteYAMLOptions) ([]byte, error) {
...
 func yamlJSONToYAMLWithFilter(j []byte, options ...*WriteYAMLOptions) ([]byte, error) {
-	// Convert the JSON to an object.
 	var jsonObj map[string]any
-	// We are using yaml.Unmarshal here (instead of json.Unmarshal) because the
-	// Go JSON library doesn't try to pick the right number type (int, float,
-	// etc.) when unmarshalling to any, it just picks float64
-	// universally. go-yaml does go through the effort of picking the right
-	// number type, so we can preserve number type throughout this process.
-	if err := rawyaml.Unmarshal(j, &jsonObj); err != nil {
+	if err := json.Unmarshal(j, &jsonObj); err != nil {
```
sbueringer: I assume the behavior of the Go SDK JSON package changed here?

The godoc of this func states that this is based on sigs.k8s.io/yaml.JSONToYAML but then it's implemented pretty differently. Why not keep this in sync with sigs.k8s.io/yaml.JSONToYAML?

pkg/genall/genall_test.go
camilamacedo86: done

pkg/genall/genall.go
camilamacedo86: No, encoding/json hasn't changed. Decoding into any still gives float64.

**It just doesn't matter here**. sigs.k8s.io/yaml.Marshal runs json.Marshal, then JSONToYAML. That last step decodes the YAML and puts the number types back, so 8080 stays 8080.

But, we can't call JSONToYAML directly. Transforms need map[string]any, but its YAML decode gives map[any]any for nested maps. Decoding with JSON keeps map[string]any all the way down. That is why we also use UseNumber(). However, it is covered in comments and in tests as well now.

Please, feel free to check and let me know if you have any additional question.

pkg/genall/genall.go
camilamacedo86: Good catcher we can drop

camilamacedo86: Done

camilamacedo86:
Hi @sbueringer and @JoelSpeed 

I addressed all comments and suggestions
Please, feel free to re-check this one.
Hopefully, it is good to get merged now.

sbueringer:
/lgtm

But needs rebase

camilamacedo86:
Hi @sbueringer @JoelSpeed 

it is rebased with main now.
So, we are good to 🪰 

Thank you for your help on this one.

PR #1441, opened 2026-06-19
🐛 pkg/crd: reject pointer-receiver-only TextMarshaler types as map keys

camilamacedo86:
encoding/json marshals map keys by value, not by pointer. Map keys are not addressable in Go's reflect model, so if a type implements MarshalText only via a pointer receiver, encoding/json returns an error at runtime for any map that uses it as a key.

PR #1419 introduced the TextMarshaler map key check using the implements() helper, which accepts both value and pointer receivers. This caused controller-gen to accept types that would fail at runtime. Fix both occurrences in mapToSchema to use types.Implements directly, which checks value receivers only.

The testdata example had the same problem: URL3 implements MarshalText only via a pointer receiver, so map[URL3]string would have failed at runtime. Replace it with a new TextMarshalerKey type that uses a value receiver and actually works. Add a Ginkgo test suite covering all three cases: accepted (value receiver), rejected (pointer receiver only), and rejected (no TextMarshaler at all).

### DIFF
--- a/pkg/crd/schema_textmarshaler_test.go
+++ b/pkg/crd/schema_textmarshaler_test.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crd
+
+import (
+	"go/ast"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	pkgstest "golang.org/x/tools/go/packages/packagestest"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-tools/pkg/loader"
+	testloader "sigs.k8s.io/controller-tools/pkg/loader/testutils"
+	"sigs.k8s.io/controller-tools/pkg/markers"
+)
+
+var _ = Describe("mapToSchema", func() {
+	Context("TextMarshaler map key validation", func() {
+		// cleanup holds the teardown function returned by the fake package loader.
+		// AfterEach calls it so temp directories are removed after every It block.
+		var cleanup func()
+
+		AfterEach(func() {
+			if cleanup != nil {
+				cleanup()
+				cleanup = nil
+			}
+		})
+
+		// schemaForType loads src as a fake package and returns the schema for the
+		// named type, together with the package so callers can inspect errors.
+		// It sets cleanup so AfterEach can release temp files after the It block ends.
+		schemaForType := func(src, typeName string) (*apiextensionsv1.JSONSchemaProps, *loader.Package) {
+			const mod = "sigs.k8s.io/controller-tools/pkg/crd"
+			pkgs, exported, err := testloader.LoadFakeRoots(pkgstest.Modules, []pkgstest.Module{{
+				Name:  mod,
+				Files: map[string]any{"test.go": src},
+			}}, mod)
+			if exported != nil {
+				cleanup = exported.Cleanup
+			}
+			Expect(err).NotTo(HaveOccurred())
+			Expect(pkgs).To(HaveLen(1))
+
+			pkg := pkgs[0]
+			pkg.NeedTypesInfo()
+
+			schemaCtx := newSchemaContext(pkg, nil, true, false).ForInfo(&markers.TypeInfo{})
+
+			var definedType ast.Expr
+			for _, decl := range pkg.Syntax[0].Decls {
+				if gd, ok := decl.(*ast.GenDecl); ok {
+					for _, spec := range gd.Specs {
+						if ts, ok := spec.(*ast.TypeSpec); ok && ts.Name.Name == typeName {
+							definedType = ts.Type
+						}
+					}
+				}
+			}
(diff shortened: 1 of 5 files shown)


camilamacedo86:
Hi @HadrienPatte,

I was reviewing your PR and found a scenario that I think is worth addressing before this reaches users.

The check in `mapToSchema` currently uses the `implements()` helper, which accepts both value and pointer receivers. That works well for struct fields because they are addressable in memory, allowing `encoding/json` to call a pointer-receiver `MarshalText` implementation when needed.

Map keys are a bit different, though. Go's `reflect` package does not allow taking the address of a map key, which means `encoding/json` cannot invoke a pointer-receiver method on them. As a result, if someone defines a type like `URL3` (implementing `MarshalText` only via `*URL3`) and uses it as a map key, `controller-gen` will accept it, but `json.Marshal` will fail at runtime with an unsupported type error.

The test data in the PR currently demonstrates exactly this case. `URL3` implements `encoding.TextMarshaler` only through `*URL3`, so `map[URL3]string` would not actually work at runtime.

To make the behavior align with what `encoding/json` supports, I opened a follow-up PR that:

- Fixes both `implements()` calls in `mapToSchema` to use `types.Implements` (value receivers only)
- Replaces `URL3` in the test data with a new `TextMarshalerKey` type that implements `MarshalText` with a value receiver
- Adds a Ginkgo test that explicitly verifies pointer-receiver-only types are rejected

Thanks for adding this feature — the use case itself makes a lot of sense. I think it just needed this additional guard to ensure we only accept types that will actually work at runtime.

Let me know what you think or if you'd like to discuss any of the details.

@alvaroaleman please, feel free to check as well and let me know wdyt.

pkg/crd/schema_textmarshaler_test.go
```diff
@@ -0,0 +1,126 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
HadrienPatte: maybe set the copyright year to 2026 for a new file?

HadrienPatte (approved):
Thanks for catching that nuance 🙏 

In my PR I used `url.URL3` in the tests which indeed only implements `encoding.TextMarshaler` as a pointer receiver, but in the associated issue #1420, I described the real-life use case that prompted me to come up with that PR which was using a `netip.Addr` as a map key, and that type does implement `encoding.TextMarshaler` as a value receiver.

pkg/crd/schema_textmarshaler_test.go
camilamacedo86: Fixed. Thank you for catching it.

camilamacedo86:
Hi @HadrienPatte

That is great news. IMO the real use case from issue #1420, using netip.Addr as a map key, **is totally fine.** netip.Addr implements TextMarshaler via a value receiver, so it works correctly with json.Marshal and your fix still accepts it.

The problem is the testdata. URL3 only implements TextMarshaler via *URL3, so map[URL3]string would have failed at runtime. I am proposing to change that in this follow-up. Are you ok with it?

I also moved the tests to Ginkgo since it is the project standard. 
I would love to get your help on the review.  🙌 

Thank you so much for the nice contribution 🥇

camilamacedo86:
Hi @sbueringer @alvaroaleman Could we get this one merged to fix the issue ?
The author of the PR @HadrienPatte already help us in the review here and approve this one.

Thx

camilamacedo86:
@JoelSpeed this one has a LGTM/approval from the author already.
WDYT? Could we move forward here?

alvaroaleman:
I think its basically irrelevant if we error here or not, people will notice this doesn't work the first time they try to json marshal their api type

camilamacedo86:
Hi @alvaroaleman and @sbueringer,

I agree that the impact is probably low, since users will eventually get an error when the object is marshaled.

That said, I still don’t see a strong reason to reject this PR.

This change:

* Makes `controller-gen` validation match the actual behavior of the Go JSON encoder.
* Improves the developer experience by reporting the problem where the API type is defined, instead of later at runtime.
* Provides a clearer and more actionable error message.
* Helps catch the issue before it reaches tests, controllers, or the API server.
* Prevents confusing situations where the code compiles and some tests pass, but serialization fails later when a specific code path is exercised.
* Makes the validation more consistent and trustworthy. If `controller-gen` says a type is valid, developers should be able to rely on that.
* Is a small, targeted, and low-risk correctness fix.
* Does not break valid APIs. It only rejects types that cannot be serialized successfully anyway.

Moreover:
- this PR also updates the tests to use Ginkgo, keeping them consistent with the testing style used across the project. Using the same testing framework everywhere improves consistency, readability, and long-term maintainability.
- it is approved for original author as well. https://github.com/kubernetes-sigs/controller-tools/pull/1441#pullrequestreview-4535714309

I see it is as reasonable follow up of PR https://github.com/kubernetes-sigs/controller-tools/pull/1419 to address what we did not catcher in time and to shape it better.  Then, given all of the above, I'm trying to understand what is still preventing this PR from moving forward.

Could you please clarify what the remaining concerns are? Is there something specific in the implementation that should be changed, or is there another reason you don't think this fix should be merged?

**I'd really appreciate understanding what is missing for this PR to receive an approval/LGTM.**

camilamacedo86:
Hi @alvaroaleman and @sbueringer and @JoelSpeed 

I think this one should get merged before the next release as well.
So that we ensure the changes / follow up accordingly. 

Could we get this one merged? WDYT?

sbueringer:
I don't think this PR should block or defer the next CR minor release

pkg/crd/schema_textmarshaler_test.go
```diff
@@ -0,0 +1,126 @@
...
+					Allows: true,
+					Schema: &apiextensionsv1.JSONSchemaProps{Type: "string"},
+				},
+			}))
+		})
+
+		It("rejects a pointer-receiver-only TextMarshaler type as a map key", func() {
+			_, pkg := schemaForType(src, "PtrKeyMap")
+			Expect(pkg.Errors).NotTo(BeEmpty())
+			Expect(pkg.Errors[0].Msg).To(ContainSubstring("map keys must be strings"))
```
JoelSpeed, line 117: What is the full error message here?

PR #1448, opened 2026-07-01
✨ add field level marker "kubebuilder:ignore" to exclude a field from generation

zak905:
This PR was written with the assistance of generative AI

This change set adds a field level marker named "kubebuilder:ignore". As the name implies the field allows ignoring the field during the generation, regardless of whether the field has a `json` tag or no.


alvaroaleman:
/hold
https://github.com/kubernetes-sigs/controller-tools/issues/1431#issuecomment-4856456940

sbueringer:
/close
for now until there is a clear way forward identified via the discussion on the issue

PR #1449, opened 2026-07-05
:warning:  Generate ApplyConfig ExtractFuns when kubebuilder:resource marker is absent

camilamacedo86:
## What is the problem?

Kinds with no `+kubebuilder:resource` marker are not treated as kinds. That includes namespaced kinds with a regular plural, which is the **most common case**. For those kinds, the generator does not create:

- `Extract<Kind>From`
- The OpenAPI schema entry needed by `Extract`

Without these, Server-Side Apply is incomplete. A controller cannot extract the fields it owns before applying changes, making it easy to overwrite fields owned by other field managers.

The `+kubebuilder:resource` marker was never intended to identify apply root types. Its purpose is to configure how a CRD is served (path, scope, shortNames). As a result, projects often have to add markers like:

```go
// +kubebuilder:resource:path=captains
```

even though they do not change the generated CRD. They exist only to make apply generation work.

## Solution

Use the existing type-level `+kubebuilder:ac:generate=true` marker as the canonical signal for apply roots.

This matches how the object generator works. It uses `+kubebuilder:object:root=true` to identify root objects.

The apply generator already has `+kubebuilder:ac:generate=true`, but it was only partially used. It enabled `+genclient` generation, while the OpenAPI schema and `Extract` generation still depended on `+kubebuilder:resource`.

This PR completes that wiring.

When a type has `+kubebuilder:ac:generate=true`, the generator now also creates:

- The OpenAPI schema entry
- The `Extract<Kind>From` function

The `+kubebuilder:resource` marker continues to work as before, so existing projects are unaffected.

## Example

```go
// +kubebuilder:ac:generate=true

// Captain is the Schema for the captains API
type Captain struct {

    Spec CaptainSpec `json:"spec,omitempty"`
}
```

Before this PR, this generated `+genclient` but did not generate an OpenAPI schema entry or `ExtractCaptainFrom`.

After this PR, it generates everything needed for Server-Side Apply, including:

```go
func ExtractCaptainFrom(captain *crewv1.Captain, fieldManager string, subresource string) (*CaptainApplyConfiguration, error)
```

This allows a controller to safely extract, modify, and apply only the fields it owns.

```go
ac, err := ExtractCaptainFrom(captain, "captain-controller", "")
// modify only the fields this controller owns
k8sClient.Apply(ctx, ac, client.FieldOwner("captain-controller"))
```

## Why that matters 

We cannot add the fully support for SSA in kubebuilder (https://github.com/kubernetes-sigs/kubebuilder/pull/5458) as it is today because unless we set any marker like plural indeed when we have no need to use that the full generation does not occur.


camilamacedo86:
Hi @JoelSpeed @sbueringer 

We need this fix to add the support on kubebuilder side. 
It is a canonical and small fix covered with the tests. 
I tried describe well the issue/solution in the PR description.
Could you please give a look on it?

JoelSpeed:
Is the existing functionality usable today, could this result in removing code that is generated already? Or is what's generated today going to always be a subset of what's generated once this fix merges?

I wonder if this needs a more severe emoji in the title ⚠️

camilamacedo86:
Hi @JoelSpeed,

The change here mainly ensures that applyconfiguration generation behaves as expected when an API is explicitly marked with `+kubebuilder:ac:generate=true`.

As I see it, the only impact for existing users would be:

* If I added `+kubebuilder:ac:generate=true` but my API type does not have the `+kubebuilder:resource` marker, the applyconfiguration artifacts are now generated correctly. Before this change, they were not.

Existing supported behavior remains unchanged. This only affects users who explicitly opted in to applyconfiguration generation but were previously not getting the expected output.

So I don't consider this a breaking change. To me, this is a straightforward bug fix that makes the implementation match the behavior implied by `+kubebuilder:ac:generate=true`.

alvaroaleman (approved):
/hold

I think this makes sense to keep the extract func generation consistent with the AC generation. This will result in generating extract funcs when we currently don't though, so I think I agree with Joel wrt marking it as a breaking change (in the sense that when what c-t generates changes without the input changing it is arguably breaking)

alvaroaleman:
Maybe retitle to something like `Generate ApplyConfig ExtractFuns when kubebuilder:resource marker is absent` so its a bit shorter?

camilamacedo86:
Hi @alvaroaleman,

Thank you for looking into this.

Is the hold only to retitle the change using the suggested wording and add a ⚠️ warning?

If so, that has been done.
Could you please cancel the hold?

alvaroaleman:
Looks like the CI failure is legit

camilamacedo86:
Hi @alvaroaleman @JoelSpeed @sbueringer 

Could you please LGTM again?
It broken due the introduction of the check that does not allow YEAR in the header. 

Thank you so much for you support

PR #1451, opened 2026-07-09
✨ Add copyright header validation & fix findings

sbueringer:
Per https://github.com/kubernetes/steering/issues/299, we no longer need to include the year in the copyright header. This PR adds copyright header validation that does not enforce the year.

It forbids adding copyright headers with a year >= 2027 while allowing <= 2026. This seemed easier than detecting newly added files.

based on https://github.com/kubernetes/kubernetes/pull/134835


alvaroaleman (approved):

alvaroaleman (approved):

hack/boilerplate/boilerplate.py
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env python3
```
camilamacedo86, line 1: @sbueringer 

Could we do that in GO?
Why use python?
The project is build in Go and I really think we should keep using only Go and Shell.
It has not really any reason to do anything in python.

sbueringer: This is written in Python because it's almost an exact copy of the same tool in k/k (see the link in the PR description) and various other projects in the Kubernetes ecosystem. I don't see any reason to invest the time and effort to rewrite it in Go or bash

PR #1455, opened 2026-07-24
🌱Bump to k8s.io/* v0.37.0-beta.0

dongjiang1989:


pkg/applyconfiguration/testdata/cronjob/api/v1/testdata.kubebuilder.io_nomarkerjobs.yaml
```diff
@@ -0,0 +1,50 @@
+---
```
sbueringer, line 1: Why is this file being added in this PR?

dongjiang1989: Add new NoMarkerJob CRD YAML from `+kubebuilder:ac:generate=true` fix

sbueringer: Ah. So after that PR merged there was nothing enforcing that this was regenerated, right?

alvaroaleman (approved):
thanks!

pkg/applyconfiguration/testdata/cronjob/api/v1/testdata.kubebuilder.io_nomarkerjobs.yaml
alvaroaleman: Sorry, I missed this thread when reviewing 🤦 But I think its fine? We should ofc fix the fact that we don't seem to validate that generation is current, but that is probably not in scope for this change?

sbueringer: Yup, agree.

dongjiang1989: Happy to follow up with a separate PR for this!

I'll add a `hack/verify-generated.sh` (similar to the existing `verify-k8s-deps.sh`) that runs `go generate ./...` and fails with a clear message if golden files are out of date.

PR #1457, opened 2026-07-27
🌱 Add verify-generated check for golden files

dongjiang1989:
Extract the inline `go generate` + `git diff` check from `test.sh` into a dedicated `hack/verify-generated.sh` script (paired with `hack/update-generated.sh`), following the existing `verify-k8s-deps.sh` / `update-k8s-deps.sh` pattern.

The verify script provides a clear error message when generated files are out of date, telling developers exactly how to fix it: `Run 'hack/update-generated.sh' to update.`

The new `verify-generated` Makefile target is included in `test-all` so it runs automatically in CI for every PR.

This addresses the [review comment on PR #1455](https://github.com/kubernetes-sigs/controller-tools/pull/1455#discussion_r3650396865).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,7 @@ test-all:
 	$(MAKE) verify-modules
 	$(MAKE) verify-boilerplate
 	$(MAKE) verify-k8s-deps
+	$(MAKE) verify-generated
 	$(MAKE) test
 
 .PHONY: modules
@@ -116,6 +117,10 @@ verify-modules: modules ## Verify go modules are up to date
 .PHONY: verify-boilerplate
 verify-boilerplate:
 	./hack/verify-boilerplate.sh
+
+.PHONY: verify-generated
+verify-generated:
+	./hack/verify-generated.sh
 
 ## --------------------------------------
 ## Cleanup / Verification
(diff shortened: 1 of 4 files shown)


dongjiang1989:
cc @sbueringer @alvaroaleman  PTAL, thanks

sbueringer:
Could it be that git diff only covers changes in files that have been already added to git?

I tried to test the verify script by deleting a generated file, committing and then running the script but it didn't fail:

```
$ git status
On branch add-golden-check
Your branch is ahead of 'dongjiang1989/add-golden-check' by 1 commit.
  (use "git push" to publish your local commits)

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        pkg/deepcopy/testdata/zz_generated.deepcopy.go

$ git diff

```

Otherwise lgtm

dongjiang1989:
> Could it be that git diff only covers changes in files that have been already added to git?

Thanks @sbueringer 
Yes, exactly. `git diff` only detects changes in tracked files. Thanks for catching this!

Fixed in the latest commit by replacing `git diff --quiet` with `git status --porcelain`, which detects all changes: modified, new (untracked), and deleted files.

The scenario you described (delete a generated file, run the script) now correctly fails.

sbueringer:
Thx!

/lgtm
/approve

PR #1458, opened 2026-07-29
:bug: inject min boundaries for +k8s:required on strings, maps, and slices

lalitc375:
**What does this do, and why do we need it?**

This PR updates the CRD generation logic for the `+k8s:required` marker to automatically apply minimum boundaries to strings, arrays, and objects.

Specifically, when a field is marked as required, this change will now automatically inject:
*   `minLength=1` for `string` types.
*   `minItems=1` for `array` (slice) types.
*   `minProperties=1` for `object` (map) types.
 
This aligns with the implementation of +k8s:required of validation-gen of kubernetes. https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/validate/required.go

There is  no clean way to model required int in CRD as of now. We will handle it separately.

There are no prod usage outside of k/k for +k8s:required. It is fine to change now then later.  CRs are ratcheted,  It is safe to change it. https://grep.app/search?q=%2Bk8s%3Arequired+

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -114,7 +114,7 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 		WithHelp(markers.SimpleHelp("CRD validation", "specifies that this field is required.")),
 	must(markers.MakeDefinition("optional", markers.DescribesField, struct{}{})).
 		WithHelp(markers.SimpleHelp("CRD validation", "specifies that this field is optional.")),
-	must(markers.MakeDefinition("k8s:required", markers.DescribesField, struct{}{})).
+	must(markers.MakeDefinition("k8s:required", markers.DescribesField, K8sRequired{})).
 		WithHelp(markers.SimpleHelp("CRD validation", "specifies that this field is required.")),
 	must(markers.MakeDefinition("k8s:optional", markers.DescribesField, struct{}{})).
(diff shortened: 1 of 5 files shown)


pkg/crd/markers/validation.go
```diff
@@ -1187,3 +1187,33 @@ func (K8sEnum) ApplyToSchema(ctx *SchemaContext, schema *apiextensionsv1.JSONSch
 	schema.Enum = enumValues
 	return nil
 }
+
+// K8sRequired specifies that this field is required.
+// When applied to a string, it also ensures that the string has a minimum length of 1.
+// When applied to a map, it also ensures that the map has at least 1 property.
+// When applied to a slice, it also ensures that the slice has at least 1 item.
```
JoelSpeed, line 1194: What about a struct? Are we including min properties on a struct for this marker? How's that discussion going upstream?

lalitc375: We decided to allow +k8s:required on a non pointer struct field. We will allow it first in k/k soon after the code thaw.

lalitc375: Add support for +k8s:required in k/k and this PR.

JoelSpeed:
/ok-to-test

CC @sbueringer @alvaroaleman 

This aligns the markers to their upstream behaviour, but is technically a breaking change for our users who might have adopted these already (though they are pretty new). I'm in favour of merging this, so please LMK if you have concerns

PR #1461, opened 2026-08-04
🐛 (webhook): fix patch marker removal semantics

### LINKED ISSUE #1460: The webhook `patch` marker option keeps values the patch replaced
## What happens

`applyPatch` in `pkg/webhook/parser.go` builds the strategic merge patch correctly, then unmarshals the patched JSON back onto the same webhook it just patched. `json.Unmarshal` merges into what is already there, so fields the patch result does not carry survive anyway.

## How to reproduce

Take a marker that generates a rule, and patch `rules` with one entry:

```go
// +kubebuilder:webhook:...,groups=example.com,resources=widgets,verbs=create;update,versions=v1,patch=`{"rules":[{"apiGroups":["batch"]}]}`
```

Generated:

```yaml
rules:
- apiGroups: [batch]
  apiVersions: [v1]        # from the marker, should be gone
  operations: [CREATE, UPDATE]  # from the marker, should be gone
  resources: [widgets]     # from the marker, should be gone
```

Expected, because `rules` is an atomic list, so the patch replaces it:

```yaml
rules:
- apiGroups: [batch]
```

The same cause makes `patch=`{"timeoutSeconds": null}`` a no-op: strategic merge patch treats `null` as "remove this field", but the decode puts the old value back, so no field can ever be unset.

## Suggested fix

Unmarshal the patched JSON into a fresh webhook value and assign it, instead of decoding onto the populated one. `applyPatch` becomes generic over the webhook type so it can declare that value:

```go
func applyPatch[T any](webhook *T, patchStr string) error {
	// ... marshal, StrategicMergePatch as today ...
	var patched T
	if err := json.Unmarshal(patchedJSON, &patched); err != nil {
		return fmt.Errorf("failed to unmarshal patched webhook: %w", err)
	}
	*webhook = patched
	return nil
}
```

Worth a regression test that patches `rules` on a webhook whose marker already generated a rule, and asserts the patched entry carries nothing from the old one.

camilamacedo86:
Follow-up of: #1339 (**unreleased**)

Fix the webhook `patch` marker: values removed by the patch no longer reappear in the generated manifest, and unknown or misspelled fields in the patch now return an error instead of being silently ignored.

 ### What changed
  - Values removed by the patch (null fields, replaced lists) no longer leak back into the generated manifest.
  - Unknown or miscased fields in the patch now fail generation instead of being silently dropped.
  - Document the semantics and cover them with unit and golden tests.
  
  **Note:** All tests are written following the project standards (Ginkgo
  and Gomega).
  
  Fixes #1460


camilamacedo86:
Hi @sbueringer @JoelSpeed 

Could you please give a hand here?
This one is important for end users and it would need to get merged before the next release to ensure that the new feature lands accurate. Thank you so much for your time and help.

pkg/webhook/parser.go
```diff
@@ -256,32 +265,107 @@ func verbToAPIVariant(verbRaw string) admissionregv1.OperationType {
...
+				}
+			}
+			return nil
+		}
+		if t.Kind() != reflect.Struct {
+			return nil
+		}
+		for key, val := range v {
+			// Keys starting with $ are strategic merge patch directives, such as $patch: delete.
+			if strings.HasPrefix(key, "$") {
```
v47: The `$` exception is too broad. `validatePatchFields` currently accepts every key beginning with `$`, so a typo such as "$patc" or an unknown directive such as "$unknown" can still be silently ignored after the patched JSON is unmarshaled into the webhook.
Since this PR explicitly guarantees that unknown or misspelt patch fields fail generation, could we allow only the strategic-merge directives actually supported by the library and validate field-name suffixes where applicable? A regression test for `{"$patc":"replace"}` would cover the ordinary typo case.

camilamacedo86: Good catch, thanks. I removed the $ prefix exception and all the hand-rolled validation.

Now we just rely on the k8s libs: the patch is applied with StrategicMergePatch, and the result is decoded with sigs.k8s.io/json.UnmarshalStrict — the same strict decoder the API server uses. Any typo, like $patc or a miscased field, fails generation with a clear error. Nothing to keep in sync with apimachinery on our side.

Added a regression test for {"$patc":"replace"} plus tests for the other typo and directive cases.

v47:
The strict decoder fixes the `$patc` case, but I think validating only the patched result still leaves a silent-ignore path.
`StrategicMergePatch` applies with `IgnoreUnmatchedNulls: true`, so an unknown field set to null is removed before `UnmarshalStrict` sees it. For example, `{"notAField":null}` or the `miscased {"TimeoutSeconds":null}` appears to produce an unchanged valid webhook rather than an error. The same general issue may apply to unknown fields containing self-removing directives such as `{"notAField":{"$patch":"delete"}}.` Could we validate the patch keys before applying the merge, while explicitly accounting for supported strategic-merge directives? Regression tests for the two null cases would demonstrate the issue.

pkg/webhook/parser_test.go
```diff
@@ -17,290 +17,352 @@ limitations under the License.
...
-				Resources:               []string{"deployments"},
-				Versions:                []string{"v1"},
-				Verbs:                   []string{"create"},
-				AdmissionReviewVersions: []string{"v1"},
-				Patch:                   `{invalid json`,
-			},
-			expectError: true,
-		},
-	}
+		It("should treat an empty JSON object as a no-op", func() {
```
sbueringer: This looks like an error case to me. What is the use case for setting a marker with an empty patch?

pkg/webhook/parser_test.go
```diff
@@ -17,290 +17,352 @@ limitations under the License.
...
-				return
+			By("applying an empty JSON object")
+			Expect(applyPatch(webhook, `{}`)).To(Succeed())
+
+			By("verifying the webhook is unchanged")
+			Expect(webhook.Name).To(Equal("test-webhook"))
+			Expect(webhook.TimeoutSeconds).To(HaveValue(Equal(int32(10))))
+		})
+
+		It("should treat a whitespace-only patch as a no-op", func() {
```
sbueringer: Same here

pkg/webhook/parser_test.go
camilamacedo86: Good point. The test is intentional because this marker must preserve Kubernetes strategic-merge-patch semantics. {} is a valid Kubernetes strategic merge patch and is defined as a no-op. Although omitting the
marker is more practical, **keeping this test ensures our behavior remains consistent with Kubernetes upstream.**

pkg/webhook/parser_test.go
camilamacedo86: Whitespace-only input is invalid JSON, so Kubernetes’ strategic merge patch returns an error. An omitted marker (patch="") remains a no-op. That is what we covering here.

camilamacedo86:
Hi @sbueringer, @JoelSpeed 

The past two weeks have been particularly challenging at work, so I wasn’t able to address this sooner. However, I’ve now updated the PR. The goals of this PR are:

 1. Ensure that strategic merge patch behavior is implemented correctly and
   remains consistent with Kubernetes.

 2. Ensure that the tests follow the project standards by using Ginkgo and
   Gomega.

We did not bump k8s 1.37 yet so I think, if possible, it is good enough now and we should merge before the release to avoid any breaking change discussion behaviour.
If you have a time to give a look I will really appreciate your help. Lets see.

We’re very close to completing this. Thank you!

camilamacedo86:
@v47 @sbueringer could you please give a help on this one?
Thank you so much. !!

pkg/webhook/parser_test.go
sbueringer: I still don't understand why someone would set a marker with an empty patch. But let's merge anyway

PR #1463, opened 2026-08-05
✨Support XValidation on type aliases to map/slice/pointer types

### LINKED ISSUE #1462: controller-gen does not support XValidation on type aliases
### What happened?

`controller-gen` does not generate `x-kubernetes-validations` (CEL rules from `+kubebuilder:validation:XValidation` markers) when the marker is placed on a type alias defined with `=`.

In Go, a type alias (`type X = map[string]string`) is just another name for the underlying type, not a distinct type. CEL validation rules attached to such aliases are silently ignored by `controller-gen`.

### What did you expect to happen?

`+kubebuilder:validation:XValidation` markers on type aliases should produce `x-kubernetes-validations` entries in the generated CRD, just as they do for named types (`type X map[string]string`).

### Example

```go
// +kubebuilder:validation:XValidation:rule="self.size() > 0",message="must not be empty"
type Labels = map[string]string

type MySpec struct {
    Labels Labels `json:"labels,omitempty"`
}
```

The generated CRD for `labels` does not contain `x-kubernetes-validations`.

### Workaround

Use a named type instead of an alias:

```go
// +kubebuilder:validation:XValidation:rule="self.size() > 0",message="must not be empty"
type Labels map[string]string
```

### Environment

- controller-tools version: latest (main)
- Kubernetes version: 1.32+

dongjiang1989:
### What does this PR do, and why do we need it?

Fixes https://github.com/kubernetes-sigs/controller-tools/issues/1462

`controller-gen` did not generate `x-kubernetes-validations` for type aliases to complex types (e.g. `type X = map[string]string`). The alias was unwrapped to the underlying type in `localNamedToSchema`, which hit the default case and produced an error:

```
unsupported type *types.Map for identifier X
```

### Changes

For user-defined type aliases (`type X = Y`) whose RHS is a non-basic, non-named type (map, slice, pointer, etc.), emit a `$ref` to the alias type instead of unwrapping it. The flattener then resolves the `$ref` via `infoToSchema`, which correctly applies markers on the alias declaration (including `+kubebuilder:validation:XValidation`).

Behavior for other alias types is preserved:
- **Basic type aliases** (`type X = string`): unwrap, emit inline type/format + `$ref` (preserves field-level markers like MinLength/MaxLength)
- **Named type aliases** (`type X = SomeStruct`): unwrap, emit `$ref` to the underlying named type (preserves inline/embedded struct alias behavior and applyconfiguration generator compatibility)
- **Built-in aliases** (`any`): unwrap (no `$ref`, since `any` is not a user-defined type)

### Example

```go
// +kubebuilder:validation:XValidation:rule=`self.all(key, size(key) > 0)`,message="keys must be non-empty"
type Labels = map[string]string

type MySpec struct {
    Labels Labels `json:"labels,omitempty"`
}
```

Now correctly generates:

```yaml
labels:
  additionalProperties:
    type: string
  type: object
  x-kubernetes-validations:
  - message: keys must be non-empty
    rule: self.all(key, size(key) > 0)
```

### DIFF
--- a/pkg/crd/schema.go
+++ b/pkg/crd/schema.go
@@ -304,10 +304,29 @@ func localNamedToSchema(ctx *schemaContext, ident *ast.Ident) *apiextensionsv1.J
 		ctx.pkg.AddError(loader.ErrFromNode(fmt.Errorf("unknown type %s", ident.Name), ident))
 		return &apiextensionsv1.JSONSchemaProps{}
 	}
-	// This reproduces the behavior we had pre gotypesalias=1 (needed if this
-	// project is compiled with default settings and Go >= 1.23).
+	// For user-defined type aliases (type X = Y, gotypesalias=1), the handling
+	// depends on the RHS type:
+	//   - Basic types (type X = string): unwrap to preserve existing behavior
+	//     (inline type/format needed for field-level markers like MinLength).
+	//   - Named types (type X = SomeStruct): unwrap to preserve existing behavior
+	//     (emits $ref to the underlying named type, needed for inline/embedded
+	//     struct aliases and the applyconfiguration generator).
+	//   - Complex types (type X = map[string]string, []string, etc.): the
+	//     existing switch would error out because these types have no Obj().
+	//     Instead, emit a $ref to the alias type so markers on the alias
+	//     declaration (e.g. +kubebuilder:validation:XValidation) are applied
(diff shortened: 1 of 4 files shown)


dongjiang1989:
cc @sbueringer PTAL, thanks

PR #1470, opened 2026-08-27
🌱 Bump k8s.io/* to v0.37.0

dongjiang1989:
## What does this do?

Bump k8s.io dependencies from v0.37.0-beta.0 to v0.37.0.

## Changes

- Update k8s.io dependencies (`api`, `apimachinery`, `apiextensions-apiserver`, `apiserver`, `code-generator`, `client-go`, `component-base`) to v0.37.0
- Update `go.mod`/`go.sum` in main module and all testdata directories
- Regenerate golden test files (`internal.go`, CRD YAMLs) for v0.37.0 schema changes (new Job scheduling fields, updated descriptions, etc.)
- Work around k8s.io/api v0.37.0 duplicate `x-kubernetes-list-map-keys` bug (see below)
- Use `cmpopts.EquateEmpty()` in CRD tests to handle nil vs empty `Properties` maps

## Workaround: k8s.io/api v0.37.0 duplicate listMapKey bug

Several fields in k8s.io/api v0.37.0 carry both `+listMapKey=name` and `+k8s:listMapKey=name` on the same slice field — for example `core/v1.EvictionResponders` and `scheduling/v1alpha3.WorkloadPodGroupResourceClaim`. controller-gen honours both markers and emits `x-kubernetes-list-map-keys: [name, name]`, which the Kubernetes API Server rejects with:

```
must not contain duplicate entries for x-kubernetes-list-map-keys
```

Deduplication is added in `pkg/crd/markers/topology.go` when applying the `listMapKey` marker. This single location is sufficient because `applyconfiguration` uses the same `crd.Parser` and goes through the same marker application path.

A regression test field `DuplicateListMapKey` has been added to both `pkg/crd/testdata/cronjob_types.go` and `pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go` with both `+listMapKey` and `+k8s:listMapKey` on the same field to verify the deduplication works correctly.

This workaround is purely defensive; it can be removed once k8s.io/api drops the duplicate annotations.


sbueringer:
cc @JoelSpeed regarding the bug

pkg/applyconfiguration/openapi.go
```diff
@@ -351,3 +361,54 @@ func sanitizeForOpenAPIV2(schema map[string]any) {
 		}
 	}
 }
+
+// dedupListMapKeys recursively walks a schema map and removes duplicate entries
+// from x-kubernetes-list-map-keys arrays. This is needed because k8s.io/api
+// v0.37.0 has some types with both +listMapKey and +k8s:listMapKey annotations
+// on the same field, which produces duplicate entries in the generated schema.
+func dedupListMapKeys(schema map[string]any) {
```
camilamacedo86: Hi, thanks for the update. 🎉 

I agree that duplicate-key handling should not be implemented in two different places. However, I think we still need the normalization itself.

Kubernetes v1.37 contains fields annotated with both +listMapKey and +k8s:listMapKey, for example PodSpec.EvictionResponders  (https://github.com/kubernetes/kubernetes/blob/v1.37.0/staging/src/k8s.io/api/core/v1/types.go#L4474-L4480).

When both markers are processed, they produce duplicate x-kubernetes-list-map-keys, which the Kubernetes API server rejects.

Therefore, I suggest keeping the deduplication in pkg/crd/markers/topology.go (https://github.com/kubernetes-sigs/controller-tools/blob/b6e9338074e52bf33d74e4f1e6975e4acee3153f/pkg/crd/markers/topology.go#L204-L220),
  since this is the central point where the markers are applied.

  The additional recursive deduplication in pkg/applyconfiguration/openapi.go
  (https://github.com/kubernetes-sigs/controller-tools/blob/b6e9338074e52bf33d74e4f1e6975e4acee3153f/pkg/applyconfiguration/openapi.go#L139-L190)
  appears to be redundant because applyconfiguration uses the same CRD parser
  and marker implementation. Could we remove that second pass and add a focused
  regression test that applies both marker forms to the same field and verifies
  that only one key is generated?

  Thanks!

camilamacedo86 (changes requested):

sbueringer:
> This workaround is purely defensive; it can be removed once k8s.io/api drops the duplicate annotations.

I assume they will never drop both markers, and I think deduplication is something that we should have anyway in controller-gen

pkg/applyconfiguration/openapi.go
sbueringer: @dongjiang1989 Yup looks like this is not necessary on top of the change in topology.go, or did we miss something?

pkg/crd/markers/topology.go
```diff
@@ -208,7 +209,14 @@ func (l ListMapKey) ApplyToSchema(ctx *SchemaContext, schema *apiextensionsv1.JS
 	if schema.XListType == nil || *schema.XListType != string(Map) {
 		return fmt.Errorf("must apply listMapKey to an associative-list")
 	}
-	schema.XListMapKeys = append(schema.XListMapKeys, string(l))
+	key := string(l)
```
sbueringer, line 212: Let's add duplicate markers to our existing tests to verify the deduplication works

dongjiang1989: Thanks @sbueringer  Add test case done

pkg/applyconfiguration/openapi.go
dongjiang1989: Thanks @camilamacedo86  You are right. Just keep the deduplication in `pkg/crd/markers/topology.go`

dongjiang1989:
@sbueringer All changed, Please re-check it

sbueringer:
Thx!

/lgtm
/approve

PR #1474, opened 2026-09-06
:bug: crd: support Kubernetes declarative validation bounds

### LINKED ISSUE #1473: APIs embedding batchv1.JobSpec get rejected due to missing k8s:maxItems implementation
Controller-tools implementing `+k8s:immutable` and not implementing `k8s:maxItems` results in CRDs embedding `batchv1.JobSpec` to get rejected by the apiserver due to exceeding the CEL cost budget with an error like the below:

```
CustomResourceDefinition.apiextensions.k8s.io "clustertrainingruntimes.trainer.kubeflow.org" is invalid:
[spec.validation.openAPIV3Schema.properties[spec].properties[template].properties[spec]
  .properties[replicatedJobs].items.properties[template].properties[spec].properties[scheduling]
  .properties[resourceClaims].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost
  exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems,
  maxProperties, and maxLength where arrays, maps, and strings are declared),
 ...
 spec.validation.openAPIV3Schema: Forbidden: x-kubernetes-validations estimated rule cost total
  for entire OpenAPIv3 schema exceeds budget by factor of more than 100x]
```

kannon92:
This fixes an issue where `controller-gen` recognizes `+k8s:immutable` but silently drops the related bounds declared with:

- `+k8s:maxItems`
- `+k8s:minimum`
- `+k8s:maximum`

These bounds are used by the Kubernetes API server when estimating the cost of CEL validation rules. Dropping them can make an otherwise valid generated CRD exceed the CEL cost budget and be rejected, particularly when embedding Kubernetes types such as `batch/v1.JobSpec`.

This PR registers the three markers and maps them to the equivalent OpenAPI schema fields. It also supports the trailing comments used by Kubernetes declarative validation markers, for example:

```go
// +k8s:maximum=2 # Bounds the CEL validation cost.
```

The CRD golden fixtures have been updated to verify that these bounds are preserved in generated schemas.

Fixes #1473.

### DIFF
--- a/pkg/crd/markers/validation.go
+++ b/pkg/crd/markers/validation.go
@@ -137,6 +137,15 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("k8s:immutable", markers.DescribesField, Immutable{})).
 		WithHelp(Immutable{}.Help()),
+
+	// Kubernetes declarative validation aliases.  These bounds are especially
+	// important for keeping the estimated cost of CEL rules finite.
+	must(markers.MakeDefinition("k8s:maxItems", markers.DescribesField, MaxItems(0))).
+		WithHelp(MaxItems(0).Help()),
+	must(markers.MakeDefinition("k8s:minimum", markers.DescribesField, Minimum(0))).
+		WithHelp(Minimum(0).Help()),
+	must(markers.MakeDefinition("k8s:maximum", markers.DescribesField, Maximum(0))).
+		WithHelp(Maximum(0).Help()),
 }
 
 // ValidationIshMarkers are field-and-type markers that don't fall under the
@@ -207,6 +216,33 @@ type Minimum float64
 
 func (m Minimum) Value() float64 {
 	return float64(m)
+}
+
+func declarativeValidationMarkerValue(name, restFields string) string {
+	if strings.HasPrefix(name, "k8s:") {
+		if comment := strings.Index(restFields, " #"); comment >= 0 {
+			restFields = restFields[:comment]
+		}
+	}
+	return strings.TrimSpace(restFields)
+}
+
+func (m *Maximum) ParseMarker(_, name, restFields string) error {
+	value, err := strconv.ParseFloat(declarativeValidationMarkerValue(name, restFields), 64)
+	if err != nil {
+		return fmt.Errorf("invalid maximum value: %w", err)
+	}
+	*m = Maximum(value)
+	return nil
+}
+
+func (m *Minimum) ParseMarker(_, name, restFields string) error {
+	value, err := strconv.ParseFloat(declarativeValidationMarkerValue(name, restFields), 64)
+	if err != nil {
+		return fmt.Errorf("invalid minimum value: %w", err)
+	}
+	*m = Minimum(value)
+	return nil
 }
 
 // ExclusiveMinimum indicates that the minimum is "up to" but not including that value.
@@ -284,6 +320,15 @@ type Pattern string
 //
 // +controllertools:marker:generateHelp:category="CRD validation"
 type MaxItems int
+
+func (m *MaxItems) ParseMarker(_, name, restFields string) error {
+	value, err := strconv.Atoi(declarativeValidationMarkerValue(name, restFields))
+	if err != nil {
+		return fmt.Errorf("invalid maxItems value: %w", err)
+	}
+	*m = MaxItems(value)
+	return nil
+}
 
 // MinItems specifies the minimum length for this list.
 //
--- a/pkg/crd/testdata/declarativevalidation/types.go
+++ b/pkg/crd/testdata/declarativevalidation/types.go
@@ -1,0 +1,47 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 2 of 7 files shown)


pkg/crd/testdata/immutable/types.go
```diff
@@ -60,6 +60,7 @@ type ImmutableSpec struct {
 	// +optional
```
alvaroaleman, line 60: Please create a new crd under pkg/crd/testdata rather than re-using the one used for immutable testing as well as actual tests for it under pkg/crd/testdata/$crdname_test.go

pkg/crd/markers/validation.go
```diff
@@ -137,6 +137,15 @@ var FieldOnlyMarkers = []*definitionWithHelp{
 
 	must(markers.MakeDefinition("k8s:immutable", markers.DescribesField, Immutable{})).
 		WithHelp(Immutable{}.Help()),
+
+	// Kubernetes declarative validation aliases.  These bounds are especially
+	// important for keeping the estimated cost of CEL rules finite.
+	must(markers.MakeDefinition("k8s:maxItems", markers.DescribesField, K8sMaxItems(0))).
```
alvaroaleman: Instead of re-implementing, can we register the existing `MaxItems` under the new name?

alvaroaleman (commented):
cc @sbueringer @JoelSpeed 

The problem here is that we currently generate a schema for batchv1.Job that the apiserver rejects, since we implemented the immutability as a CEL rule but don't have a maxLength implementation. Does that warrant backporting this change (maybe reduced to just maxLength)? It is technically a breaking change in the sense that we start generating for something we previously ignored. Can you think of any cases where that would cause problems?

sbueringer:
> Does that warrant backporting this change (maybe reduced to just maxLength)?

Fine to backport for me

> It is technically a breaking change in the sense that we start generating for something we previously ignored

That's the same every time we implement support for a pre-existing `k8s:` marker. The difference is here mostly that it's in a patch release, right?

> Can you think of any cases where that would cause problems?

I think only the kind of problems that you would expect if a new "max*" validation shows up. I think there could be cases where folks only used max* without immtuable. So their CRDs were not invalid and they might have already CRs with values longer than max*? (but IIRC the kube-apiserver implements ratcheting for these validations?)


Maybe what we should do in general is to provide a way to disable handling for specific markers via some controller-gen command line flag? (if that's not already possible) So there is some way to opt-out from validation that suddenly pops up.

pkg/crd/testdata/immutable/types.go
kannon92: done.

pkg/crd/markers/validation.go
kannon92: I'm not sure I quite follow what you mean here. I tried to apply your suggestion here. 

Let me know if I missed the point.

mimowo:
Thank you for back porting, this is currently blocking the upgrade for JobSet which in turns blocks upgrade of Kieue - to k8s 1.37

alvaroaleman:
> Thank you for back porting, this is currently blocking the upgrade for JobSet which in turns blocks upgrade of Kieue - to k8s 1.37

There is a known workaround, leaving comments like that is not helpful, especially not if they just re-iterate known information.

pkg/crd/testdata/declarativevalidation/types.go
```diff
@@ -0,0 +1,47 @@
+/*
```
alvaroaleman, line 1: This is not the first delaractive marker added to the project, so please name it after what it is, something like `boundsvalidation` or some such

pkg/crd/markers/validation.go
```diff
@@ -209,6 +218,33 @@ func (m Minimum) Value() float64 {
 	return float64(m)
 }
 
+func declarativeValidationMarkerValue(name, restFields string) string {
```
alvaroaleman, line 221:
```suggestion
// stripCommentIfDeclarativeValidationMarker strips comments iff the marker
// is a declarative validation marker, mimicking validation-gens behavior:
// https://github.com/kubernetes/gengo/blob/25e2208e0dc371a827289e7faced19a2dbcd480b/v2/codetags/parser.go#L88
func stripCommentIfDeclarativeValidationMarker(name, restFields string) string {
```


and please open an issue so we can discuss if we should have this behavior all markers, not just the delcarative validation ones

pkg/crd/markers/validation.go
```diff
@@ -137,6 +137,15 @@ var FieldOnlyMarkers = []*definitionWithHelp{
...
 		WithHelp(Immutable{}.Help()),
+
+	// Kubernetes declarative validation aliases.  These bounds are especially
+	// important for keeping the estimated cost of CEL rules finite.
+	must(markers.MakeDefinition("k8s:maxItems", markers.DescribesField, MaxItems(0))).
+		WithHelp(MaxItems(0).Help()),
+	must(markers.MakeDefinition("k8s:minimum", markers.DescribesField, Minimum(0))).
+		WithHelp(Minimum(0).Help()),
+	must(markers.MakeDefinition("k8s:maximum", markers.DescribesField, Maximum(0))).
+		WithHelp(Maximum(0).Help()),
```
alvaroaleman, line 148: The help text for all of these is wrong, as it includes the `+kubebuilder:validation:` prefixed name of the marker

pkg/crd/markers/validation.go
kannon92: I'm not really following what you want me to write for this issue.

kannon92: Tried to draft an issue. https://github.com/kubernetes-sigs/controller-tools/issues/1479

pkg/crd/markers/validation.go
kannon92: updated.

pkg/crd/markers/validation.go
kannon92: applied similar suggestion.

pkg/crd/testdata/boundsvalidation/types.go
```diff
@@ -0,0 +1,47 @@
+/*
```
kannon92: done.

